PR #10, opened 2015-11-23
Fix ML compilation w.r.t. trunk.

ppedrot:
This commit fixes the compilation of ssreflect for Coq trunk branch (6474fa6c49).

### DIFF
--- a/mathcomp/ssreflect/plugin/trunk/ssreflect.ml4
+++ b/mathcomp/ssreflect/plugin/trunk/ssreflect.ml4
@@ -44,6 +44,7 @@ open Coqlib
 open Glob_term
 open Util
 open Evd
+open Sigma.Notations
 open Extend
 open Goptions
 open Tacexpr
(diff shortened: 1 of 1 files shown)


gares:
Thanks for the patch!
But I guess I'd prefer a single function adapting the new API to the old one... most of the patch seems to repeat the very same pattern over and over.

ppedrot:
Well, I think ssreflect would gain a lot by switching to the new (unfinished) API, as it would allow for a static expression of relative timing between evarmaps. The current patch is compatible with a gradual typing of extensibility constraints...

gares:
Sorry but I'm not following.  I'm OK for the new API, but this can be done only when the API is final.  Your pull request seem to wrap the new API into a compatibility function, but that function is not defined, it is just inlined.  What do you mean by "compatible with gradual..."? An why this is not possible if one defines a wrapper once an forall?  Best.

ppedrot:
I mean that under the assumption that in ssreflect you want to use the new API rather than just port the code to it (i.e. making explicit constraints in the type of ssreflect functions), this is going to require a non negligible amount of work. Therefore, you'll have to do it on several passes, and you'll end up needing several variants of the wrapper, according to the types of its arguments and returned value (depending of what part of the code around you upgraded). The advantage of the current patch is that you can remove the compatibility lines one by one on a per-use basis.

gares:
merged

PR #12, opened 2015-12-03
Removing the only use of globTacticIn.

ppedrot:
The globTacticIn primitive is a hack that was chiefly motivated by historical reasons. The following patch gets rid of its only use and actually even simplifies the code.


gares:
thanks

PR #13, opened 2016-01-06
do not use `sed -i' in ssrcoqdep -- this is not portable

strub:
This prevents compilation of ssreflect on OS-X/*BSD.


gares:
Thanks

PR #20, opened 2016-01-31
Partially fixing ML compilation on trunk.

ppedrot:
Here is a partial fix for ML compilation on trunk. I fixed as much as I could, but I am unaware of some changes in intropatterns, so that I stopped were I couldn't progress anymore.

There is a FIXME related to the fact we need to be more clever in the persistence of typing information of tactic printing.

### DIFF
--- a/mathcomp/ssreflect/plugin/trunk/ssreflect.ml4
+++ b/mathcomp/ssreflect/plugin/trunk/ssreflect.ml4
@@ -207,7 +207,7 @@ let add_genarg tag pr =
   let wit = Genarg.make0 None tag in
   let glob ist x = (ist, x) in
   let subst _ x = x in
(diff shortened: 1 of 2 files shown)


gares:
Many thanks PMP!

I've a question.  Clearly having a value escape the monad via an assignment is not very elegant.  Is what we do in ssr completely unsupported by the new API, or I should "just" rewrite everything in that monad?

ppedrot:
This is clearly a hack, but at the same time the proper API to escape the Ftactic monad into non-unit Proofview.tactic is unclear. Note that the same problem arises for the Proofview.tactic monad, because Proofview.V82.of_tactic does lose information and is, as such, a hack. It would be indeed more reasonable to put the code of ssreflect into the (various) monads.

PR #24, opened 2016-02-09
[div] Move dvdn_fact from prime to div.

ejgallego:
Hi,

I was wondering whether it would make sense to merge this lemma by GG.

```` coq
Lemma dvdn_fact m n : 0 < m <= n -> m %| n`!.
````

My request has no other reason than a selfish desire to improve the welcome page at https://x80.org/rhino-coq.

### DIFF
--- a/mathcomp/ssreflect/div.v
+++ b/mathcomp/ssreflect/div.v
@@ -457,6 +457,12 @@ Qed.
 Lemma dvdn_exp k d m : 0 < k -> d %| m -> d %| (m ^ k).
 Proof. by case: k => // k _ d_dv_m; rewrite expnS dvdn_mulr. Qed.
 
+Lemma dvdn_fact m n : 0 < m <= n -> m %| n`!.
+Proof.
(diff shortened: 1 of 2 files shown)


amahboubi:
Hi Emilio, 

this lemma already exists in [prime.v](https://github.com/math-comp/math-comp/blob/master/mathcomp/ssreflect/prime.v#L593), under the same name (and with the same proof).

But I agree it looks misplaced, so we could indeed move it to div. Any objection?

ejgallego:
Hi Assia,

Ups! Sorry for the noise, I had this on my TODO from mathcomp 1.5 times I guess and forgot to double check. Please close.

Indeed it would make more sense to have it in `div`, the lemma seems to be added in 1.6 so I guess not a lot of users should be impacted by the change. I've updated the PR just in case.

amahboubi:
Many thanks for the PR update!

ejgallego:
Thanks Assia !

PR #25, opened 2016-02-15
Fixing ML compilation with trunk.

ppedrot:


gares:
Thanks!

ppedrot:
I'm still pushing some patches that are going to break the ML compilation. I'll propose another pull-request when I'm done for today.

PR #33, opened 2016-03-03
[search] Use msg_info to notify search results.

ejgallego:
By default, the search command calls `Pp.msg` to print search results.

Unfortunately, this bypasses the `log_via_feedback` option and produces
not very nice results on feedback-depending IDES like JsCoq.

A proper fix would involve merging coq/coq#143 , and the upcoming search
cleanup, but this should do the trick for now.

I couldn't observe any problem with the usual testing.

### DIFF
--- a/mathcomp/ssreflect/plugin/trunk/ssreflect.ml4
+++ b/mathcomp/ssreflect/plugin/trunk/ssreflect.ml4
@@ -1456,7 +1456,7 @@ let interp_modloc mr =
 
 let ssrdisplaysearch gr env t =
   let pr_res = pr_global gr ++ spc () ++ str " " ++ pr_lconstr_env env Evd.empty t in
-  msg (hov 2 pr_res ++ fnl ())
+  msg_info (hov 2 pr_res ++ fnl ())
 
 VERNAC COMMAND EXTEND SsrSearchPattern CLASSIFIED AS QUERY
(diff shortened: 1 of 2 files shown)


gares:
thanks

ejgallego:
thanks Enrico!

thery:
Hi,

Does this solve the problem of the search that was not working properly 
in jsCoq?

ejgallego:
@thery there have been many improvement in the search support, both in Coq and in jsCoq so IMHO search works much better now.

However, large searches still generate `StackOverflows`, I think I know what the problem is and won't be easy to fix; for now I suggest to workaround this issue using `Search ... in module`.

ejgallego:
Just for reference @thery, I think I have fully nailed the search issue, see: ejgallego/jscoq#13

gmalecha:
@ejgallego What are the largest searches that you've done?

thery:
I guess a search that returns more than 1000 results is useless

ejgallego:
@gmalecha searching with Mathcomp loades easily goes throu filtering ten-of-thousands of entries (as @thery points out, not all of them should be returned thou).

ejgallego:
Hi @thery , recent improvements to the Chrome recursion optimizer means that large searches seem to work OK now. Do you have a concrete test case? Thanks.

[BTW, I won't spam this issue anymore, people interested in Search performance can follow ejgallego/jscoq#13 ]

PR #38, opened 2016-03-27
Fixing ML compilation.

ppedrot:
This patch fixes the ML compilation of mathcomp. Yet, the compilation of the Coq files dies early because of the failure of an assertion added by commit 4d25b224. I'm not sure what to do about this one, though.


ppedrot:
The assertion issue was fixed by Hugo in 111e5edf.

gares:
Thanks!

PR #39, opened 2016-04-08
Fixing compilation after the merge of PR trunk-function_scope.

ppedrot:


ppedrot:
Up.

gares:
Many thanks, but I had no time to look at it.  It is small, but I quite don't get why "Local" and what/how it fixes.  Tomorrow I'll have some time, but an explanation may help.

ppedrot:
The merge introduced a new notation scope for functions, so that some notations that used a syntaxical parameter in two different scopes (e.g. as an argument vs. as a function) would fail. There were two places where it broke, so that I added the Argument directive to remove the %function scope for the problematic arguments. As this only affects the notation below, I made it local so as not to interfere with the remainder of the files, even though I think it would not have changed anything in practice.

PR #40, opened 2016-05-02
Fixing compilation after the merge of ML-tactic-notation branch.

ppedrot:
This patch fixes trunk compilation, which was broken because of change in API. It is not entirely satisfactory though, because some hand-defined tactics in ssreflect won't be printed properly. There is currently no fix for it, but I am not sure how much this is an issue.


gares:
Thanks

PR #42, opened 2016-05-09
Fix compilation after the merge of the dynamic tactic value branch.

ppedrot:
This fixes compilation after commit 9fe0471ef, which is mostly an API change.


gares:
Thanks!

PR #46, opened 2016-05-18
Fix compilation after the change of API in Tactics.

ppedrot:
This fixes compilation after all tactics in Tactics were moved to the Proofview.tactic type.


gares:
thanks

PR #57, opened 2016-08-12
Fix compilation with respect to commit 8e38b0e in the "trunk" branch of Coq

matej-kosik:
The commit in this pull-request proposes a possible replacement for the call of `Context.Rel.Declaration.of_tuple` function which was removed from Coq trunk.


gares:
Thanks

PR #58, opened 2016-08-24
Removing calls of "Context.Named.Declaration.{of,to}_tuple" functions

matej-kosik:
Coq trunk currently provides these two functions:
```
Context.Named.Declaration.of_tuple
Context.Named.Declaration.to_tuple
```

These were temporarily defined to enable safe transition from the old definition of the `Context.Named.t` datatype to the new one. These functions now are no longer either necessary or beneficial.

In this pull-request I propose one possible way how to get rid of them and get the full benefit of the new representation. Primarily, __the reader__ will have zero doubt:
- what is being constructed
- what is being destructed (and the meaning of particular bound variables).

### DIFF
--- a/mathcomp/ssreflect/plugin/trunk/ssreflect.ml4
+++ b/mathcomp/ssreflect/plugin/trunk/ssreflect.ml4
@@ -107,11 +107,6 @@ let anomaly s = CErrors.anomaly (str s)
 let ppnl  = msg_info
 let msgnl = msg_info
 
-let mk_reldecl name obody ty =
-  match obody with
-  | None -> RelDecl.LocalAssum (name, ty)
-  | Some bo -> RelDecl.LocalDef (name, bo, ty)
-
 (** look up a name in the ssreflect internals module *)
 let ssrdirpath = make_dirpath [id_of_string "ssreflect"]
 let ssrqid name = make_qualid ssrdirpath (id_of_string name) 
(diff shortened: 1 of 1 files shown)


mathcomp/ssreflect/plugin/trunk/ssreflect.ml4
```diff
@@ -2093,10 +2090,15 @@ let abs_wgen keep_let ist f gen (gl,args,c) =
...
+     (match pf_get_hyp gl x with
+     | LocalAssum (_,ty) ->
+         gl,
+         mkVar x :: args,
+         mkProd_or_LetIn (RelDecl.LocalAssum (Name (f x),ty)) (subst_var x c)
+     | LocalDef (_,b,ty) ->
+         gl,
+         args,
+         mkProd_or_LetIn (RelDecl.LocalDef (Name (f x),b,ty)) (subst_var x c))
   | _, Some ((x, _), None) ->
```
gares: Sorry but I don't like such duplication of code, this is why I did not merge this part of your previous pull request (and defined the mk_reldecl function).

gares:
Thanks for the PR!

matej-kosik:
I've added a commit  https://github.com/math-comp/math-comp/pull/58/commits/6bc53af07a100aad305393edb14c4a3d73b3e3b7 which addresses the code duplication that you pointed out.

Consequences:

__(1)__

This:
```Ocaml
bo <> None
```
was replaced with:
```Ocaml
NamedDecl.is_local_def decl
```

The first version works flawlessly.

The second version likewise, but there is an additional bonus;
(for those ones who read the code, not for those ones who have written it)
it is obvious what that condition actually means.

__(2)__

This:
```Ocaml
let _, bo, ty = Named.Declaration.to_tuple (pf_get_hyp gl x) in
mkProd_or_LetIn (mk_reldecl (Name (f x)) bo ty) (subst_var x c)
```
was replaced with:
```Ocaml
let decl = pf_get_hyp gl x in
mkProd_or_LetIn (decl |> NamedDecl.to_rel |> RelDecl.set_name (Name (f x)))
                (subst_var x c)
```
The only point here is that we drop conversion `to_tuple`.
So, instead of converting named-declaration to tuple which was then converted to rel-declaration,
we convert named-declaration directly to rel-declaration.

gares:
thanks

PR #65, opened 2016-09-07
abstract_context utility lemma

gares:
I'm not sure this lemma belongs here, but it is somewhat useful.


amahboubi:
I like this one and actually used a similar device in my own proof scripts. Any reason why it's not merged yet?

PR #68, opened 2016-09-19
Fix ML compilation after Ltac refactoring.

ppedrot:


gares:
I think the change to Make is wrong, isn't the Makefile adding that line (on some versions of Coq)?

ppedrot:
Oh, you're right. I tend to use `git commit -a` too easily. I'll push a fixup.

gares:
I could not find a better hack, patching Make is ugly, but only 8.6+ wants the mlpack file...

ppedrot:
Couldn't you that using some form of conditional inclusion in the makefile?

ejgallego:
@gares, note that the makefile patching mess things for me in the following way:
- after a make the git checkout won't be clean.
- repeated copies of `ssreflect_plugin.mlpack` get added every time make is run.

PR #69, opened 2016-09-20
[field] Remove unnecessary use of `Program Definition` use

ejgallego:
Simple `Definition` should work fine here.

This avoids the problem:

`Error: Library Coq.Program.Tactics has to be required first.`

in math-comp versions that depends on a minimal (or no) Coq stdlib.

Tested on 8.5/8.6


ejgallego:
Thanks!!!

PR #79, opened 2016-11-04
Add direct link to the archives to help google indexing.

ejgallego:
The information page is forbidden by INRIA robots so we may want to hint
google where the accessible archives are.

### DIFF
--- a/index.html
+++ b/index.html
@@ -94,7 +94,8 @@ The Ssreflect language comes with a dedicated
 Interested?
 <a href="mailto:sympa@inria.fr?subject=SUBSCRIBE%20ssreflect">Subscribe to the ssreflect mailing list</a>
 and let us know what you are using our libraries for, ask questions, etc.
-You can also browse the <a href="https://sympa.inria.fr/sympa/info/ssreflect">archives of the list</a>.
+You can also browse the <a href="https://sympa.inria.fr/sympa/arc/ssreflect">archives of the list</a> or consult the
+<a href="https://sympa.inria.fr/sympa/info/ssreflect">general information page</a>.
 </p>
 <h3>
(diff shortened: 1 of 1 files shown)


gares:
@maximedenes what was the magic trick to get coq-club indexed?

amahboubi:
Ciao @gares , I opened a ticket for this 2 days ago. The IT guys did the same trick as for coq-club : let's wait for a couple of days to see if it works.

ejgallego:
Indeed @gares that should be set, a direct link to the archives was missing for google to pick them up, as robots.txt still forbids accesing the mailing list information page.

Now that I think of it, @amahboubi , it could make sense to request the INRIA ops to allow access to that page. Now if you google "ssreflect mailing list" you'll go directly to the archives.

ejgallego:
@maximedenes the same applies to coq-club, if you google "coq club mailing list" you land into the archives.

[Generally the whole Disallow: / of sympa's robots seems a bit strange IMVVHO]

amahboubi:
@ejgallego : can you confirm that "your" google finds the archive with this query? I do not manage to reproduce this success...

ejgallego:
@amahboubi indeed the list seems poorly indexed still, I was hoping that this trick would improve matters but so far it hasn't.

Currently I need to google for the title of the page, thus "ssreflect discussion list" ; what is weird is that discussions doesn't seem to be followed by google yet.

ejgallego:
@amahboubi It seems that google has picked up with the archives and the situation has improved a bit, example: https://www.google.com/#q=ssreflect+tensor+product gets me to the right place.

amahboubi:
Thanks for reporting @ejgallego. Actually I got a message yesterday  from  the Inria sysadmins  about this issue. May be they found the right incantation to be placed in robot.txt. I'll keep you posted here once I get feedback from them.

amahboubi:
Indeed, this issue seems to be solved. @maximedenes : they applied the same trick for coq-club :).

PR #104, opened 2017-02-06
using opam in travis before_install script

CohenCyril:
And triggering a build

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+
 dist: trusty
 sudo: required
 language: c
@@ -32,11 +33,10 @@ install:
 - "[ -e .opam ] || opam init -j ${NJOBS} --compiler=${COMPILER} -n -y"
 - eval $(opam config env)
 - opam config var root
-- opam install -j ${NJOBS} -y ocamlfind camlp5 ${EXTRA_OPAM}
-- opam list
 - opam repo add coq-core-dev https://coq.inria.fr/opam/core-dev || true
-- opam pin add -y -n coq ${TEST_TARGET}
-- opam install coq -y -v -j ${NJOBS}
+- opam pin add -y -n coq -k version ${TEST_TARGET}
+- opam install -y -v -j ${NJOBS} coq ${EXTRA_OPAM}
+- opam list
 
 script:
 - echo 'Building Mathcomp...' && echo -en 'travis_fold:start:mathcomp.build\\r'
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,14 +19,14 @@ env:
   - COMPILER="system"
   # Main test targets
   matrix:
-  - TEST_TARGET="v8.5"
-  - TEST_TARGET="v8.6"
-  - TEST_TARGET="trunk"
+  - TEST_TARGET="8.5.3"
+  - TEST_TARGET="8.6"
+  - TEST_TARGET="dev"
 
 matrix:
   allow_failures:
   # v8.5 is too slow to build mathcomp
-  - env: TEST_TARGET="v8.5"
+  - env: TEST_TARGET="8.5.3"
 
 install:
 - "[ -e .opam ] || opam init -j ${NJOBS} --compiler=${COMPILER} -n -y"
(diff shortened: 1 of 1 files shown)


ejgallego:
Note the interaction of '.opam' caching vs installing a coq `.dev` version.
Indeed, to get fully correct builds, you may need to disable caching.

We chose to cache `.opam` as we were really tight on timing and were installing a complete, heavy toolchain there. However, for this kind of setup, `.opam` caching is indeed more of an historical artifact.

ejgallego:
I would also personally test 8.6 vs 8.6.dev, but of course that is a matter of preferences. Also, note that we have two different coq packages in opam, the one in the official repos and the one in coq-dev, @maximedenes , what is going on here?

CohenCyril:
@ejgallego would an opam upgrade be sufficient to trigger the rebuild of coq dev ?

ejgallego:
@CohenCyril that seems like a great thing to try, I am not expert in opam but I think it should work.

ejgallego:
Something to take into account is that caching is far from free in Travis; indeed the download/unpacking + packing/upload time does count for the build. So it could be well the case that we end up losing net time if we download a fully cached Coq version just to rebuild it.

CohenCyril:
Do you have an idea of how much time it takes to use a the cache?
Do you know if it possible to reuse cached data from a Travis build of Coq, so that we decorelate building Coq from building math-comp?

ejgallego:
For a full coq building environment, the cache takes 8 seconds to download an unpack. Then, it takes 30 seconds to reupload it, but this should be surely skipped as the reload is caused by opam adding some unnecessary logging information in `.opam/log`. See: https://travis-ci.org/coq/coq/builds/198948812 for an example.

> Do you know if it possible to reuse cached data from a Travis build of Coq, so that we decorelate building Coq from building math-comp?

Not from the Coq travis run unless we setup something special, I think it is not a good idea. If you want binary packages you'll have to use apt for them.

ejgallego:
From a testing point of view, I think it is a bad idea to decorrelate Coq and Math-comp, IMHO you want to be sure they run well together with the respective tips of the branches, otherwise you delay detection of problems.

More complex CI setups force a rebuild of math-comp whenever Coq changes, however I think we don't have the manpower to do that kind of integration so Coq will take care of its own test and math-comp can use any policy they steem convenient.

ejgallego:
The main reason I am a bit wary of caching coq is that it makes the build process too non-deterministic for my taste; build times may go well when the cache is hit, then a cache invalidation may make things to timeout. And note that the cache is generated at the end of the build.

CohenCyril:
OK, maybe this is not a good idea after all.

ejgallego:
Just to be clear in case I sounded too negative, I don't think that using opam is a bad idea, in fact it seems great to me, however I don't know how to integrate it well with the way Travis learns about dependencies.

CohenCyril:
I think it is technically possible to (automatically) build docker images for each version of Coq we want to test against (8.5.3, 8.6, and latest **compiling** 8.6.dev and dev versions), and reuse these images in our math-comp build, so that each job time is "time to download the appropriate Coq docker image" + "compilation time of mathcomp"... Given there is a docker integration to github, which someone already made work for Coq (cf https://hub.docker.com/r/skippa/coq/builds/)

And if odd-order is still too long, we could put it in an independent repository, store the build of mathcomp in a docker (cf https://sebest.github.io/post/using-travis-ci-to-build-docker-images/) and use this docker to build odd-order.

What do you think @ejgallego ?

ejgallego:
I am not familiar with Docker so I need to read the links first, but it sounds great!

One question thou is how the docker build would interact with the Travis job which is launched at commit time. How do we make the Travis odd_order wait for the math-comp docker image to be ready?

I was also thinking of using the `deploy` capabilities of Travis to indeed build some kind of caching, this should work well for the Coq case without adding too much overhead, (the first job will just upload a `tar.gz` of coq for the rest of ci jobs), however I am not sure mathcomp can really benefit from it.

ejgallego:
For instance, could we teach Coq's Travis build to deploy a Coq docker image at every Coq commit ?

CohenCyril:
yes we can (https://docs.travis-ci.com/user/docker/#Branch-Based-Registry-Pushes)

ejgallego:
That looks great. I could will try to add a docker push to Coq's travis , but it won't happen in February I'm afraid.

PR #108, opened 2017-02-09
Use pp_with as msg_with is being removed upstream.

ejgallego:
We are proposing to deprecate `msg_with` as it has a different purpose than the rest of `msg_*` functions. This change has no functional impact (`Pp.pp_with` doesn't flush but the ssr code does flush just after use).


ejgallego:
pinging

PR #118, opened 2017-04-03
Adapt the ssr plugin to Coq's PR#417.

maximedenes:
Let-ins in constrexpr and glob_constr now take an optional type, instead
of relying on a cast in the body.

Please review because it is not completely trivial.


gares:
I've the impression mkCLetIn is dead code :-/
Said that it looks good to me.  

Just in case @ggonthier has a minute to look at the patch: Coq's constr_expr for LetIn just got a place for the type of the expression being named. My understanding is that the BFdef needs no more a boolean argument (lines 5328 , 5384 are the places where the change seem to benefit SSR).

gares:
Merging now, the PR for Coq is going to be merged now...

PR #120, opened 2017-04-17
Coq PR #565: G_vernac.subgoal_command is replaced by query_command

SkySkimmer:


gares:
Thanks

SkySkimmer:
I should probably have made this explicit, but until the Coq PR is merged this makes ssreflect incompatible with Coq trunk. (pre coq/coq#565 query_command doesn't include the dot at the end of a `Check foo.`)

gares:
I see.  But I think no user of ssr trunk writes `- Check foo`. So I thought it would just make ssr "forward compatible".

PR #128, opened 2017-06-06
Fix plugin after Sigma removal.

maximedenes:
This PR adapts the plugin to the removal of the `Sigma` API (coq/coq#623).

### DIFF
--- a/mathcomp/ssreflect/plugin/trunk/ssreflect.ml4
+++ b/mathcomp/ssreflect/plugin/trunk/ssreflect.ml4
@@ -47,7 +47,6 @@ open Glob_term
 open Util
 open Evd
 open Proofview.Notations
-open Sigma.Notations
 open Extend
 open Goptions
 open Tacexpr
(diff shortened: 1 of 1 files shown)


ejgallego:
I guess it would have been more time-effective to have merged ssr first and then avoid the need of all those overlays as the changes can be folded in the PRs themselves, but YMMV.

maximedenes:
> I guess it would have been more time-effective to have merged ssr first and then avoid the need of all those overlays as the changes can be folded in the PRs themselves, but YMMV.

I see your point but:
1- I tried to be nice to everyone, so I scheduled my PRs (and the API, that I'll help fixing) after the ones conflicting with them, so only the Sigma removal could have been done after the ssr merge.
2- These overlays were often created before we agreed on the ssr merge.
3- Adapting the overlays one by one to the code that we merge (which is significantly different from the one in the ssr repo) is probably more costly than fixing it once and for all.

But indeed, I wasn't sure about which way was best. Anyway, I think I made it as smooth as I could for everyone except Matej and I.

PR #129, opened 2017-06-07
Ssr merge

maximedenes:
This PR makes math-comp use the ssreflect plugin that is being merged in Coq. I had to disable a test on primitive projection unfolding, will need to investigate if there's a bug.


mathcomp/ssreflect/Makefile.coq-makefile
```diff
@@ -1,22 +1,24 @@
...
-		EXTRA="ssrmatching.mli ssrmatching.ml4 ssrmatching.v";\
+		EXTRA="ssrmatching.mli ssrmatching.ml4 ssrmatching.v ssreflect.ml4";\
 		MLLIB=ssreflect_plugin.mllib;\
+		$$LN $(1)/plugin/$(V)/ssreflect.ml4 .;\
 	;;\
-	*)\
+	v8.6*)\
 		$$LN $(1)/plugin/$(V)/ssreflect_plugin.mlpack .;\
+		$$LN $(1)/plugin/$(V)/ssreflect.ml4 .;\
+		MLLIB=ssreflect_plugin.mlpack;\
```
gares: no EXTRA here? that is probably a bug

maximedenes: It is surprising, but was already like that before my PR, right? Do you confirm it's a bug?

maximedenes:
@gares Do you have a suggestion on how to restore 8.6 compatibility at the `.v` level? The current setup requires a `From Coq Require Import ssreflect`.

ejgallego:
@maximedenes you could play `LoadPath` tricks, but that won't be fun; otherwise I guess you would need to have a different `ssreflect.v` for 8.5/8.6 and a different one for `master`, that just exports the Coq shipped module.

ejgallego:
There is an even better option that is to revert to the `mathcomp` prefix, I cannot stress how annoying is now that all my ssreflect examples have to come in two versions...

maximedenes:
> There is an even better option that is to revert to the mathcomp prefix

Isn't that the same as your loadpath suggestion? I didn't change the prefix in any way, it is just the fact that we moved the files that had this side effect. To compensate for it, we should tweak the loadpath I guess.

maximedenes:
Btw, can a prefix be bound to several physical paths?

ejgallego:
> Isn't that the same as your loadpath suggestion?

The prefix is set arbitrarily by Coq, (in fact in SerAPI/jsCoq I have to imitate that arbitrary behavior).

You could well have a rule that declares `plugins/ssr` to be under prefix `mathcomp`. It is a bit uglier, but compatible.

ejgallego:
> Btw, can a prefix be bound to several physical paths?

it will complain with a warning, `overriding-logical-loadpath` whether it does make sense, I don't know.

maximedenes:
So I guess I don't understand your suggestion. You'd loose access to the `mathcomp` libraries, right?

ejgallego:
Likely you would have to put the full `mathcomp.ssreflect` in Coq, the rest should fine, right?

maximedenes:
I don't follow. If the `mathcomp` prefix points to `plugins/ssr`, `From mathcomp Require Import ssralg` will fail. If it points to the place where mathcomp is installed, it is `Require Import mathcomp.ssreflect` that will fail, right?

ejgallego:
The `mathcomp.ssreflect` prefix would point to `plugins/ssr`, `From mathcomp Require Import ssralg` resolves to `mathcomp.algebra.ssralg` I believe.

`Require Import mathcomp.ssreflect` already fails in 8.6 as far as I can see; it makes no sense with the current organization.

ejgallego:
Anyways the current loadpath story seems kinda bogus, the main reason duplication is not allowed is that `find_load_path` can determine the logical path a module uses in `start_library`.

IMHO this is the wrong approach, as it should be the user of `start_library` who provides that and I even believe that I saw some bug / non-flexible behavior due to that.

I'm afraid that the whole story with `start/declare_library` is a mess now, including real bugs such as 5556; I  am close to have a first patch ready thou to attack 5556 that in fact will properly pack together libraries with their loadpaths and other metaproperties that are set in an inconsistent way today. This will be done by enriching `Stm.init` so in fact the STM becomes in full control of the notion of "document/compilation unit".

Lets see where we arrive.

maximedenes:
> The mathcomp.ssreflect prefix would point to plugins/ssr

Ok, now I understand. There will still be an incompatibility for, e.g. `Require Import mathcomp.ssreflect.finfun` and all other files there. But it was recommended to do `From mathcomp Require Import finfun`, and that should still work.

Where do you think I should add this binding? Just as an ad-hoc thing in `coqinit.ml`, like it is done for the `Coq` prefix?

ejgallego:
> But it was recommended to do From mathcomp Require Import finfun, and that should still work.

How would that work now?

maximedenes:
I thought your suggestion was to bind the `mathcomp.ssreflect` prefix, but not the `mathcomp` one.

ejgallego:
Unfortunately my suggestion does require moving the whole of `ssreflect` into Coq with the current state of affairs, what would you bind `mathcomp` to?

So I thought indeed the original plan here was to provide a different version of `ssreflect.v` for 8.6/8.7 . IMHO that would be fine, even if we have the pain to have users to install mathcomp even for just the tactics. Well, now you'll have to provide also different versions of `ssrbool` etc... Pity we still don't have cpp support in coq_makefile.

The only other alternative I see is to move the full `ssreflect` prefix into `plugins/ssr`; this was my original idea so people could just use the thing without any extra effort in a compatible way.

I think other alternatives will require rework of the loadpath mode, or adding cpp support to coq_makefile.

PR #134, opened 2017-08-13
Fix typo in fingroup documentation

PatrickMassot:
A tiny typo in the documentation header of ``fingroup.v`` (but the cited lemma is correct!).


gares:
Thanks!

PR #145, opened 2017-10-19
New packager

CohenCyril:
The purpose of this tool is to generate opam package from the opam descriptions located in each directory


CohenCyril:
@gares any contraindication?

PR #150, opened 2017-10-23
using opam in travis before_install script

CohenCyril:
- simplifies the travis script 
- no caching as discussed in https://github.com/math-comp/math-comp/pull/104

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,8 +3,6 @@ sudo: required
 language: c
 cache:
   apt: true
-  directories:
-  - $HOME/.opam
 addons:
   apt:
     sources:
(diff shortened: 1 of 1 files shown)


ejgallego:
Coq's opam used to run some extra stuff such as the test suite so at the beginning it was unsuitable for the Travis time limits. That may have changed now.

CohenCyril:
You are right, it is still not suitable...

ejgallego:
Note @CohenCyril that Coq will hopefully provide soon some better scripts [written in OCaml!] to manage CI testing of Coq itself and of course libraries, so indeed a sensible plan of action could be to reuse that.

PR #157, opened 2017-11-15
Adds generalizations of theorems relying on injectivity

ybertot:
- f_finv
- finv_f
- fconnect_sym
- iter_order
- iter_finv
- cycle_orbit
- fpath_finv (* I need to sub-theorems for this case. *)

All generalizations are named "subset_..." the existing theorems are now
instances of the generalizations.

### DIFF
--- a/mathcomp/ssreflect/fingraph.v
+++ b/mathcomp/ssreflect/fingraph.v
@@ -464,18 +464,122 @@ Qed.
 
 End Loop.
 
-Hypothesis injf : injective f.
+Section subset_orbit.
 
-Lemma f_finv : cancel finv f.
+Variable S : pred_sort (predPredType T).
+
+Hypothesis stable : {in S, forall x, f x \in S}.
+Hypothesis injf : {in S &, injective f}.
+
+Lemma stable_in_iter : {in S, forall x i, iter i f x \in S}.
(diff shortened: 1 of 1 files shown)


ggonthier:
These should be useful generalisations. Three quick observations:
  - the naming convention for relativised lemmas is an `_in` suffix, see `seq.v` or `fintype.v` for instances.
  - the generic predicate parameter `S` should be declared with the `pred_class` idiom, as this can make it easier to instantiate it explicitly.
  - `ssrbool.v` provides helper lemmas (`in1T` and such) to derive unbounded lemmas from relativised ones.

mathcomp/ssreflect/fingraph.v
```diff
@@ -464,18 +464,122 @@ Qed.
...
+Lemma finv_inj_in : {in S &, injective finv}.
+Proof.
+by move => x y sx sy q; rewrite -(f_finv_in sx) q (f_finv_in sy).
+Qed.
+
+Lemma fconnect_sym_in :
+  {in S &, forall x y, fconnect f x y = fconnect f y x}.
+Proof.
+suff Sf x y (sx : x \in S) (sy : y \in S): fconnect f x y -> fconnect f y x.
+  by move => x y sx xy; apply/idP/idP; auto.
```
CohenCyril: remove auto

PR #160, opened 2017-11-22
Add addrKA and subrKA (addrK and addrNK modulo Associativity)

CohenCyril:
Hesitating on the names, any advice?


maximedenes:
@ybertot Cyril was asking about the naming. What's your take on it?

ybertot:
The naming seems reasonable to me.  There is no clash in the library (the KA combination has not been used before in lemma names) and the patterns used in the lemma come often in my own experience.

ggonthier:
`subrKA` could be generalised slightly by changing `- y` to `+ y` in the statement.

PR #162, opened 2017-11-24
Draft contributing guide

CohenCyril:

PR #163, opened 2017-11-24
[standby] Linters

CohenCyril:
Experimental linters, depends on https://github.com/math-comp/math-comp/pull/162


hivert:
I like the typo saying that the script is written in "Qed" instead of "sed" ;-) !

amahboubi:
Hello @CohenCyril, what is the status of this PR?

CohenCyril:
@amahboubi the scripts should be modified to make sure the problems of #164 do not happen again.

PR #41, opened 2016-05-05
Missing lemmas in seq

hivert:
Dear Math-Comp developpers,

During my development in combinatorics, I've been needing a bunch of lemmas about seq. I've selected those which seems to me of general interest. Here are they. 

Best,

Florent

PR #166, opened 2017-12-11
Resubmitted missing lemma in seq.v

hivert:
See discussion on (https://github.com/math-comp/math-comp/pull/41)

### DIFF
--- a/mathcomp/ssreflect/seq.v
+++ b/mathcomp/ssreflect/seq.v
@@ -134,6 +134,13 @@ Require Import ssrfun ssrbool eqtype ssrnat.
 (*                      [:: x_(r1 + 1); ...; x_(r0 + r1)];                    *)
 (*                      ...;                                                  *)
 (*                      [:: x_(r1 + ... + r(k-1) + 1); ...; x_(r0 + ... rk)]] *)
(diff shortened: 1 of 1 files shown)


hivert:
There are some trivial conflicts, but since I Messed things up by trying to resolve them using github interface, I leave it to you to do the merge.

CohenCyril:
I'm on it

hivert:
@CohenCyril: Thanks. One question: I'm ready to resubmit the removed "reshape_index_leq" lemma, however if I do so, you'll have all the other from this pull request too. It doesn't really makes sense to strip the code to what is needed for this single lemma. Should I wait for this pull request being integrated ?

hivert:
One final last comment: now that I have a non trivial contribution to math-comp, I'd be glad to have my name written somewhere in a contributor section ;-) Yes I know: those reseacher/developper with their ego...

CohenCyril:
Good point, until the next release your name is at least in a commit so github automatically lists you as a contributor (https://github.com/math-comp/math-comp/graphs/contributors). Now, when a tarball is made for a release, we have indeed to make sure to have contributor section.

PR #167, opened 2017-12-12
Resubmitted lemma reshape_index_leq for discussion

hivert:
in https://github.com/math-comp/math-comp/pull/41, @CohenCyril requested me to remove this lemma from the pull request and resubmit it, because it needed some discussion about the statement. Here it is.

### DIFF
--- a/mathcomp/ssreflect/seq.v
+++ b/mathcomp/ssreflect/seq.v
@@ -2623,6 +2623,23 @@ elim: ss => //= s ss IHss in i *; rewrite subn_eq0 nth_cat.
 by have [//|le_s_i] := ltnP; rewrite subnDA subSn /=.
 Qed.
 
+Lemma reshape_index_leq sh i1 i2
+      (r1 := reshape_index sh i1) (c1 := reshape_offset sh i1)
+      (r2 := reshape_index sh i2) (c2 := reshape_offset sh i2) :
+  i2 < sumn sh ->
+  (i1 <= i2) = ((r1 < r2) || ((r1 == r2 ) && (c1 <= c2))).
+Proof.
+rewrite {}/r1 {}/c1 {}/r2 {}/c2 /reshape_offset /reshape_index.
+elim: sh i1 i2 => [i1 i2 |] //= s0 s IHs i1 i2.
+rewrite !subn_eq0.
+case: (ltnP i1 s0) => Hi1s0 //= ; case: (ltnP i2 s0) => Hi2s0 //=.
+- by rewrite !subn0.
+- by have := leq_trans Hi1s0 Hi2s0 => /ltnW ->.
+- by have := leq_trans Hi2s0 Hi1s0; rewrite ltnNge => /negbTE ->.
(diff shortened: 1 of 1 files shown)


mathcomp/ssreflect/seq.v
```diff
@@ -2623,6 +2623,23 @@ elim: ss => //= s ss IHss in i *; rewrite subn_eq0 nth_cat.
 by have [//|le_s_i] := ltnP; rewrite subnDA subSn /=.
 Qed.
 
+Lemma reshape_index_leq sh i1 i2 :
+  i1 <= i2 -> i2 < sumn sh ->
```
CohenCyril: I guess that if we had the order library in mathcomp, the statement would look like:
(i1 <= i2) = ((reshape_index sh i1, reshape_offset i1) <= (reshape_index sh i2, reshape_offset i2))
where <= is the canonical lexicographic order on pairs, and that there should also be
(flatten_index sh r1 c1 <= flatten_index sh r2 c2) = ((r1, c1) <= (r2, c2))

So in anticipation of that, I think the following variant would be better suited:
(r1 := reshape_index sh i1) (c1 := reshape_offset sh i1) (r2 := reshape_index sh i2) (c2 := reshape_offset sh i2)
(i1 <= i2) = ((r1 < r2) || ((r1 == r2 ) && (c1 <= c2))
and/or the following
(flatten_index sh r1 c1 <= flatten_index sh r2 c2) = ((r1 < r2) || ((r1 ==r2 ) && (c1 <= c2))

CohenCyril:
Dear Florent, can you please refactor your pull request and squash all commits into only one?

hivert:
@CohenCyril: Once again, I'm in trouble with github workflow. I understood that **you** where supposed to do it when merging the pull request: https://github.com/blog/2141-squash-your-commits
Now I'm not sure about what I should do. If I squash it on my computer, rewriting the history and then do a `push --force` on my PR2 branch on github, will it correctly update the pull request ? Or am I supposed to do something else ?

mathcomp/ssreflect/seq.v
hivert: I think I'll prefer the first. I'll give a try with my code and get back to you.

CohenCyril:
Yes Florent, you are supposed to push --force (and that's another reason why you should always create a new branch before making a pull request)

maximedenes:
> If I squash it on my computer, rewriting the history and then do a push --force on my PR2 branch on github, will it correctly update the pull request ?

Yes it will. PRs are just pointers to branches. When the branch is updated, so is the PR.

hivert:
Git is not my friend today. The squash went all wrong generating lots of conflicts. I ended up creating a brand new branch an coping by hand the file from one branch to the other. I'm starting do dislike GH workflow... 

Anyway, the lemma should be here in a proper shape. I also checked that it is usable with the current statement (https://github.com/hivert/Coq-Combi/commit/d1f5e4db4182ec43423be02cee8f7a666f15028d). Finally, you may want to check if there is a simpler proof.

PR #169, opened 2017-12-12
div_eq1 lemma

hivert:
I had trouble with this one. I'm wondering if there is a simpler proof or even if there is a one line script which makes it redundant. If not it's definitely a natural statement.


CohenCyril:
Indeed I happened to lack such a statement at some point in my life.

CohenCyril:
Thanks for bring this to my attention again, I wrote a piece of code to generalize and subsume what you did.

PR #170, opened 2017-12-12
Using x * y = 1 and x / y = 1 to derive invertibility and the inverse

CohenCyril:
fixes #169 in a more general way

### DIFF
--- a/mathcomp/algebra/ssralg.v
+++ b/mathcomp/algebra/ssralg.v
@@ -738,8 +738,10 @@ Proof. by rewrite opprD addrA addrK. Qed.
 Lemma subrKA z x y : (x - z) + (z + y) = x + y.
 Proof. by rewrite addrA addrNK. Qed.
 
-Lemma subr0_eq x y : x - y = 0 -> x = y.
-Proof. by rewrite -(subrr y) => /addIr. Qed.
+Lemma addr0_eq x y : x + y = 0 -> - x = y.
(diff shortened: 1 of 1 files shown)


ggonthier:
I'm afraid this is too much of a generalisation of #169, which provided a backchaining lemma in the vein of `subr0_eq`. These lemmas can't be used this way: one would have to formulate explicitly the `x / y = 1` subgoal, which could be inconvenient if `x` or `y` are large. Plus, I don't see the point of the rewrite multirule formulation, since either `x` or `y` can't be inferred from the left or right hand side. Finally, switching the equation sides is unnatural (the `y = ...` conclusions).
  So I'd suggest sticking to the `subr0_eq` model, with perhaps a `mulr1_eq` lemma defining the inverse of `x`; perhaps we should also add an `addr0_eq` lemma for consistency?

```
Lemma mulr1_eq x y : x * y = 1 -> x^-1 = y.
Proof.
by move=> xy_eq1; rewrite -[LHS]mulr1 -xy_eq1; apply/mulKr/unitrPr; exists y.
Qed.

Lemma divr1_eq x y : x / y = 1 -> x = y. Proof. by move/mulr1_eq/invr_inj. Qed.
```

CohenCyril:
I agree this makes much more sense for backchaining. I updated my pull request accordingly.
We can merge when it compiles.
(My original intent was not to forget that x was a unit while using this view in a forward chain. Now the user needs to make the duplication himself if he needs to, which is ok I guess.)

PR #171, opened 2017-12-12
The spaces generated by some delta_mx are in a direct sum

CohenCyril:
proof by @ggonthier


CohenCyril:
@ggonthier you are right, this looked weird I put `F I P` in the section context. I also did a small generalization by requiring injectivity only in `P` (as in `mxdirect_sum_eigenspace`)

PR #174, opened 2017-12-18
Enable parallel build, fix #88

### LINKED ISSUE #88: make -j X option lost
MAkefile.coq does not receive -j

JasonGross:
Apparently, the jobserver flags don't show up until we're actually
executing a rule.  So rather than save the old flags, we instead record
whether or not we've seen -B before adding it (actually, whether we've
seen -B or B, since it seems that make stores B rather than -B in
MAKEFLAGS?), and then, later, we filter out B and -B only if we didn't
see them before we added them.

### DIFF
--- a/mathcomp/Makefile
+++ b/mathcomp/Makefile
@@ -15,7 +15,9 @@ else
 COQDEP=$(COQBIN)coqdep
 endif
 
-OLD_MAKEFLAGS:=$(MAKEFLAGS)
+HAS_B:=$(filter -B B,$(MAKEFLAGS))
+# OLD_MAKEFLAGS must use =, not :=, to capture MAKEFLAGS at call time
+OLD_MAKEFLAGS=$(if $(HAS_B),$(MAKEFLAGS),$(filter-out -B B,$(MAKEFLAGS)))
 MAKEFLAGS+=-B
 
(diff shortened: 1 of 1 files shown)


CohenCyril:
This looks useful, but I am unable to understand the implications of this pull request. @gares can you give your opinion?

amahboubi:
Hi @JasonGross, could you help us assessing this pull request? I not the most qualified dev on this kind of stuff, but  I just do not understand why you need this.

JasonGross:
In the present state of the repo, you do `OLD_MAKEFLAGS:=$(MAKEFLAGS)`, which gets evaluated once in the global context.  Apparently, `make` does not put jobserver flags (used for parallel make) in `MAKEFLAGS` in the global context, but only when you are executing a rule, and so when you pass the makeflags on to `MAKEFLAGS="$(OLD_MAKEFLAGS)" $(MAKE) -f Makefile.coq`, you lose the parallel build.  This PR uses `OLD_MAKEFLAGS=` instead of `OLD_MAKEFLAGS:=`, which means that the body of `OLD_MAKEFLAGS` gets evaluated at use-site instead of at definition site, and so the jobserver flags, which only show up in `MAKEFLAGS` in the context of executing a rule, do get picked up appropriately.

PR #177, opened 2018-02-06
add 3 tests to Make

gares:


gares:
@CohenCyril it now compiles

PR #178, opened 2018-02-16
generalize nth_iota

thery:


CohenCyril:
@ybertot: I think it hardly happens, but I guess it may if some induction hypothesis is generalized in your proof. I am very curious to see Laurent's use case though :)

Anyway, this generalization introduces no overhead when the rewrite rule is used from left to right, a minor overhead when used from right to left, so I am in favor of merging.

amahboubi:
Hi @thery ! Any response to @CohenCyril, who was asking about your use case? Anyway, I am merging.

thery:
@amahboubi  sorry but when looking at my usage with @ybertot, we realize that the proof could be fixed and use the nth 0 version. Anyway the generalization seems natural to me.

Issue #90, opened 2017-01-19
Fix warnings


gares:
8.6 generates many warnings, we should review them and fix/silence them

gares:
In particular: warning for redundant coercions

maximedenes:
I'll start by fixing the warnings about the deprecated `Local` syntax, because of the upcoming https://github.com/coq/coq/pull/1049

ejgallego:
@gares couldn't at least disable `deprecated-arguments-scope` and `deprecated-implicit-arguments` for now?

The build log is math-comp is truly unreadable...

CohenCyril:
@maximedenes what's the status of Coq about:
- silencing warnings from math-comp during the import of mathcomp libraries in other libraries
- silencing warnings like `Notation [...] was already used in scope nat_scope.` when we made the overloading on purpose?

anton-trunov:
> silencing warnings like Notation [...] was already used in scope nat_scope. when we made the overloading on purpose?

@CohenCyril Adding the following line in `Make` files silences the notation overloading warnings at least during compilation of math-comp itself:
```
-arg -w -arg -notation-overridden
```

Would it be a good idea to silence the warnings for math-comp and then if there is no solution to silence the warnings transitively, each project built on top of math-comp would decide what works best for them?

PR #179, opened 2018-02-22
Deprecate Implicit Arguments command

### LINKED ISSUE #93: Switch to Arguments syntax for implicit arguments declaration

jashug:
Makes progress on #90. See also [coq/6802](https://github.com/coq/coq/pull/6802)

I think this removes all uses of the `Implicit Arguments` command.
In most cases, this was just removing `Implicit` from the beginning, with occational tweaks for maximal implicits, non implicit arguments between implicit arguments, and reordered implicit argument names.

A side note: I saw a lot of times when a `reflect` was proven, and immediately followed by an Arguments command setting all arguments implicit (which became maximal since there are no arguments to follow). It would be possible to achieve the same effect using `{x}` binders in the statement of the lemma.

Fixes #93.

### DIFF
--- a/mathcomp/algebra/finalg.v
+++ b/mathcomp/algebra/finalg.v
@@ -62,8 +62,7 @@ Definition gen_pack T :=
 
 End Generic.
 
-Implicit Arguments
-  gen_pack [type base_type class_of base_of to_choice base_sort].
+Arguments gen_pack [type base_type class_of base_of to_choice base_sort].
 Local Notation fin_ c := (@Finite.Class _ c c).
 Local Notation do_pack pack T := (pack T _ _ id _ _ id).
 Import GRing.Theory.
--- a/mathcomp/algebra/intdiv.v
+++ b/mathcomp/algebra/intdiv.v
@@ -186,11 +186,11 @@ rewrite {1}(divz_eq m d) mulrDr mulrCA divzMDl ?mulf_neq0 ?gtr_eqF // addrC.
 rewrite divz_small ?add0r // PoszM pmulr_rge0 ?modz_ge0 ?gtr_eqF //=.
 by rewrite ltr_pmul2l ?ltz_pmod.
 Qed.
-Implicit Arguments divzMpl [p m d].
+Arguments divzMpl [p m d].
 
 Lemma divzMpr p m d : p > 0 -> (m * p %/ (d * p) = m %/ d)%Z.
 Proof. by move=> p_gt0; rewrite -!(mulrC p) divzMpl. Qed.
-Implicit Arguments divzMpr [p m d].
+Arguments divzMpr [p m d].
 
 Lemma lez_floor m d : d != 0 -> (m %/ d)%Z * d <= m.
 Proof. by rewrite -subr_ge0; apply: modz_ge0. Qed.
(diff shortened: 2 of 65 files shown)


jashug:
Something went wrong on v8.5 on Travis: https://travis-ci.org/math-comp/math-comp/jobs/344736906

jashug:
The issue on v8.5 appears to be that, in that version, trailing arguments need to be included, so for a function like `Definition foo x y := t`, `Arguments foo [x]` works in 8.6- but fails in 8.5, where you need to write `Arguments foo [x] y`. I'll go through and add the trailing arguments.

maximedenes:
I would suggest you wait a bit before doing that. There are many incompatibilities related to `Arguments` between 8.5 and 8.6, and mathcomp developers may want to reconsider 8.5 support now what 8.8 is being released. At least, it's worth discussing.

gares:
Many thanks for the patch!

I too agree with @maximedenes that v8.5 support is not a major concern. We already support 8.6 and 8.7, so the "rule of the two versions" is respected IMO.

I'll give another spin for 8.6 that did timeout.

gares:
> I saw a lot of times when a reflect was proven, and immediately followed by an Arguments command setting all arguments implicit (which became maximal since there are no arguments to follow). It would be possible to achieve the same effect using {x} binders in the statement of the lemma.

I'm not against it, but please not in this PR.

Actually, I don't know why `{ binder : type }` is not used in this library. Maybe it is just because it is a recent addition. @ggonthier do you recall any issue with this Coq feature?

ggonthier:
Indeed, `{arg}` was too recent a feature to be used when most of the library was written. I'm somewhat reluctant to rely on it for the `reflectP`-style lemmas, because of the risk of inconsistencies: it often happens that some of the arguments to reflection lemmas are section arguments, in which case they need to be made explicitly implicit (sic) after the section. As this may change during refactoring, I think it is sound practice to always have an `Arguments` declaration for each reflection lemma. Similarly I feel an `Arguments` declaration should serve as documentation of the parameters of a constant: there should only be one such declaration in the source code, and it should list explicitly all the parameters and not merely amend the original definition.

jashug:
I have no intention of changing the content of this PR now that Travis is happy with 8.6.

Having the Arguments declaration act as documentation makes sense to me. My comment was really a side note, after seeing the pattern (what felt like) ~50 times, I just had to mention it.

In the spirit of acting as documentation though, when `foo` takes exactly 3 arguments, `Arguments foo [x y z].` would perhaps be better as `Arguments foo {x y z}.`, since implicits at the end automatically get promoted to maximal. That way it is clear that `foo` means `@foo _ _ _` not `@foo`, which is what `[x y z]` would normally suggest to me.
(The same comment applies to the old style `Implicit Arguments foo [x y z].` vs. `Implicit Arguments foo [[x] [y] [z]].` I preserved whichever was used in this PR.)

ggonthier:
An implicit declaration on a non-dependent argument is actually treated as a maximal implicit declaration, and this behaviour actually predates the implementation of maximal implicits. This was how mathcomp got around the Coq dev's initial reluctance to support maximal implicits. We just continued using the coding pattern even after maximal implicits were introduced (partly, I'll admit, because I found the `[[x] [y]]`syntax a bit clunky).

gares:
I've disables travis on 8.5, I guess we are good to go

Issue #63, opened 2016-09-01
ssrmatching.v & Coq 8.6 : what about adding a "transitional" ssrmatching.v file?


erikmd:
Dear MathComp developers,
Given that ssrmatching.v will be provided with Coq 8.6, assuming someone wants to distribute a library Foo that is based on ssrmatching.v _and_ MathComp, and that is compilable as-is with Coq 8.5 and Coq 8.6, do you think that it would make sense to add an extra `ssrmatching.v` file in
https://github.com/math-comp/math-comp/tree/master/mathcomp/ssreflect/plugin/v8.6
that would just do `Require Import ssrmatching.`?
In this case the Foo theories could just start with

```
Require Import mathcomp.ssreflect.ssreflect.
From mathcomp Require Import ssrmatching.
```

and stay compatible with both Coq 8.5 and Coq 8.6 (and Coq 8.4 as well IINM).
What's your opinion on this change?
Kind regards,
Erik

erikmd:
On second thought, adding a quasi-empty file `ssrmatching.v` in MathComp for compatibility is maybe not very elegant... so feel free to close this issue if you consider that this solution is not worth it.

erikmd:
As the latest release 1.6.4 of MathComp is compatible with Coq >= 8.5, this question may be still topical.

Actually I need to refine a bit what I had suggested in the initial description:
- the plan would be to add some file `ssrmatching.v` containing `Require Export ssrmatching.` in folder `mathcomp/ssreflect` (not in folder `mathcomp/ssreflect/plugin/v8.6`)
- for Coq 8.4 and 8.5, this quasi-empty file `ssrmatching.v` would be symlinked/replaced with the file `ssrmatching.v` taken from the appropriate `plugin` subfolder thanks to [Makefile.coq-makefile](https://github.com/math-comp/math-comp/blob/master/mathcomp/ssreflect/Makefile.coq-makefile).

Admittedly it would be a not-very-elegant solution for compatibility, but for example it would facilitate the compilation of CoqEAL (which relies on MathComp and `ssrmatching`) with Coq 8.5 + Coq 8.6 etc., keeping the same Coq theories for any version of Coq and needing no extra configure code in CoqEAL.

@gares @maximedenes @CohenCyril, do you have an opinion on this suggestion?

erikmd:
Hi math-comp devs,
Would you have an objection to the solution I had proposed in this issue?
[...]

PR #180, opened 2018-02-25
Add ssrmatching.v transitional file

erikmd:
Dear math-comp devs,

As suggested in #63 I propose to add this version of file `ssrmatching.v` directly in folder `ssreflect`.

The content of this file is similar to that of [`ssrfun.v`](https://github.com/math-comp/math-comp/blob/master/mathcomp/ssreflect/ssrfun.v) and aims to increase compatibility with Coq 8.6+ for third-party libraries that depend on both math-comp and `ssrmatching.v`

### DIFF
--- a/mathcomp/Make
+++ b/mathcomp/Make
@@ -130,6 +130,7 @@ ssreflect/ssreflect.v
 ssreflect/ssrfun.v
 ssreflect/ssrnat.v
 ssreflect/ssrnotations.v
+ssreflect/ssrmatching.v
 ssreflect/tuple.v
 ssrtest/absevarprop.v
 ssrtest/abstract_var2.v
--- a/mathcomp/ssreflect/Make
+++ b/mathcomp/ssreflect/Make
@@ -18,6 +18,7 @@ path.v
 prime.v
 tuple.v
 ssrnotations.v
+ssrmatching.v
 
 -I .
 -R . mathcomp.ssreflect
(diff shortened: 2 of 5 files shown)


CohenCyril:
@erikmd thanks for the PR, however travis failed on every single version of Coq, with which set of packages does it compile on your computer?

erikmd:
@CohenCyril before pushing my commit I had successfully compiled the ssreflect package (with Coq 8.6.1 and 8.5.3) by following the commands mentioned in the opam file, basically: `git reset --hard && git clean -d -x -f` then `cd mathcomp/ssreflect && make && make install`. So indeed I'm puzzled to see that the command `make -f Makefile.coq -j ${NJOBS} all` run in the root folder by Travis CI systematically fails... maybe I forgot to update something? I'll try to investigate, but feel free to react if you've an idea where this configuration issue could come from! :)

erikmd:
Actually I think I just forgot to update [the global Make file](https://github.com/math-comp/math-comp/blob/master/mathcomp/Make) ; I'll test this and keep you informed

erikmd:
After fixing this, the test build passed with Coq 8.6.1 on my PC. Finally I also updated `ssreflect/plugin/v8.6/ssreflect.v` to directly rely on `ssrmatching` from Coq and thus "avoid one indirection".
(I'll rebase and force-push just now to make Travis do more tests.)

erikmd:
The Travis build passed except for 8.5, because `ssrmatching.v` was copied in the wrong folder... Now it should be OK :)

PR #183, opened 2018-03-02
Add a link to the book home-page

ybertot:


amahboubi:
Hi Yves, good idea to have this link! I am tempted to merge this seemingly innocuous PR but  I don't understand why Travis fails here (why is Travis concerned at all with this file?). @gares, any idea?

gares:
It's a random failure (it happens all the times in Coq too).
Travis runs stuff like `apt-get update` or `rake` in this case... if the network is down, or God know what, everything fails. 

Also, travis checks the library, not the homepage, so I'll just merge.

ejgallego:
The reason for the failure is that the PR is based on the `gh-pages` branch that has no `.travis.yml` file.

PR #185, opened 2018-03-05
Change deprecated Arguments Scope to Arguments

### LINKED ISSUE #184: Use Arguments to declare scopes
Continuing after #93 and #179, I'll work on changing the deprecated `Arguments Scope` to `Arguments`. Travis suggests about 300 occurrences. Helps work towards #90.

jashug:
Closes #184 Ref. coq/coq#6802

I believe this PR preserves the current behavior.

Some notes I took while making the changes:

* In `algebra/poly.v` I removed a trailing empty scope from `coefp_head`.
* The following scopes didn't seem to have a bound key, so I used the scope name:
  + In `fingroup/presentation.v`, scopes `group_presentation` and `nt_group_presentation`
  + In `algebra/ssrint.v`, scope `distn_scope`
  + In `solvable/gseries.v`, scope `group_rel_scope`
  + In `odd_order/PFsection3.v`, scope `defclause_scope`
* In `fingroup/action.v`, the scopes for `act_morph` seem off (The section variable `D` is dropped, but seems to be counted in the scope list).
* In `fingroup/gproduct.v`, three other products get scopes, but `direct_product` doesn't.
* In `algebra/mxalgebra.v`, `has_mxring_id` seems to have an extra unused scope.
* In `character/mxrepresentation.v`, the scopes for `classg_base` and `irrType` seem a bit off.

In case you wanted to take a closer look at any of the above in a separate PR.

I'll also note that using the arguments command just to specify scopes doesn't change implicit argument status, so even after `Arguments foo _%a _%b` `foo` might have some implicit arguments. As documentation, that is perhaps somewhat confusing.

### DIFF
--- a/mathcomp/algebra/matrix.v
+++ b/mathcomp/algebra/matrix.v
@@ -2660,7 +2660,7 @@ Coercion GLval ph (u : GLtype ph) : 'M[R]_n.-1.+1 :=
 End FinUnitMatrix.
 
 Bind Scope group_scope with GLtype.
-Arguments Scope GLval [nat_scope _ _ group_scope].
+Arguments GLval _%N _ _ _%g.
 Prenex Implicits GLval.
 
 Notation "{ ''GL_' n [ R ] }" := (GLtype n (Phant R))
--- a/mathcomp/algebra/mxalgebra.v
+++ b/mathcomp/algebra/mxalgebra.v
@@ -216,9 +216,9 @@ Definition pinvmx : 'M_(n, m) :=
 
 End Defs.
 
-Arguments Scope mxrank [nat_scope nat_scope matrix_set_scope].
+Arguments mxrank _%N _%N _%MS.
 Local Notation "\rank A" := (mxrank A) : nat_scope.
-Arguments Scope complmx [nat_scope nat_scope matrix_set_scope].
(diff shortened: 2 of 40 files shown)


gares:
> I'll also note that using the arguments command just to specify scopes doesn't change implicit argument status, so even after Arguments foo _%a _%b foo might have some implicit arguments. As documentation, that is perhaps somewhat confusing.

Yep I know, it is a choice I made a long ago. Since then many found it surprising, but I'm not sure the other option (unset all options that are set) would make the command more usable (I agree it will make it more "documenting").
You can append a `: clear implicits` if you want to signal the constant has no implicit arguments. Did you see many of these?

gares:
I'm asking Assia and Cyril to review the "odd cases" you listed.

BTW, many thanks for the PR!

jashug:
In most cases, I didn't look at whether there were implicit arguments or not, so I'm not sure how commonly you have some/no implicit arguments.

jashug:
@gares @amahboubi @CohenCyril ping

gares:
I think there are no reasons not to merge this now.
The list of odd cases is there, and if we find out something can be improved, we can do it later.

Thanks again for your patch!

PR #192, opened 2018-04-17
move odd_order to its own repository

gares:
To be merged only after https://github.com/math-comp/odd-order stabilizes (eg CI works)


gares:
It builds locally but times out on travis...

CohenCyril:
Travis will not succeed because the current version of odd-order relies on the development version of math-comp, and not any released version (the same goes for real-closed). I guess we should merge the few waiting PRs and release a math-comp 1.7~beta version, so that we can test real-closed and odd-order against it.

PR #193, opened 2018-04-17
Removing undocumented compatibility module

CohenCyril:
I had put this for compatibility with mathcomp 1.6 when we were still using svn, but I am afraid it got under the radar.
We should decide
- if we revert the change of `ltngtP`,
- if we document (and extend) the compatibility module
- or if we just remove the module and keep the change to `ltngtP`

I am personally in favour of the last

### DIFF
--- a/mathcomp/ssreflect/ssrnat.v
+++ b/mathcomp/ssreflect/ssrnat.v
@@ -1596,15 +1596,3 @@ Ltac nat_congr := first
      apply: (congr1 (addn X1) _);
      symmetry
    end ].
-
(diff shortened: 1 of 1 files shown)


amahboubi:
I am not in favour of a compatibility module. The other instances of similar spec inductives, in ssrnum, share the same structure, with the two symmetrical versions of the equalities so I guess that your last option is just fine. However, I must confess that this change and the subsequent compat module went under my radar as well (as many things which happened in summer 2016 it seems) so I would be glad to hear other voices.

PR #201, opened 2018-07-03
Generalizing homo-mono-morphism lemmas and extremum

CohenCyril:
TODO after @amahboubi's review:
- [x] add extremum to header of `fintype.v` and explain how this was generalized in `ChangeLog`
- [x] rename all mentions to "homomorphism" or "monomorphism" to "order" or "monotonicity" in `path.v`, `ssrnat.v` and `ChangeLog`
- [x] comment the code to explain why so many `Let` in `ssrnat.v` and `ssrnum.v`

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -5,6 +5,39 @@
 
 	* Added companion matrix of a polynomial `companionmx p` and the
 	  theorems: companionmxK, map_mx_companion and companion_map_poly
+	* Extended generic theory about homomorphisms and monomophismes.
+	* Added: homoW_in, inj_homo_in, mono_inj_in, anti_mono_in,
+	  total_homo_mono_in, homoW, inj_homo, monoj, anti_mono,
+	  total_homo_mono.
+	* Extended theory about homo and mono for leq/or and Num.le
+	* Renamed: (together with the _in suffix counterpart)
+	  mono_inj -> incr_inj
+	  nmono_inj -> decr_inj
+	  leq_mono_inj -> incnr_inj
+	  leq_nmono_inj -> decnr_inj
+	  homo_inj_ltn_lt -> incnr_inj
+	  nhomo_inj_ltn_lt -> decnr_inj
+	  homo_inj_in_lt -> inj_homo_ltr_in
+	  nhomo_inj_in_lt -> inj_nhomo_ltr_in
+	  ltn_ltrW_homo -> ltnrW_homo
+	  ltn_ltrW_nhomo -> ltnrW_nhomo
+	  leq_lerW_mono -> lenrW_mono
+	  leq_lerW_nmono -> lenrW_nmono
+	  homo_leq_mono -> lenr_mono
+	  nhomo_leq_mono -> lenr_nmono
+	  homo_inj_lt -> inj_homo_ltr
+	  nhomo_inj_lt -> inj_nhomo_ltr
+	  homo_inj_ltn_lt -> inj_homo_ltnr
+	  nhomo_inj_ltn_lt -> inj_nhomo_ltnr
+	  homo_mono -> ler_mono
+	  nhomo_mono -> ler_nmono
+	* Added sorted_lt_nth, ltn_index, sorted_le_nth, leq_index.
+	* Added: [arg minr_( i < n | P ) F] and [arg maxr_( i < n | P ) F]
+	  with all their variants, following the same convention as for nat
+	* Generalized extremum_spec and its theory, added generic extremumP.
+	* Added missing contra* lemma: contra_neqN, contra_neqF,
+	  contra_neqT, contra_neq_eq, contra_eq_neq.
+	* Addded missing seq theorems: take_subseq, drop_subseq
 
 	* Reshuffled theorems inside files and packages:
 	  * countalg goes from the field to the algebra package
--- a/ChangeLog
+++ b/ChangeLog
@@ -4,12 +4,16 @@
 	  optimization is not strictly necessary with modern Coq versions.
 
 	* Added companion matrix of a polynomial `companionmx p` and the
-	  theorems: companionmxK, map_mx_companion and companion_map_poly
-	* Extended generic theory about homomorphisms and monomophismes.
+	  theorems: companionmxK, map_mx_companion and companion_map_poly.
+
+	* Extended generic theory about monotonocity (*mono* and *homo* lemmas).
+
 	* Added: homoW_in, inj_homo_in, mono_inj_in, anti_mono_in,
 	  total_homo_mono_in, homoW, inj_homo, monoj, anti_mono,
 	  total_homo_mono.
+
 	* Extended theory about homo and mono for leq/or and Num.le
+
 	* Renamed: (together with the _in suffix counterpart)
 	  mono_inj -> incr_inj
 	  nmono_inj -> decr_inj
@@ -31,12 +35,20 @@
 	  nhomo_inj_ltn_lt -> inj_nhomo_ltnr
 	  homo_mono -> ler_mono
 	  nhomo_mono -> ler_nmono
+
 	* Added sorted_lt_nth, ltn_index, sorted_le_nth, leq_index.
-	* Added: [arg minr_( i < n | P ) F] and [arg maxr_( i < n | P ) F]
-	  with all their variants, following the same convention as for nat
-	* Generalized extremum_spec and its theory, added generic extremumP.
+
+	* Generalized extremum_spec and its theory, added extremum and
+	  extremumP, generalizing arg_min for an arbitrary eqType with an
+	  order relation on it (rather than nat). Redefined arg_min and
+	  arg_max with it.
+
+	* Added [arg minr_( i < n | P ) F] and [arg maxr_( i < n | P ) F]
+	  with all their variants, following the same convention as for nat.
+
 	* Added missing contra* lemma: contra_neqN, contra_neqF,
 	  contra_neqT, contra_neq_eq, contra_eq_neq.
+
 	* Addded missing seq theorems: take_subseq, drop_subseq
 
 	* Reshuffled theorems inside files and packages:
--- a/mathcomp/algebra/ssrnum.v
+++ b/mathcomp/algebra/ssrnum.v
@@ -129,6 +129,16 @@ Require Import bigop ssralg finset fingroup zmodp poly.
 (*   e : exterior = in [1, +oo[ or ]1; +oo[                                   *)
 (*   w : non strict (weak) monotony                                           *)
 (*                                                                            *)
+(* [arg minr_(i < i0 | P) M] == a value of i : T minimizing M : R, subject    *)
+(*                   to the condition P (i may appear in P and M), and        *)
+(*                   provided P holds for i0.                                 *)
+(* [arg maxr_(i > i0 | P) M] == a value of i maximizing M subject to P and    *)
+(*                   provided P holds for i0.                                 *)
+(* [arg minr_(i < i0 in A) M] == an i \in A minimizing M if i0 \in A.         *)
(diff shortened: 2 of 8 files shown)


mathcomp/ssreflect/fintype.v
```diff
@@ -2,7 +2,7 @@
 (* Distributed under the terms of CeCILL-B.                                  *)
 Require Import mathcomp.ssreflect.ssreflect.
 From mathcomp
-Require Import ssrfun ssrbool eqtype ssrnat seq choice.
+Require Import ssrfun ssrbool eqtype ssrnat seq choice path.
 
 (******************************************************************************)
```
amahboubi: Please add an entry for extremum to the header.

mathcomp/ssreflect/path.v
```diff
@@ -891,3 +891,46 @@ End CycleArc.
 
 Prenex Implicits arc.
 
+Section Homomorphism.
```
amahboubi: I am a bit disturbed by the use of "homomorphism" here... I am aware that it is inherited from ssrfun, but I do not understand what is an homomorphism without an algebraic structure. This section is "just" about (almost) "sorted" lists. Changing the notation is a big change, I am just suggesting to rename the section here.

mathcomp/ssreflect/ssrnat.v
```diff
@@ -1375,6 +1381,132 @@ rewrite -[4]/(2 * 2) -mulnA mul2n -addnn sqrnD; apply/leqifP.
 by rewrite ltn_add2r eqn_add2r ltn_neqAle !nat_Cauchy; case: ifP => ->.
 Qed.
 
+Section NatHomomorphism.
```
amahboubi: Similar remark about homomorphisms here.

mathcomp/ssreflect/ssrnat.v
```diff
@@ -1375,6 +1381,132 @@ rewrite -[4]/(2 * 2) -mulnA mul2n -addnn sqrnD; apply/leqifP.
...
+
+Lemma homo_leq (f : nat -> T) (r : T -> T -> Prop) :
+   (forall x, r x x) -> (forall y x z, r x y -> r y z -> r x z) ->
+  (forall i, r (f i) (f i.+1)) -> {homo f : i j / i <= j >-> r i j}.
+Proof. by move=> rrefl /(@homo_leq_in predT f r) fr fS i j; apply: fr. Qed.
+
+Section NatToNat.
+Variable (f : nat -> nat).
+
+Let ltn_neqAle := ltn_neqAle.
```
amahboubi: Why this handful of Let?

ChangeLog
```diff
@@ -1,6 +1,39 @@
 ??/??/???? - version 1.7.1
 	* Added companion matrix of a polynomial `companionmx p` and the
 	  theorems: companionmxK, map_mx_companion and companion_map_poly
+	* Extended generic theory about homomorphisms and monomophismes.
```
amahboubi: I think that this line might be misleading (see below). It is not clear what a "homomorphism" is. And the `mono` in the {mono...} notation does not stand for monomorphism but for monotonic. If I am correct, a monomorphism is usually something else, e.g. an injective homomorphism. Plus in English there is no final 'e'.

mathcomp/algebra/ssrnum.v
```diff
@@ -1265,173 +1275,260 @@ Implicit Types m n p : nat.
 Implicit Types x y z : R.
 Implicit Types u v w : R'.
 
+Let leqnn := leqnn.
```
amahboubi: Why this handful of Let (when the abbreviation and the name of the constant coincide)?

ChangeLog
```diff
@@ -1,6 +1,39 @@
...
+	  homo_inj_lt -> inj_homo_ltr
+	  nhomo_inj_lt -> inj_nhomo_ltr
+	  homo_inj_ltn_lt -> inj_homo_ltnr
+	  nhomo_inj_ltn_lt -> inj_nhomo_ltnr
+	  homo_mono -> ler_mono
+	  nhomo_mono -> ler_nmono
+	* Added sorted_lt_nth, ltn_index, sorted_le_nth, leq_index.
+	* Added: [arg minr_( i < n | P ) F] and [arg maxr_( i < n | P ) F]
+	  with all their variants, following the same convention as for nat
+	* Generalized extremum_spec and its theory, added generic extremumP.
```
amahboubi: Please explain in which direction it is generalized (from nat to eqtypes).

mathcomp/ssreflect/fintype.v
```diff
@@ -971,19 +973,49 @@ apply: contraFF (no_i i) => /andP[Pi /eqP def_n]; rewrite /= Pi.
...
+Hypothesis ord_total : total ord.
+
+Definition extremum := odflt i0 (pick (arg_pred P F ord)).
+
+Hypothesis Pi0 : P i0.
+
+Lemma extremumP : extremum_spec P F ord extremum.
 Proof.
-rewrite /arg_max; case: pickP => [i /andP[Pi /forall_inP/= max_i] | no_i].
+rewrite /extremum; case: pickP => [i /andP[Pi /'forall_implyP/= min_i] | no_i].
```
amahboubi: What is /' doing here? Not a request but a real question :).

mathcomp/algebra/ssrnum.v
```diff
@@ -129,6 +129,16 @@ Require Import bigop ssralg finset fingroup zmodp poly.
 (*   e : exterior = in [1, +oo[ or ]1; +oo[                                   *)
 (*   w : non strict (weak) monotony                                           *)
 (*                                                                            *)
+(* [arg minr_(i < i0 | P) M] == a value of i : T minimizing M : R, subject    *)
```
amahboubi: I would suggest "a value i : T" minimizing M..."
Also, what does the `arg` keyword refer to?

mathcomp/algebra/ssrnum.v
```diff
@@ -129,6 +129,16 @@ Require Import bigop ssralg finset fingroup zmodp poly.
...
+(*                   to the condition P (i may appear in P and M), and        *)
+(*                   provided P holds for i0.                                 *)
+(* [arg maxr_(i > i0 | P) M] == a value of i maximizing M subject to P and    *)
```
amahboubi: "a value i maximizing M..."

ChangeLog
amahboubi: Sure, but I still think that the wording you added to the ChangeLog is misleading. It is not generic: it is about homo/mono morphism of *ordered* structures. See e.g. [homomorphism](https://en.wikipedia.org/wiki/Homomorphism). E.g. there is no equational theory preserved here. Adding that this is about order would address my comment. 

I have no complain about the name of the constants, and the comments in ssrfun are clear that they are about stability/morphism/homomorphism *for unary and binary relations*.

amahboubi:
I am not sure it is the best place to discuss this but I still think that the litany of lemmas about compatibility with order in `ssrnum` is a nasty issue. For instance I rather often have to use that:
``x >= 0 -> y <= z ->  y * x <= z * x``.  I have never managed to remember that the name of this lemma is ``ler_wpmul2r``, and its statement is: 
```Lemma ler_wpmul2r x : 0 <= x -> {homo *%R^~ x : y z / y <= z}.```
which is brown belt ssr-style notation (mind the ``%R^~ x``), next to impossible to `Search` and quite hard to teach to a newbie.
It would be awesome to find a way to improve this...

amahboubi:
Here is another difficult one, this time in `ssrint`:

``About ler_piexpz2l.

ler_piexpz2l :
forall (R : realFieldType) (x : R),
(0 < x)%R ->
(x < 1)%R ->
{in (>= 0)%R &, {mono exprz (R:=R) x : y x2 / (x2 <= y)%R >-> 
(y <= x2)%R}}

Expanded type for implicit arguments
ler_piexpz2l :
forall (R : realFieldType) (x : R),
(0 < x)%R ->
(x < 1)%R ->
forall x2 y : int_numDomainType,
x2 \in (>= 0)%R ->
y \in (>= 0)%R ->
(fun x3 : R => [eta (>= x3)%R]) (x ^ x2)%R (x ^ y)%R =
(fun y0 : int => <=%R%R^~ y0) x2 y``

ChangeLog
CohenCyril: > it is about homo/mono morphism of ordered structures

oh, right!

amahboubi:
Another caveat: `{homo ... }` lemmas like the aforementionned `ler_wpml2r` should be `apply`ed, but the analogue for the strict relation, `ltr_pmul2r` is formulated as a `{mono ..}` predicate, and is a `rewrite` rule.

mathcomp/ssreflect/fintype.v
CohenCyril: ok

mathcomp/ssreflect/ssrnat.v
CohenCyril: This is necessary until we have canonical structures for orders. It allows the subsequent proofs to be oneliners since all the necessary hypotheses are in context and can thus be found by `ssrdone`.

mathcomp/algebra/ssrnum.v
CohenCyril: This is helping `ssrdone` concluding every subsequent proofs. (I definitely should put comments)

mathcomp/ssreflect/fintype.v
CohenCyril: This is a use of the view (`/`) named `'forall_implyP`

amahboubi: Amazing, I did not know this (family of) view(s), or may be I forgot about it. It's been too long since I worked finite objects it seems... Sorry for the noise and thanks for the answer.

CohenCyril: Off-topic: this family proved useful to me and should be complete with `'forall_in_`, etc... variants.
Alternatively/Additionally, @bgregoir and @gares have an experiment on automated views using typeclasses that I would much like to see as a PR here.

CohenCyril:
> Another caveat: `{homo ... }` lemmas like the aforementionned `ler_wpml2r` should be `apply`ed, [...]

This is not completely true, most `{homo ...}` lemma can be rewritten with they are rules of the form `_ <= _ = true` with preconditions...

mathcomp/ssreflect/fintype.v
CohenCyril: @bgregoir @gares do you think you could open a PR about these?

gares: Benjamin did use it for a while, but it found it unsatisfactory (eg see the vlevel business)
The code is here: https://github.com/gares/autoview
Maybe you could try to use it here and see how it goes.

gares: The point is: you have 5 variants (only logical, recursive, under binders, ...) but sometimes none fits, eg they go too deep or too shallow :-/
At least, this was the argument Benjamin communicated me in private.

 For your use case you probably have what you need.

gares: If it works for you I can make a PR and write some doc (the .v file is not too bad).
Now that I think about it, other users were interested.

CohenCyril:
@amahboubi I addressed your requests, should we merge?

mathcomp/ssreflect/fintype.v
amahboubi: HI @CohenCyril, @gares, I am about to merge the present PR, but I would like to keep a pointer on this conversion, so that we keep track of this benchmark for the autoview experiment. What is the best way of doing so?

amahboubi: done.

PR #208, opened 2018-07-28
Companion matrix of a polynomial

CohenCyril:
- [x] Header Documentation
- [x] ChangeLog update


CohenCyril:
> Just a stupid comment:
> is there a reason to define it
> differently than in https://en.wikipedia.org/wiki/Companion_matrix

Yes there is one: the companion matrix should act on row basis vector and not column matrix vectors.
As for why just not choosing the transpose, it's a bit arbitrary.
This one has the property that
```coq
forall i j : 'I_(size p).-1, i <= j -> delta_mx 0 j *m companion_mx p ^ i = delta_mx 0 (j - i) 
```
(and we should prove it and add it to the lib)
while the traditional one can be obtained by conjugation with the matrix_perm of the reversal permutation and the property it would satisfy would be:
```coq
forall i j : 'I_(size p).-1, 0 < i -> i + j <= (size p) -> delta_mx 0 i *m companion_mx p ^ i = delta_mx 0 (i + j) 
```
which maybe looks a bot more complicated...

thery:
Is what you explain relates to the linear recurrence section in the wiki page?
In this case we should take the transpose

CohenCyril:
@thery I changed it

PR #209, opened 2018-07-28
Moving countalg and closed_field around

CohenCyril:
- countalg goes from the field to the algebra package
- finalg now get the expected inheritance from countalg
- closed_field now contains the construction of algebraic closure for countable fields (previously in countalg)
- proof of quantifier elimination for closed field rewritten in a monadic style

TODO:
- [x] update ChangeLog


amahboubi:
HI @CohenCyril. What is the status of this PR? The conflicts seem easy to solve. If I understand correctly, the scopes of the definitions and notations supporting your monadic style are limited to the ClosedFieldQE module, in library closed_field, module which is not imported, right?

gares:
The monadic style seem a great improvement, I cant believe I'm writing this.

Personally I'd prefer making bind and return 2 keywords and use that in place of 'let and ret.

CohenCyril:
> What is the status of this PR?

@amahboubi I've been bad, this PR contains too many things. Now I told @ggonthier and you about the fintype+counttype-refactoring part.

> If I understand correctly, the scopes of the definitions and notations supporting your monadic style are limited to the ClosedFieldQE module, in library closed_field, module which is not imported, right?

Yes

> The conflicts seem easy to solve.

Yes, the so-called conflict are just about rebasing the ChangeLog onto the previous PR that got merged

PR #216, opened 2018-08-07
Small scale tool for proving circular implications, and using any pair of statements as an equivalence

CohenCyril:


CohenCyril:
Depends on #201, please review independently...

CohenCyril:
@gares I could swear I saw a review from you in my notifications about this PR, however I cannot find it here... was I dreaming?

CohenCyril:
> Hi @CohenCyril. This PR has more content than the title suggests. I see several independent natures of things: 1) lemmas in seq and (yet other) variants of contraposition with equalities 2) homo-mono-morphisms 3) generalization of extremas 4) infrastructures for chains of equivalences. I would agree to approve 1) and also may be 3) right away. I have a few questions about 2) and 4).

Hi @amahboubi, this is a depend PR, you can post your comments about 1)2)3) in #201, only comments about 4) may go here.

mathcomp/ssreflect/seq.v
```diff
@@ -2783,3 +2795,42 @@ by apply: inj_f => //; apply/allpairsP; [exists (x, y1) | exists (x, y2)].
 Qed.
 
 End EqAllPairs.
+
+Section AllIff.
+
+Inductive all_iff_and (P Q : Prop) := AllIffConj of P & Q.
```
amahboubi: Annotate with the sort of the inductive.

mathcomp/ssreflect/seq.v
```diff
@@ -2783,3 +2795,42 @@ by apply: inj_f => //; apply/allpairsP; [exists (x, y1) | exists (x, y2)].
...
+Definition all_iff P0 (Ps : seq Prop) : Prop :=
```
amahboubi: Provide the type of the P0 argument.

mathcomp/ssreflect/seq.v
```diff
@@ -2783,3 +2795,42 @@ by apply: inj_f => //; apply/allpairsP; [exists (x, y1) | exists (x, y2)].
...
+  (fix aux (P : Prop) (Qs : seq Prop) : Prop :=
+      if Qs is Q :: Qs then all_iff_and (P -> Q) (aux Q Qs)
+      else P -> P0 : Prop) P0 Ps.
+
+Notation "[ '<->' x0 ; x1 ; .. ; xn ]" := (all_iff x0 (x1 :: .. [:: xn] ..))
+  (at level 0, format "[ '<->' '['  x0 ;  '/' x1 ;  '/'  .. ;  '/'  xn ']' ]")
+  : form_scope.
+
+Lemma all_iffLR P0 Ps : all_iff P0 Ps ->
```
amahboubi: What about having also a version of the lemma which does not rely on indices? Something like
`forall P Q P0 Ps, all_iff P0 Ps -> P \in P0 :: Ps -> Q \in P0 :: Ps -> P -> Q`.

mathcomp/ssreflect/seq.v
```diff
@@ -2783,3 +2795,42 @@ by apply: inj_f => //; apply/allpairsP; [exists (x, y1) | exists (x, y2)].
...
+exact: (P0ton _ \o Pnto0 _).
+Qed.
+
+Lemma all_iffP P0 Ps : all_iff P0 Ps ->
+   forall m n, nth P0 (P0 :: Ps) m <-> nth P0 (P0 :: Ps) n.
+Proof. by move=> /all_iffLR iffPs m n; split => /iffPs. Qed.
+
+End AllIff.
+Arguments all_iffLR {P0 Ps}.
+Coercion all_iffP : all_iff >-> Funclass.
```
amahboubi: I was about to ask wether it is clear that the coercion should be `all_ifP`, as opposed to `all_iffLR`, but I think it's ok.

CohenCyril:
Hi @amahboubi, thanks for your comments.
An example is here: https://github.com/math-comp/finmap/blob/various_improvements/finmap.v#L2964-L2996

mathcomp/ssreflect/seq.v
CohenCyril: That is impossible, since these are lists of `Prop` and `\in` is specialized to discrete types.

mathcomp/ssreflect/seq.v
CohenCyril: I will fix this.

mathcomp/ssreflect/seq.v
CohenCyril: Ok

mathcomp/ssreflect/seq.v
CohenCyril: I also wondered, but I think this is better. (Of course it would even be better if `Hint View` supported this, but it's probably not necessary given the number of usecases...)

CohenCyril:
> I think this is ok, but I would like to see an example of instance before merging. For now, it is a lone piece of infrastructure.

Do you want to see examples inside the mathcomp library? Or is my finmap example enough?

mathcomp/ssreflect/seq.v
amahboubi: Oops, sorry for the noise, I am doing too much classical stuff these days it seems :). Would it make sense to have a boolean version? Interestingly, this morning I found a related thread in the Lean users chat room: 
https://leanprover.zulipchat.com/#narrow/stream/113488-general/topic/TFAE

CohenCyril: @amahboubi thanks for the pointer. As for booleans I do not think it would make sense: since it is a small scale reflexion tactic, the `P \in Ps` should be convertible to true whenever a property is in the list, but is won't ever happen (except for `true` and `false`, which is not very intersting)

amahboubi: I did not mean a boolean version of this particular lemma per se, but more generally a support for boolean TFAE, i.e. a notation for lists of boolean statements that are pairwise equivalent. The expected benefit would not be the computational behaviour but classical reasoning. For instance you can state that they are either all valid, or all wrong.

CohenCyril: Would it a combination of `all_if` together with with `eq_true_iff_eq` to get equalities and with `equivP idP` to get reflects?

CohenCyril:
> May be would it be worth having this exact wording in a comment.

OK

CohenCyril:
BTW, when I have the time, I also want to contribute an extension of this where, when you want to prove the TFAE, you provide a graph, which you check is connected (by dfs, or tarjan :)) and you generate the implications to prove from the arrows of the graph.

amahboubi:
I am wondering: is it the appropriate workflow? I would like both to merge this (once you've implemented the 2 minor changes I requested) and to continue the discussion. It probably depend on when you want this to be merged.

CohenCyril:
When this pull request is merged, we should move the discussion to an issue we create to continue the discussion and reference this PR for the record. Note also that we must not merge this pull request yet since it depends on PR #201

amahboubi:
>  Note also that we must not merge this pull request yet since it depends on PR #201

But the "circular implication" code is in fact independent isn't it?

CohenCyril:
> But the "circular implication" code is in fact independent isn't it?

No, it relies on `homo_ltn`

amahboubi:
This dependency is blocking an easier merge. Are your sure this is not overkill? 
In my opinion the dependency is not essential: without the machinery introduced by #201, `homo_ltn` is an easy 2-liner:

```
Lemma homo_ltn T (f : nat -> T) (r : T -> T -> Prop) :    
  (forall y x z, r x y -> r y z -> r x z) ->    
  (forall i, r (f i) (f i.+1)) -> {homo f : i j / i < j >-> r i j}.    
Proof.    
move=> hr hi x y; elim: y x => // y ihy x; rewrite ltnS leq_eqVlt.    
case/orP=> [/eqP-> // | ltxy]; apply: hr (hi _); exact: ihy.    
Qed.
```

CohenCyril:
> This dependency is blocking an easier merge. Are your sure this is not overkill?

I would not say "Overkill" is the word (because #201 is way more useful to me than this one in my other developments), but it can be made temporarily independent, yes. I rebased, if you are happy with the modifications, you can merge. I'll fix the proof again in €201.

amahboubi:
Agreed, my "overkill" referred to the dependency. Now the PR is much cleaner I think, thanks!

amahboubi:
Last question before merging (once checks have all completed). I am wondering about how to formalize best the fact that proving any item of a TFAE is proving the conjunction of statements. But if you prefer, I can merge first and ask this question again in an issue.

CohenCyril:
> But if you prefer, I can merge first and ask this question again in an issue.

yes please

mathcomp/ssreflect/seq.v
```diff
@@ -153,6 +153,12 @@ Require Import ssrfun ssrbool eqtype ssrnat.
 (*   We are quite systematic in providing lemmas to rewrite any composition   *)
 (* of two operations. "rev", whose simplifications are not natural, is        *)
 (* protected with nosimpl.                                                    *)
+(*  ** The following are alle equivalent:                                     *)
```
amahboubi: alle -> all

mathcomp/ssreflect/seq.v
```diff
@@ -153,6 +153,12 @@ Require Import ssrfun ssrbool eqtype ssrnat.
...
+(*  [<-> P0; P1; ..; Pn] == P0, P1, ..., Pn are all equivalent                *)
+(*                       := P0 -> P1 -> ... -> Pn -> P0                       *)
+(*  if (T : [<-> P0; P1; ..; Pn]) is such an equivalence, and i, j in nat     *)
+(*  then T i j is a proof of the equivalence Pi <-> Pj, between Pi and Pj     *)
+(*  (when i (resp j) is out of bound, Pi (resp Pj) defaults to P0             *)
```
amahboubi: spurious '('

mathcomp/ssreflect/seq.v
```diff
@@ -2783,3 +2789,52 @@ by apply: inj_f => //; apply/allpairsP; [exists (x, y1) | exists (x, y2)].
 Qed.
 
 End EqAllPairs.
+
+Section AllIff.
+(* The Following Are Equivalent *)
+
+(* specific conjunction to tell appart members of *)
```
amahboubi: I do not understand this comment line. What about:
`We introduce a specific conjunction, used to chain the consecutive items in a circular list of implications.`

PR #217, opened 2018-08-07
Adding `allsigs`, the dependent version of `allpairs`

CohenCyril:


ChangeLog
```diff
@@ -1,6 +1,8 @@
 ??/??/???? - version 1.7.1
 	* Added companion matrix of a polynomial `companionmx p` and the
 	  theorems: companionmxK, map_mx_companion and companion_map_poly
+	* Added allsigs, the dependent version of allpairs,
+	  with notation `[seq E | i <- s & j <- t i]`
```
thery: `[seq E | i <- s & j <- t]`

CohenCyril: The phrasing `[seq E | i <- s & j <- t i]` has the advantage to explicit the dependency of `i` in `t`, do you think it is better kept implicit in the doc?

thery: From finset
[set E | x in A, y in B] == the set of values of E for x drawn from A and *)
(*                      and y drawn from B; B may depend on x.

CohenCyril: @thery ok, fixed

thery:
merging before the meeting

PR #218, opened 2018-08-07
removing multiple definitions of [tT]ag*

CohenCyril:
they are already defined in ssrfun


gares:
I think the changes should mention this (maybe someone is using this directly and forgot to import ssrfun).

CohenCyril:
Should it be delayed to release 1.8.0 then?

PR #219, opened 2018-08-08
fix some bugs in Makefile

CohenCyril:


amahboubi:
I do not know what to do with this PR. I am unable to review without at least a small description of the pb.
@gares may be more relevant as a reviewer?

gares:
I'd be happy if the code could call
```
$ ~/COQ/coq/bin/coqtop -print-version
8.9+alpha 4.06.0
```
and "parse" that string, that was made easier to parse (the second version is the ocaml compiler)

CohenCyril:
@gares actually `-print-version` seems to be unavailable for coq 8.6, so the modification you suggest actually implies dropping support...

CohenCyril:
@gares I'd rather push this fix first, and later make another PR taking your remark into account and drop 8.6 when there is a consensus. I left a comment in the `Makefile.common` with the instructions.

gares:
I guess I added --print-version on 8.6, and then the more standard -print-version was also added. Could you try that? (I don't have a 8.6 around)

CohenCyril:
> I guess I added --print-version on 8.6, and then the more standard -print-version was also added. Could you try that? (I don't have a 8.6 around)

With nix, coq 8.6 was around in a few seconds :heart:

gares:
I'm a bit lost. Is the change I propose doable on 8 6 too or not? If not I shall merge as is. Else I should wait for a fix.

CohenCyril:
@gares the fix was made (unless there is more to fix than I thought) and it passed.

mathcomp/Makefile.common
```diff
@@ -64,7 +76,7 @@ install: __always__ Makefile.coq
 # counting lines of Coq code -----------------------------------------
 .PHONY: count
 
-COQFILES := $(shell grep '.v$$' Make)
+COQFILES = $(shell grep '.v$$' $(COQPROJECT))
```
gares: @CohenCyril @SkySkimmer wrt #238 I think this line is fishy. `$(COQPROJECT)` points to `_CoqProject` that, IIRC, only contains `-R` directives but not the list of files.

SkySkimmer: I believe it points to https://github.com/math-comp/math-comp/blob/d6dc5741ba44808e5f2f01a238d972ec2c11737f/mathcomp/fingroup/Make (which does have v files) as per https://github.com/math-comp/math-comp/blob/d6dc5741ba44808e5f2f01a238d972ec2c11737f/mathcomp/fingroup/Makefile#L3

CohenCyril: > [...] as per
> 
> [math-comp/mathcomp/fingroup/Makefile](https://github.com/math-comp/math-comp/blob/d6dc5741ba44808e5f2f01a238d972ec2c11737f/mathcomp/fingroup/Makefile#L3)
> 
> Line 3 in [d6dc574](/math-comp/math-comp/commit/d6dc5741ba44808e5f2f01a238d972ec2c11737f)
> 
>  COQPROJECT="Make"

Yes indeed, but... could the double quotes be guilty here?

SkySkimmer: No, it gets sent to the shell which will remove them.

mathcomp/Makefile.common
```diff
@@ -23,14 +23,26 @@ VERBOSE?=V
...
+COQVV:= $(shell echo $(COQVVV) | cut -d"." -f1-2)
 
 # all: ---------------------------------------------------------------
 all: config build
 
 # Makefile.coq: ------------------------------------------------------
 .PHONY: pre-makefile
 
-Makefile.coq: pre-makefile Makefile
+Makefile.coq: pre-makefile $(COQPROJECT) Makefile
```
SkySkimmer: I think the issue is the double quotes, although they are a problem on this line not the grep line.

PR #223, opened 2018-09-03
[warnings] -w "+compatibility-notation" clean

### LINKED ISSUE #222: Please make your project compile with -w "+compatibility-notation"
Please make your project compile with `-w "+compatibility-notation"` to prepare for the Coq 8.9 release. We would appreciate if this could be done ASAP. If you have trouble doing so, please raise your concern at coq/coq#8383.

gares:
Fix #222

### DIFF
--- a/mathcomp/field/algnum.v
+++ b/mathcomp/field/algnum.v
@@ -354,10 +354,10 @@ Lemma extend_algC_subfield_aut (Qs : fieldExtType rat)
 Proof.
 pose numF_inj (Qr : fieldExtType rat) := {rmorphism Qr -> algC}.
 pose subAut := {Qr : _ & numF_inj Qr * {lrmorphism Qr -> Qr}}%type.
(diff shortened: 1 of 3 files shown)


gares:
I've no clue why, but it seems to break 8.8 and 8.7, while work on 8.6 and master. Ideas?

Zimmi48:
The error is:
```
File "./ssreflect/ssrfun.v", line 2, characters 29-41:
Error:
Cannot find a physical path bound to logical path matching suffix
<> and prefix mathcomp.
```
I cannot see how this could possibly be connected to this PR, but if it is, it must be a bug in coq_makefile or how it is used in math-comp. Have you tried testing only the changes in the `.v` files? That would be enough for now.

Zimmi48:
The push check passed. The pr check started later but it should give the same result.

Zimmi48:
@gares or @amahboubi Can you merge? This is needed to know what other projects depending on math-comp are requiring fixes.

Issue #156, opened 2017-11-14
provide all2


bgregoir:
The function all2 on list is not present in ssr.
We proving relation over programs, this is the kind of predicate which is very useful.
It can be easily defined either by a fixpoint or using all, zip, and size. 
Both solution have advantage it is not my point.
But it is a little bit boring to redefine it (and it associated lemma) 
each time I start a new development.

CohenCyril:
Please consider making a pull request, we can then work towards the best version to integrate.

CohenCyril:
@bgregoir I understand you have several copies of this, can you make a little syntesis either here or as a pull request? Cheers ;)

strub:
I took one of the several copies and pushed it. Now, I'm not sure which lemmas we may want to import.

PR #224, opened 2018-09-03
Implementation of all2

strub:
This is not a PR per se, but a way to start a discussion related to #156.


CohenCyril:
@strib thanks! 
@bgregoir what are the lemmas you usually have about `all2`?

bgregoir:
I think that the lemma all2P (of the pull request) is the only one needed.
Most of the time, proof is done by induction and the only important part is the fact that all2 simplify.
B

gares:
I imagine we also want `Arguments all2 {T U} p !s1 !s2.` so that it unfolds only when *both* lists are concrete.

mathcomp/ssreflect/seq.v
```diff
@@ -2783,3 +2783,21 @@ by apply: inj_f => //; apply/allpairsP; [exists (x, y1) | exists (x, y2)].
 Qed.
 
 End EqAllPairs.
+
+Section All2.
```
CohenCyril: And how about factoring `T U p` as a section context and have `all2E` stated on two line instead of three?

strub: If this makes you happy :)

CohenCyril: it does

amahboubi:
@bgregoir, can you give an example of behaviour you do not like?

bgregoir:
Reduction of all2 f (a::b) (c::d) —> f a c && all2 f b d

bgregoir:
Maybe my last email is not very clear.

> Reduction of all2 f (a::b) (c::d) —> f a c && all2 f b d

This is a reduction that I like that we do not have using zip and length.
We can add the rewriting rule, but it will not be reduction.
B

mathcomp/ssreflect/seq.v
```diff
@@ -2783,3 +2783,22 @@ by apply: inj_f => //; apply/allpairsP; [exists (x, y1) | exists (x, y2)].
...
+  match s1, s2 with
+  | [::], [::] => true
+  | x1 :: s1, x2 :: s2 => p x1 x2 && all2 s1 s2
+  | _, _ => false
+  end.
+
+Lemma all2E s1 s2: all2 s1 s2 =
+  (size s1 == size s2) && (all [pred xy | p xy.1 xy.2] (zip s1 s2)).
+Proof.
+by elim: s1 s2 => [|x s1 ih] [|y s2] //=; rewrite ih andbCA eqSS.
```
amahboubi: The final `eqSS` is useless, as `by` with do the conversion job.

amahboubi:
Sure, this is where the two definitions differ. You certainly know best, as you have been the one using it. However, I guess that we should consider including a minimal theory, if we include this definition in `seq`. At least fixpoint equations, but may be also connections with `allpairs`, `cat`, `subseq`,etc. I do not want to block the merge though.

amahboubi:
I do not know what to do here:
- github tells me this PR now has conflicts
- I do not think the change is worth 7 commits.

Can you at least squash?

strub:
I can resolve the conflicts. If I understand well, this is then up to you to squash the commit by choosing the correct merge option.

amahboubi:
It seems it's a global option of the repo:
https://help.github.com/articles/about-pull-request-merges/
which is apparently not set for math-comp.

amahboubi:
Yes, it is set, sorry.

CohenCyril:
@strub 
> this is then up to you to squash the commit by choosing the correct merge option.

The problem is: github has no option for squashing and creating a merge commit :disappointed:

amahboubi:
Isn't it what I just did?

strub:
On the other end, what the point of having a merge commit when you squash the development into one branch?

PR #226, opened 2018-09-11
adds relevant theorems when fcycle f (orbit f x) and the needed lemmas

ybertot:
The theorems in this collection are useful when using a function over a finite type to model pointers
in memory, especially for the case where the intended structure is a circular list.
the naming of these theorems needs a review, especially fconnect_step_cycle_orbit

[EDIT BY CYRIL]
Left TODO:
- [ ] think more carefully about the names of the theorems and find more meaningful names for sections
- [ ] add `Hint Resolve` outside of sections
- [ ] add documentation (if necessary) and ChangeLog
- [ ] cut this PR into bits? (e.g. TFAE fix gets its own PR?)

[/EDIT BY CYRIL]

### DIFF
--- a/mathcomp/algebra/ssralg.v
+++ b/mathcomp/algebra/ssralg.v
@@ -153,9 +153,10 @@ Require Import finfun bigop prime binomial.
 (*                  WARNING: while it is possible to omit R for most of the   *)
 (*                           XxxType functions, R MUST be explicitly given    *)
 (*                           when UnitRingType is used with a mixin produced  *)
-(*                           by ComUnitRingMixin, otherwise the resulting     *)
-(*                           structure will have the WRONG sort key and will  *)
-(*                           NOT BE USED during type inference.               *)
+(*                           by ComUnitRingMixin, in a Canonical definition,  *)
+(*                           otherwise the resulting structure will have the  *)
+(*                           WRONG sort key and will NOT BE USED during type  *)
+(*                           inference.                                       *)
 (* [unitRingType of R for S] == R-clone of the unitRingType structure S.      *)
 (*    [unitRingType of R] == clones a canonical unitRingType structure on R.  *)
 (*     x \is a GRing.unit <=> x is a unit (i.e., has an inverse).             *)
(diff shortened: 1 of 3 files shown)


pi8027:
Hello. I have used `fconnect` and related lemmas to reason about union-find. My situation is a bit different but this would be nice!

I found that `fconnect_step` can be a bit generalized like this:

```coq
Lemma fconnect_step (T : finType) (f : T -> T) (x y : T) :
  fconnect f x y = (x == y) || fconnect f (f x) y.
Proof.
apply/idP/idP => [/iter_findex <- | /predU1P [<- |] //].
- by case: (findex f x y) => [| i]; rewrite ?eqxx // iterSr fconnect_iter orbT.
- by apply/connect_trans/fconnect1.
Qed.
```

pi8027:
The proof of `order_step` can be simplified as follows by using above `fconnect_step`:

```coq
Lemma order_step x :
  order x = order (f x) /\ iter (order x) f x = x \/
  order x = (order (f x)).+1.
Proof.
have [| /negP noc] := boolP (fconnect f (f x) x); last first.
  right; rewrite /order; set M := fconnect f (f x); set N := pred1 x.
  have ->: #|fconnect f x| = #|[predU N & M]|
    by apply: eq_card=> z; rewrite !inE fconnect_step eq_sym.
  rewrite -[LHS]addn0 -add1n -(card1 x) -cardUI; congr addn.
  by rewrite eq_card0 // => z; rewrite !inE; apply/negP => /andP [/eqP ->].
rewrite fconnect_step_cycle_orbit.
left; split; last by apply: cycle_orbit_iter_order.
by apply/esym/cycle_orbit_order_eq; rewrite // -fconnect_orbit fconnect1.
Qed.
```

ybertot:
@pi8027, can you package your two suggestions in a pull request to branch fingraph-complements of ybertot/math-comp?   This way your contribution will be attributed to you.

pi8027:
@ybertot Thanks for merging. I think that `fcycle f (orbit f x)` ---more intuitively, `x` is in the cyclic part of the "lasso" or not--- is a fundamental condition of vertices in the finite graphs generated by `frel`, and there are more useful equivalent conditions and theorems around it. But I have no time to write out and reorganize them now...

CohenCyril:
Hi @ybertot  and @pi8027 !

While reviewing this PR I noticed a number of things. Here they are, and how I deat with them.

* lemmas `cycle_orbit_iter_order` and `cycle_orbit_finv_f_in` are trivial consequences of lemmas from section `orbit_in`, the only requirement was to prove  `{in orbit f x &, injective f}` first...
* in fact many things on `orbit x` can and should be generalized to an abstract `p` which is a `fcycle` and `uniq`, this shortens and generalize many proofs. Now `{in orbit f x &, injective f}` becomes `{in p &, injective f}` (such that `fcycle f p` and `uniq p`).
* lemmas from `orbit_in` may all be specialized to this domain `p` on which `f` is injective, this naturally trigger a simple code reorganization (section`Loop` went down). I chose to prefix them all with `cycle_`
* `cycle_orbit_iter_order_in` actually subsumes `cycle_orbit_iter_order` and gets to be the only left of those two. Its name is now shorter: `cycle_iter_order`.
* in the library, lemmas `orbit_rot_cycle` and `rot_to` are not informative enough -- about the natural number used for rotating -- for some proofs to go through easily. I added `rot_index` and `orbitE`
* The proof script of `fconnect_step_cycle_orbit` was a bit too hard to follow to my taste, moreover, the proof actually combined and squashed several useful equivalences. I separated them in the form of a "The Following Are Equivalent" (TFAE) lemma giving 4 different non-trivially equivalent characterizations of an orbit being a cycle, for the same price as before (10 loc)
* Doing the latter, I realized there was a bug in TFAE (notation not exported) and I fixed it.
* The statement of `order_step` was too specific and did not explicit the fact that the disjuncts are mutually exclusive. This is better stated as a "spec" lemma. What should be an index and what should be a constructor parameter is still debatable...

Left TODO:
- [ ] think more carefully about the names of the theorems and find more meaningful names for sections
- [ ] add `Hint Resolve` outside of sections
- [ ] add documentation (if necessary) and ChangeLog
- [ ] cut this PR into bits (e.g. TFAE fix gets its own PR?)

CohenCyril:
@ybertot please rebase on top of master.

CohenCyril:
@ybertot do you have any plans on progressing on this PR in the near future?

ybertot:
Expect an action by Nov. 28.

CohenCyril:
Accidentally closed

PR #230, opened 2018-09-12
Removing bool_irrelevance

### LINKED ISSUE #229: bool_irrelevance should be named differently
Let me cite a [comment](https://github.com/math-comp/finmap/issues/28#issuecomment-420630801) by @CohenCyril  which I agree with:

> the misnaming here would rather be `bool_irrelevance` since it is proof irrelevance for equalities of booleans... which by the way is subsumed by `eq_irrelevance` which deserves its name well...

Should `bool_irrelevance` be renamed or deleted altogether? Is there a policy on API-breaking changes?

CohenCyril:
Fixes #229


gares:
Discussion at the meeting: keep `Lemma bool_irrelevance b (H1 H2 : is_true b) : H1 = H2.`

anton-trunov:
@gares Thanks for sharing this. I am very interested in the Mathcomp contribution policy. Are there any plans of sharing it?

I guess the policy would shed some light on the motivation for this particular decision. Was it backward compatibility or some other considerations as well?

gares:
The statement up there is different, so it is not about backward compatibility. I guess the old statement was a but wrong, while this one states a property that is relevant to math comp, even if the lemma is not strictly necessary in this form. At least, this was my understanding of the discussion.

PR #242, opened 2018-11-12
Document the need for canonical mixins

anton-trunov:
As of now there is no use case for canonical mixins,
see this relevant SSReflect mailing list thread:
https://sympa.inria.fr/sympa/arc/ssreflect/2018-11/msg00000.html

CC @CohenCyril (thank you for the confirmation)

### DIFF
--- a/mathcomp/ssreflect/eqtype.v
+++ b/mathcomp/ssreflect/eqtype.v
@@ -107,7 +107,7 @@ Require Import ssrfun ssrbool.
 (*   [eqMixin of S by <:]                                                     *)
 (* construct to declare the equality mixin; this pattern is repeated for all  *)
 (* the combinatorial interfaces (Choice, Countable, Finite). As noted above,  *)
-(* such mixins should not be mad Canonical.                                   *)
+(* such mixins should not be made Canonical.                                  *)
 (*   We add the following to the standard suffixes documented in ssrbool.v:   *)
 (*  1, 2, 3, 4 -- explicit enumeration predicate for 1 (singleton), 2, 3, or  *)
 (*                4 values.                                                   *)
@@ -155,6 +155,17 @@ Export Equality.Exports.
 
 Definition eq_op T := Equality.op (Equality.class T).
 
+(* eqE is a generic lemma that can be used to fold back recursive comparisons *)
+(* after using partial evaluation to simplify comparisons on concrete         *)
+(* instances. The eqE lemma can be used e.g. like so: rewrite !eqE /= -!eqE.  *)
+(* For instance, with the above rewrite, n.+1 == n.+1 gets simplified to      *)
+(* n == n. For this to work, we need to declare equality _mixins_             *)
+(* as canonical. Canonical declarations remove the need for specific          *)
+(* inverses to eqE (like eqbE, eqnE, eqseqE, etc.) for new recursive          *)
+(* comparisons, but can only be used for manifest mixing with a bespoke       *)
+(* comparison function, and so is incompatible with PcanEqMixin and the like  *)
+(* - this is why the tree_eqMixin for GenTree.tree in library choice is not   *)
+(* declared Canonical.                                                        *)
 Lemma eqE T x : eq_op x = Equality.op (Equality.class T) x.
 Proof. by []. Qed.
 
@@ -745,7 +756,7 @@ Section ProdEqType.
 
 Variable T1 T2 : eqType.
 
-Definition pair_eq := [rel u v : T1 * T2 | (u.1 == v.1) && (u.2 == v.2)].
+Definition pair_eq : rel (T1 * T2) := fun u v => (u.1 == v.1) && (u.2 == v.2).
 
 Lemma pair_eqP : Equality.axiom pair_eq.
 Proof.
@@ -753,7 +764,7 @@ move=> [x1 x2] [y1 y2] /=; apply: (iffP andP) => [[]|[<- <-]] //=.
 by do 2!move/eqP->.
 Qed.
 
-Definition prod_eqMixin := EqMixin pair_eqP.
+Canonical prod_eqMixin := EqMixin pair_eqP.
 Canonical prod_eqType := Eval hnf in EqType (T1 * T2) prod_eqMixin.
 
 Lemma pair_eqE : pair_eq = eq_op :> rel _. Proof. by []. Qed.
@@ -770,9 +781,8 @@ Proof. by case/andP. Qed.
 
 End ProdEqType.
 
-Arguments pair_eqP [T1 T2].
-
-Prenex Implicits pair_eqP.
+Arguments pair_eq {T1 T2} _ _ /.
+Arguments pair_eqP {T1 T2}.
 
 Definition predX T1 T2 (p1 : pred T1) (p2 : pred T2) :=
   [pred z | p1 z.1 & p2 z.2].
@@ -796,6 +806,8 @@ Canonical option_eqMixin := EqMixin opt_eqP.
 Canonical option_eqType := Eval hnf in EqType (option T) option_eqMixin.
 
 End OptionEqType.
+
+Arguments opt_eq {T} !_ !_.
 
 Section TaggedAs.
 
@@ -841,8 +853,8 @@ Proof. by rewrite -tag_eqE /tag_eq eqxx tagged_asE. Qed.
 
 End TagEqType.
 
-Arguments tag_eqP [I T_ x y].
-Prenex Implicits tag_eqP.
+Arguments tag_eq {I T_} !_ !_.
+Arguments tag_eqP {I T_ x y}.
 
 Section SumEqType.
 
@@ -865,5 +877,5 @@ Lemma sum_eqE : sum_eq = eq_op. Proof. by []. Qed.
 
 End SumEqType.
 
-Arguments sum_eqP [T1 T2 x y].
-Prenex Implicits sum_eqP.
+Arguments sum_eq {T1 T2} !_ !_.
+Arguments sum_eqP {T1 T2 x y}.
--- a/mathcomp/ssreflect/tuple.v
+++ b/mathcomp/ssreflect/tuple.v
@@ -360,7 +360,12 @@ Section UseFinTuple.
 
 Variables (n : nat) (T : finType).
 
-Canonical tuple_finMixin := Eval hnf in FinMixin (@FinTuple.enumP n T).
+(* tuple_finMixin could, in principle, be made Canonical to allow for folding *)
+(* Finite.enum of a finite tuple type (see comments around eqE in eqtype.v),  *)
+(* but in practice it will not work because the mixin_enum projector          *)
+(* has been burried under an opaque alias, to avoid some performance issues   *)
+(* during type inference.                                                     *)
+Definition tuple_finMixin := Eval hnf in FinMixin (@FinTuple.enumP n T).
 Canonical tuple_finType := Eval hnf in FinType (n.-tuple T) tuple_finMixin.
 Canonical tuple_subFinType := Eval hnf in [subFinType of n.-tuple T].
 
--- a/mathcomp/algebra/vector.v
+++ b/mathcomp/algebra/vector.v
@@ -1302,7 +1302,7 @@ Section LfunZmodType.
 Variables (R : ringType) (aT rT : vectType R).
 Implicit Types f g h : 'Hom(aT, rT).
 
-Canonical lfun_eqMixin := Eval hnf in [eqMixin of 'Hom(aT, rT) by <:].
+Definition lfun_eqMixin := Eval hnf in [eqMixin of 'Hom(aT, rT) by <:].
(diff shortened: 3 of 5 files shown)


amahboubi:
Hi @anton-trunov, thanks for this PR. Yet before merging, I would like to wait for @ggonthier's feedback, since I believe he was the author of these `Canonical` declarations.

anton-trunov:
@amahboubi Sure thing! I would also love to learn from his input! Thanks.

ggonthier:
Making a mixin `Canonical` does serve a purpose: it lets type inference extrapolate the mixin and hence the corresponding packed class (when the mixing is also a class) from one of its value projections, just as one does for value projections of packed classes, such as `topred` in `predType`, or `val` and `Sub` in `subType`.
  In the MathComp library this is only really implemented for `eqType`. The idea is that this makes it possible to rewrite right-to-left with lemma `eqE`, to fold back recursive comparisons after using partial evaluation to simplify a comparison on concrete instances.
   The generic way to do this would be`rewrite eqE /= -!eqE`.
The reason one may need to fold back is that we need to give the full computational rule for recursive datatypes such as `nat` or `seq` _before_ we can declare a mixin, and thus it is not possible to use eq_op to do the recursive comparison.
  This feature isn't used in the MathComp library, which favours using more precise specialisations of `-eqE`, such as `eqbE` or `eqseqE`, or even helpful corollaries like `eqSS` or `eqseq_cons`.
   Despite this I still believe having a generic reversible `eqE` is useful, so I'd rather we let the `Canonical` declarations stand - perhaps while documenting it above `eqE` in the text, with a pointer in the `eqtype` file header.
   We should point out that such `Canonical` declarations remove the need for specific inverses to `eqE` for new recursive , but can only be used for manifest mixing with a bespoke comparison function, and so is incompatible with `PcanEqMixin` and the like - this is why the `tree_eqMixin` for `GenTree.tree` in library `choice` is not declared `Canonical`.

ggonthier:
There is nevertheless one instance, in library `tuple`, of a `Canonical` `finMixin`. This could, in principle, make it possible to similarly fold the `Finite.enum` of a finite `tuple` type... but in practice will not work because the `mixin_enum` projector has been berried under an opaque alias, to avoid some nasty performance issues during type inference. Thus in that case you _have_ spotted an outdated `Canonical` declaration that has slipped by, and should indeed be changed to `Definition`.
   Well done!

anton-trunov:
@ggonthier Thank you very much for the explanation! Things are much clearer now.

**(1)** I will restore the `Canonical` declarations and add some documenting comments.

**(2)** If I understand correctly, declaring the mixins of *non-recursive* types as `Canonical` is still desirable if one wants to be less precise with rewriting context, e.g.`do ! [rewrite !eqE /= -!eqE].`
But this will break down for non-symmetrical equality operators such as `opt_eq` for `option` type, e.g.
```coq
Lemma foo (b : bool) (o : option nat) :
  (Some 42 == o) && (o == Some 42) && (b == false) && (false == b).
Proof.
do !rewrite !eqE /= -!eqE.

(* the goal at this point is
  oapp (fun n : nat => match n with
                       | 42 => true
                       | _ => false
                       end) false o && (o == Some 42) && (b == false) && (false == b)
*)
Abort.
```
Should I try and fix this behavior?

**(3)** It seems that `lfun_eqMixin` should *not* be declared as `Canonical`, as
```coq
Canonical lfun_eqMixin := Eval hnf in [eqMixin of 'Hom(aT, rT) by <:].
```
produces
```
Warning: Projection value has no head constant: fun x y : 'Hom(aT, rT) => f2mx x == f2mx y in canonical instance lfun_eqMixin of Equality.op, ignoring
it. [projection-no-head-constant,typechecker]
```

ggonthier:
Well observed, again!
  Quick answers for (2) and (3): The fix for (2) would be to use an `Arguments` declaration to require both arguments be constructors for simplification to take place. That feature did not yet exist when the math comp library was written. Do note however that you suggested command is likely to diverge, as, if I recall correctly, `do ![tactic]` does not check for progress (perhaps it should?).
   Finally, you are correct regarding `lfun` - well spotted, again.

anton-trunov:
> The fix for (2) would be to use an `Arguments` declaration to require both arguments be constructors for simplification to take place. 

Thank you, `Arguments opt_eq [T] !_ !_.` seems to work fine.

> if I recall correctly, `do ![tactic]` does not check for progress (perhaps it should?)

My experiments seem to show `do` does check for progress, but I could not find this in the docs. The manual says "`!` the step tactic is repeated as many times as possible, and done at least once." Perhaps, @gares could clarify this a bit. If `do` really checks for progress, we could fix this behavior in Coq's test suite and add it to the manual.

anton-trunov:
I just pushed an updated version with more documenting commentaries. I'd be happy to add any improvements or clarifications!

mathcomp/ssreflect/eqtype.v
```diff
@@ -98,6 +98,9 @@ Require Import ssrfun ssrbool.
 (*   We add the following to the standard suffixes documented in ssrbool.v:   *)
 (*  1, 2, 3, 4 -- explicit enumeration predicate for 1 (singleton), 2, 3, or  *)
 (*                4 values.                                                   *)
+(* eqE is a generic lemma that can be used to fold back recursive comparisons *)
+(* after using partial evaluation to simplify a comparison on                 *)
+(* concrete instances, see more in the comments above eqE                     *)
 (******************************************************************************)
```
ggonthier: This is not quite the right information to put in the header: the key point to make is that `EqMixin` instances should be declared `Canonical`, and this info should appear in or just after the entry for EqMixin ... which is missing in the documentation. Indeed the documentation for factories for eqType is seriously incomplete - it should have entries for all the major definitions and notation, as in `ssralg` or `fintype` (`choice` is also lacking, though less so). This needs to be addressed beforehand, ideally in a separate commit.
I'll do this first so you can proceed.

mathcomp/ssreflect/tuple.v
```diff
@@ -360,7 +360,13 @@ Section UseFinTuple.
...
 Variables (n : nat) (T : finType).
 
-Canonical tuple_finMixin := Eval hnf in FinMixin (@FinTuple.enumP n T).
+(* tuple_finMixin could, in principle, be made Canonical to allow for folding
+Finite.enum of a finite tuple type (see comments around eqE in eqtype.v),
+but in practice it will not work because the mixin_enum projector
+has been burried under an opaque alias, to avoid some performance issues
+during type inference
+*)
+Definition tuple_finMixin := Eval hnf in FinMixin (@FinTuple.enumP n T).
```
ggonthier: See above remark concerning the multiline comment style (and punctuation).

mathcomp/ssreflect/eqtype.v
```diff
@@ -784,6 +799,8 @@ Canonical option_eqType := Eval hnf in EqType (option T) option_eqMixin.
 
 End OptionEqType.
 
+Arguments opt_eq [T] !_ !_.
```
ggonthier: All other explicit comparison functions will have a similar issue, so you should run through them and add similar annotations throughout.
As a rule, the type parameter(s) of a polymorphic function should be maximally inserted, as this makes it possible to use said function as one would in ML or Haskell, so you should have `{T}` instead of `[T]`.

mathcomp/ssreflect/eqtype.v
```diff
@@ -142,6 +145,18 @@ Export Equality.Exports.
...
+rewrite, n.+1 == n.+1 gets simplified to n == n. For this to work, we need to
+declare equality _mixins_ as canonical.
+
+Canonical declarations remove the need for specific inverses to eqE (like eqbE,
+eqnE, eqseqE, etc.) for new recursive comparisons, but can only be used for
+manifest mixing with a bespoke comparison function, and so is incompatible with
+PcanEqMixin and the like - this is why the tree_eqMixin for GenTree.tree in
+library choice is not declared Canonical
+*)
 Lemma eqE T x : eq_op x = Equality.op (Equality.class T) x.
```
ggonthier: The MathComp library uses exclusively block comments, with 80-character lines enclosed in the `(*` / `*)` delimiters. Multiline comments are strictly limited to out-commented code. Please conform to this stye.
Also, that last sentence is missing its final stop.

mathcomp/ssreflect/eqtype.v
ggonthier: Actually, while editing the header I realised there _was_ a comment to that effect in the entry for eqType. I'll just insert the forward pointer to eqE so you won't have to do anything else.

mathcomp/ssreflect/eqtype.v
anton-trunov: The case of `pair_eq` is a bit tricky if we'd like to be able to fold back for the products.

If we did, then `pair_eq`'s definition would be tweaked a bit to allow for an annotation like this:
```coq
Arguments pair_eq {T1 T2} !_ !_ /.
```
Then `prod_eqMixin` could be made canonical.

I tried this option and only two lines (in mathcomp) needed simple fixes. I'll push the tentative changes, but I'm fine with reverting those as this breaks some things for the users relying on that previous simplifying behavior.

anton-trunov:
@ggonthier I tried to address some of your suggestions in two separate commits (I'll squash all three of the new commits afterwards). I hope I didn't miss any explicit comparison functions.

mathcomp/ssreflect/eqtype.v
ggonthier: I actually did add a forward reference to `eqE` along with more precise advice on declaring `Canonical` equality mixins in the updated documentation in the master branch. This makes this extra comment somewhat redundant, and too detailed for the header section, where, as a rule, we try to document types, definitions and notations precisely, but only describe the lemmas and theorems in general terms, because we don't want to discourage users from actually reading the documentation. There are some exceptions for some particularly important theorems, but `eqE` isn't one of those.

mathcomp/ssreflect/eqtype.v
```diff
@@ -732,15 +744,16 @@ Section ProdEqType.
...
 
 Lemma pair_eqP : Equality.axiom pair_eq.
 Proof.
 move=> [x1 x2] [y1 y2] /=; apply: (iffP andP) => [[]|[<- <-]] //=.
 by do 2!move/eqP->.
 Qed.
 
-Definition prod_eqMixin := EqMixin pair_eqP.
+Canonical prod_eqMixin := EqMixin pair_eqP.
 Canonical prod_eqType := Eval hnf in EqType (T1 * T2) prod_eqMixin.
```
ggonthier: The intent in defining `pair_eq` as a `simpl_rel`, which always simplifies when applied, is clearly to make such unfolding automatic and irreversible. This is reasonable because this yields in a readable expression. Even if you've only found one instance of this being used in the math comp library itself, I don't think it's a good idea to change this behaviour arbitrarily; I suspect there may be more instances in the Odd Order theorem proof, for one. So I would advise against that particular change.
  However, in this case the `Canonical` declaration _is_ an error, because the head constant of the comparison is the generic `rel_of_simpl_rel` coercion, rather than `pair_eq` - so you are correct in using `Definition` here.
   Alternatively, one could declare `pair_eq` as a `rel (T1 * T2)` and use the `/` annotation of `Arguments` to implement the unfolding behaviour; then the head constant would be `pair_eq`, and the `Canonical` declaration would be proper (though only marginally useful).

anton-trunov:
@ggonthier I squashed the commits, removed the redundant `eqE` description from the header and incorporated your suggestion for `pair_eq` as a `rel (T1 * T2)`.

mathcomp/ssreflect/eqtype.v
```diff
@@ -770,8 +781,8 @@ Proof. by case/andP. Qed.
 
 End ProdEqType.
 
+Arguments pair_eq {T1 T2} _ _ /.
 Arguments pair_eqP [T1 T2].
-
 Prenex Implicits pair_eqP.
```
ggonthier: While we're at it, perhaps this you should merge the `Arguments` and `Prenex Implicits` lines...

anton-trunov: Sure! Would you prefer if I did it for the whole file for consistency? Or just locally for the parts I touched? (so the rest of the declarations could be  addressed in a separate PR)

ggonthier: It's probably best to do only this change for this PR (as you're touching those lines, might as well do a clean job). If you are willing, it would indeed be good to clean up the `Arguments` declarations for reflection lemmas, which are the result of a general port of old-style implicit declaration. For one thing, the intent was always to have maximal implicit arguments for reflection lemmas, but the old-style declarations only used to provide non-maximal implicits. Hence the `Prenex Implicits` declarations, which are only occasional because implicits just happen to behave like maximal implicits when the last argument is non-dependent. Putting in the proper declarations would certainly be cleaner.

anton-trunov: Updated. I'll try to address the rest of the `Prenex Implicits` declarations for reflection lemmas in a separate PR

PR #246, opened 2018-11-15
Add a section on documenting code

anton-trunov:
Based on the discussion here:
https://github.com/math-comp/math-comp/pull/242


anton-trunov:
@CohenCyril Thanks for merging!

PR #249, opened 2018-11-21
Merge Arguments and Prenex Implicits

anton-trunov:
See the discussion here:
https://github.com/math-comp/math-comp/pull/242#discussion_r233778114

### DIFF
--- a/mathcomp/algebra/matrix.v
+++ b/mathcomp/algebra/matrix.v
@@ -911,14 +911,13 @@ End VecMatrix.
 
 End MatrixStructural.
 
-Arguments const_mx [R m n].
-Arguments row_mxA [R m n1 n2 n3 A1 A2 A3].
-Arguments col_mxA [R m1 m2 m3 n A1 A2 A3].
+Arguments const_mx {R m n}.
+Arguments row_mxA {R m n1 n2 n3 A1 A2 A3}.
+Arguments col_mxA {R m1 m2 m3 n A1 A2 A3}.
 Arguments block_mxA
-  [R m1 m2 m3 n1 n2 n3 A11 A12 A13 A21 A22 A23 A31 A32 A33].
-Prenex Implicits const_mx castmx trmx lsubmx rsubmx usubmx dsubmx row_mx col_mx.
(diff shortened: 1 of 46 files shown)


anton-trunov:
@CohenCyril Thank you, it's mu pleasure.

anton-trunov:
@ggonthier Sure, I'll try to address this in a separate PR.

Just to be on the safe side, could you give an example of what you'd like to have?

mathcomp/algebra/matrix.v
```diff
@@ -2667,8 +2663,7 @@ Coercion GLval ph (u : GLtype ph) : 'M[R]_n.-1.+1 :=
 End FinUnitMatrix.
 
 Bind Scope group_scope with GLtype.
-Arguments GLval _%N _ _ _%g.
-Prenex Implicits GLval.
+Arguments GLval {_%N _ _} _%g.
```
ggonthier: Well for example this should be `Arguments GLval {n%N R ph} u%g.` to document the name of the arguments (one might put `-` instead of `ph` here, however, since it's not expected one would pass a phantom argument explicitly).

anton-trunov: Thank you! This looks nicer, indeed.

anton-trunov:
@CohenCyril Thanks!

mathcomp/algebra/matrix.v
anton-trunov: One more question, if you don't mind.
What do we do with the following declaration?
```
Prenex Implicits pred1 pred2 pred3 pred4 predU1 predC1 predD1.
```

Do we leave it as it is or a convert to `Arguments`?

ggonthier: My preference would be to leave it. The `Prenex Implicits` command is a natural way of stating that a function constant should behave as in functional programming languages - something that ought to be supported by a global option or a pragma (once Coq gets these); `Context` declarations wouldn't quite work here as they would also affect lemmas. It seems rather wasteful to add one extra line for each function definition for this.

anton-trunov: Great! Thank you. I'll try to address your comments this week.

PR #250, opened 2018-11-22
do not rely on `elim E: x => i` not clearing `x` if `x` not used in `i`

gares:
I'm not very fond of the new proof script.

I believe that this specific aspect of the equation generation feature was not documented.

With the introduction of new intro patterns it is not so clear what it means to be "not used in `i`" since
`i` may contain `[^ foo ]`. Hence the attempt to simplify the code by removing this special case, that is used only once in the entire library.

### DIFF
--- a/mathcomp/ssreflect/seq.v
+++ b/mathcomp/ssreflect/seq.v
@@ -2043,7 +2043,7 @@ Qed.
 Lemma perm_to_subseq s1 s2 :
   subseq s1 s2 -> {s3 | perm_eq s2 (s1 ++ s3)}.
 Proof.
-elim Ds2: s2 s1 => [|y s2' IHs] [|x s1] //=; try by exists s2; rewrite Ds2.
+pose s := s2; elim: s2 s1 => [|y s2 IHs] [|x s1] //= in s *; try by exists s.
 case: eqP => [-> | _] /IHs[s3 perm_s2] {IHs}.
   by exists s3; rewrite perm_cons.
 by exists (rcons s3 y); rewrite -cat_cons -perm_rcons -!cats1 catA perm_cat2r.


mathcomp/ssreflect/seq.v
```diff
@@ -2043,7 +2043,7 @@ Qed.
 Lemma perm_to_subseq s1 s2 :
   subseq s1 s2 -> {s3 | perm_eq s2 (s1 ++ s3)}.
 Proof.
-elim Ds2: s2 s1 => [|y s2' IHs] [|x s1] //=; try by exists s2; rewrite Ds2.
```
gares: note how the more natural name `s2` was avoided (using `s2'`) in order to have `elim` reuse `s2` for the LHS of `Ds2`.
The LHS is inaccessible (eg fresh) after coq/coq#6705

mathcomp/ssreflect/seq.v
```diff
@@ -2043,7 +2043,7 @@ Qed.
...
+elim Ds2: s2 s1 => [|y ys IHs] [|x xs] //=; only 1,2: by eexists; rewrite -Ds2.
```
gares: FTR: I would have love to write `first 2 by ...`

mathcomp/ssreflect/seq.v
CohenCyril: After discussing with @gares, the fix to still access `s2` is to write `(s2)` instead.

gares:
Unfortunately the fix you propose does not work, since you would keep `s2` but the equation would be `_s_ = ...` not `s2 = ...`. The current script is one line longer, but is way more pedestrian.

gares:
Yes, I'll look into that. As I explained to you in person the code is now split into two files, so it may not be trivial to implement correctly. If I manage the right fix could be to just put parentheses in `elim Ds2: (s2)`.

If I don't manage, is this PR an acceptable fix?

mathcomp/ssreflect/seq.v
```diff
@@ -2043,8 +2043,9 @@ Qed.
 Lemma perm_to_subseq s1 s2 :
   subseq s1 s2 -> {s3 | perm_eq s2 (s1 ++ s3)}.
 Proof.
-elim Ds2: s2 s1 => [|y s2' IHs] [|x s1] //=; try by exists s2; rewrite Ds2.
-case: eqP => [-> | _] /IHs[s3 perm_s2] {IHs}.
+elim: s2 s1 => [|y s2 IHs] [|x s1] //=; first by exists [::].
+  by exists [::y & s2].
+case: eqP => [-> | _] {}/IHs[s3 perm_s2].
```
gares: the `{}/IHs` syntax is new in 8.9, so I should revert this change or it fails on older versions

gares:
It looks all green now.

gares:
Oh, given that the fix is backward compatible I'd like it to be merged before coq/coq#6705 so that I can remove the overlay file from that PR.

ggonthier:
Could one of you clarify what the fix is about? The code that is being changed relies only on the standard behaviour of equation generation for `elim:`, which is to generate an abbreviation for the principal term of the elimination, reusing the name of that term if it has just been cleared by the `:` pseudo-tactical (and defaulting to a generated hidden name if that was not the case).
  I don't recall an exception for non-dependent `elim`, though there might have been one to accommodate `elim`s that behave more like `case`, specifically those that use a view. At any rate, this would not apply here, and the rationale is clearly wrong, because `s2` _is cleared_ here, before being reused.
  If the idea is to align the specific equation-generation behaviour of `elim` to that of `case`, then I would like to object. While it's not used very much in the final library scripts, it is a very useful _debugging_ tool and has been used extensively in producing the scripts in the first place. It's also more useful when doing small-scale reflection style program correctness proofs; there aren't many in MathComp per se, but I recall there are more in the 4-color proof.

mathcomp/ssreflect/seq.v
```diff
@@ -2043,7 +2043,8 @@ Qed.
 Lemma perm_to_subseq s1 s2 :
   subseq s1 s2 -> {s3 | perm_eq s2 (s1 ++ s3)}.
 Proof.
-elim Ds2: s2 s1 => [|y s2' IHs] [|x s1] //=; try by exists s2; rewrite Ds2.
```
ggonthier: If you're going to remove the `elim`  equation idiom (which I'm opposed to, see my general comment), then you should at least propose to replace it with something that is similar in functionality, and not imply that users need to duplicate cases in their scripts. Here the solution could be:
 ```
pose s := s2; elim: s2 s1 => [|y s2 IHs] [|x s1] //= in @s *; try by exists s.
```

ggonthier:
Partly answering myself now (a sure sign of senility) I do recall one feature of `elim` equation generation that may have sparked the confusion: if the cleared eliminated name has somehow been reused before the equation is introduced, then the introduction is done with a generated name. The rationale is that because this is primarily intended as a debugging feature, it should avoid introducing failures of its own.
  I still don't see how the new intro patterns should prevent this from working - it's exactly as if the user had introduced the name again at the point where the equation is introduced.

gares:
What I'd like to get rid of is the "if" in the code doing "if the name is not used later on in the intro pattern then reuse the name (if it is a name and not a general term) else pick a fresh name" since this is not very meaningful now that the intro pattern can mention a "prefix/suffix" of the actual name.

I see 3 options:
- never reuse the name: always pick a fresh id for the LHS. This is obviously the simplest, and the one currently implemented in the PR
- always reuse the name (when it is a name, and not a general term): could break scripts and force naming such as `s2'` as in the current proof script.
- reuse the name (when it is a name, and not a general term) if the user did put `()` around it (requesting it not to be cleared). This is what Cyril suggested.

I'm afraid all options could break scripts. I've no strong opinions.

I think the choice is about what status to give to elim+equation. If it is a debugging only thing, then I think option 1 is best. If it is a legit proof step, options 2 and 3 are fine but we should also fix the manual (that presents it as a debugging feature).

mathcomp/ssreflect/seq.v
gares: thanks for nicer script!

ggonthier:
As I see it there is no "used later in the intro pattern" behaviour that is required. The equation is introduced after the first intro pattern - so the most important question is whether the name has been reused at that very point, not whether it is being reused at some later point, in essence acting as if, after the first intro pattern, one pushes `forall _x_, _x_ = t -> ` on the goal (where `t` is the eliminated term), then insert `x Dx` in the intro pattern, where `x` is the (cleared) original name of the elimination target, and `Dx` the name of the equation. Should this fail because `x` was reused in the meantime, then `_x_` is used instead to avoid a secondary failure.
   It seems to me this should be fairly easy to implement. Yes, it does allow for different behaviour in different branches of the elimination (if the fallback behaviour is only triggered in some branches), as well as some secondary failures (if the elimination name is reused later in the pattern), but I feel these are acceptable limitations.
   I can even imagine ways of avoiding these without static prediction of the names used in an intro pattern, such as creating the equation with a generated name, then renaming it at the end of the intro pattern if no conflict has appeared in any of the branches, similar to how delayed clears are handled. I don't feel it's necessary to go there, however, as the behaviour I outlined above does cover all intended use cases well enough.

gares:
Oh, I see. The implementation you suggest is clearly simpler than the previous one and requires no lookahead. Thanks.

gares:
coq/coq#6705 updated, there is no need for this anymore

PR #257, opened 2018-12-05
Adding lemma `eqmxMunitP`

CohenCyril:
Missing useful theorem for mxalgebra

### DIFF
--- a/mathcomp/algebra/mxalgebra.v
+++ b/mathcomp/algebra/mxalgebra.v
@@ -1609,6 +1609,20 @@ rewrite (eqmxMr _ (genmxE U)) injfU genmxE addrC -mulmxDl subrK.
 by rewrite mul1mx mulVmx ?row_ebase_unit.
 Qed.
 
+(* Two matrices with the same shape represent the same subspace *)
+(* iff they differ only by a change of basis.                   *)
+
+Lemma eqmxMunitP m n (U V : 'M_(m, n)) :
+  reflect (exists2 P, P \in unitmx & U = P *m V) (U == V)%MS.
+Proof.
+apply: (iffP eqmxP) => [eqUV | [P Punit ->]]; last first.
+  by apply/eqmxMfull; rewrite row_full_unit.
+have [D defU]: exists D, U = D *m V by apply/submxP; rewrite eqUV.
+have{eqUV} [Pt Pt_unit defUt]: {Pt | Pt \in unitmx & V^T *m D^T = V^T *m Pt}.
+  by apply/complete_unitmx; rewrite -trmx_mul -defU !mxrank_tr eqUV.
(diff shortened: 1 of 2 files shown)


mathcomp/algebra/mxalgebra.v
```diff
@@ -1609,6 +1609,19 @@ rewrite (eqmxMr _ (genmxE U)) injfU genmxE addrC -mulmxDl subrK.
 by rewrite mul1mx mulVmx ?row_ebase_unit.
 Qed.
 
+(* Two matrices have the same image iff                 *)
+(* they are equal up to a change of basis of the domain *)
+
+Lemma eqmsP m n (f g : 'M_(m, n)) :
```
thery: what is the rationale behind the name 
eq ok
P ok,
 ms matrice space?

CohenCyril: Yes... on second thoughts I'm not sure it is good. However `eqmxP` is taken as it is the reflection lemma between `eqmx A B` (i.e. `A :=: B`) and `(A == B)%MS`. Other names I have in mind are `submx2P`, `samemxP`,  Do you have other suggestions?

ggonthier: I'd suggest `eqmxMunitP`, by analogy with `eqmxMfull`, which is the converse implication of the equivalence stated by this lemma (or rather, its generalisation to matrices with different shapes).
I'd also suggest changing the caption to 

> two matrices with the same shape represent the same subspace iff they differ only by a change of basis.

and changing the name of the variables of the lemma to `U` and `V`, which are the names commonly used in `mxalgebra` to denote matrices representing subspaces. While it is the case that the row space of a matrix is the codomain of the linear function it represents (by right multiplication), in general `mxalgebra` tries to avoid introducing this functional interpretation when it is not germane.

Finally, the proof should use `eqmxMfull` for the converse implication, and I think the forward implication would benefit from a more declarative style, even at the cost of a extra script line:
```
Lemma eqmxMunitP m n (U V : 'M_(m, n)) :
  reflect (exists2 P, P \in unitmx & U = P *m V) (U == V)%MS.
Proof.
apply: (iffP eqmxP) => [eqUV | [P Punit ->]]; last first.
  by apply/eqmxMfull; rewrite row_full_unit.
have [D defU]: exists D, U = D *m V by apply/submxP; rewrite eqUV.
have{eqUV} [Pt Pt_unit defUt]: {Pt | Pt \in unitmx & V^T *m D^T = V^T *m Pt}.
  by apply/complete_unitmx; rewrite -trmx_mul -defU !mxrank_tr eqUV.
by exists Pt^T; last apply/trmx_inj; rewrite ?unitmx_tr // defU !trmx_mul trmxK.
Qed.
```

ggonthier: I forgot to note you should also update the `Changes` blurb...

CohenCyril:
@ggonthier I think this is ready for merge

Issue #258, opened 2018-12-06
Update ChangeLog for PR #253


CohenCyril:
PR #253 introduces changes that must be documented in the ChangeLog

anton-trunov:
Will do. What do you think of an entry like this?
```
	* Documented argument names for functions and lemmas. Renamed some
	  of the arguments for consistency reasons. Changed implicit status from
	  non-maximal to maximal for a number of arguments.
```

CohenCyril:
Yes

Issue #255, opened 2018-12-03
Document scopes and implicitness for arguments


anton-trunov:
As suggested [here](https://github.com/math-comp/math-comp/pull/253#discussion_r238213677) by @ggonthier.

PR #262, opened 2018-12-11
Cancel lemmas implicit

ggonthier:
Make trailing implicit parameters of cancel f g and {in D, cancel f g} lemmas maximal, so that these lemmas are easier to pass to canLR, canLR_in, and similar.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -24,7 +24,20 @@
 	  from eqtype.v. They were already in ssrfun.v.
 
 	* Specialized `bool_irrelevance` so that the statement reflects
-	  the name
+	  the name.
+
+	* Changed the shape of the type of FieldMixin to allow one-line
+	  in-proof definition of bespoke fieldType structure.
+
+	* Refactored and extended Arguments directives to provide more
+	  comprehensive signature information.
+
+	* Added maximal implicits to reflection, injectivity and cancellation
+	  lemmas so that they are easier to pass to combinator lemmas such as
+	  sameP, inj_eq or canLR.
+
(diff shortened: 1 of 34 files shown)


anton-trunov:
Would it make sense to add an entry to ChangeLog in this PR to fix #258? Or should I open a new one?
I guess the proposed entry 
```
	* Documented argument names for functions and lemmas. Renamed some
	  of the arguments for consistency reasons. Changed implicit status from
	  non-maximal to maximal for a number of arguments.
```
should be changed anyways.

ggonthier:
@anton-trunov , yes, one entry should suffice to cover both PRs, and it should be a little more precise than the on proposed. I'm working on extending slightly the PR to cover uses of the lemmas  as suggested by @CohenCyril, and also to include partial function cancellation lemmas.

ggonthier:
Added usage, partial function cancel (`pcancel` / `ocancel`), and some residual `injective` / `cancel` lemmas.
Rebased, documented (in Changes, also covering #255), and squashed.

gares:
odd order is now part of Coq's CI and runs on each PR, so breaking it is not nice.

What I do for elpi is to have Coq CI track a branch other than master that I synchronize with master when I'm sure. See https://github.com/coq/coq/blob/master/dev/ci/README-users.md#recommended-branching-policy for example.

We could also improve CI here, so that we test odd-order too before merging.

ggonthier:
I've corrected the issue in odd-order,  and also submitted #264 to correct related issues in `mxrepresentation.v`.

PR #265, opened 2018-12-18
swap mingroup / maxgroup arguments

ggonthier:
Moved set argument before predicate argument for mingroup and maxgroup
because Coq only displays notation with identifier parameters that are
both bound and free if there is at least one free occurrence to the
left of the binder.


ggonthier:
I'm debating this with @herbelin in coq/coq#9180 right now. This is a longstanding issue that may eventually get fixed, but this PR should permanently  settle the issue for `mingnroup` and `maxgroup`. I only realised recently that there was some partial support for externing notation with mixed free/bound variables.

PR #252, opened 2018-11-26
Fix some new warnings emitted by Coq 8.10

anton-trunov:
```
Warning: Adding and removing hints in the core database implicitly is
deprecated. Please specify a hint database.
[implicit-core-hint-db,deprecated]
```

Partial fix for #90

Issue #233, opened 2018-10-11
ltngtP


thery:
commit d762ebb5a8c5191d49a75aa89ec34966de00eb9b

extends the use of ltngt

https://github.com/math-comp/math-comp/blob/5f8d45b54aa98732ec3de43d91814459d5a2f2e4/mathcomp/ssreflect/ssrnat.v#L450

from lt to leq so this lemma becomes somewhat wrongly named

thery:
Here is what is puzzling me:

From mathcomp Require Import all_ssreflect.
Goal forall a b, a <= b <= a -> a = b.
by move=> a b; case: ltngtP.

gares:
Looks like that the indexes of the spec are in the wrong order wrt what the name suggests.
Eg the one for == is matched before the one for < that is what one could expect.
@CohenCyril

CohenCyril:
@thery, @gares: the commit you mention did not extend the lemma from `lt` to `leq`, there were and there are still three cases which are, in order of generated subgoals `m < n`, `m > n` and `m = n`, which reflects the choice of name. What changed are the matched indexes: I increased the number of indexes to match the ones of `ltrgtP` from  `ssrnum`, which causes more substitutions when performing  a `case`. I also preserved the order of the matching by keeping the equality in the last position (i.e. matched first) and putting the large inequalities in the first position (i.e. matched last).

thery:
Je suis sans ordinateur, donc je ne peux pas verifier mais le truc qui m'embetait je crois c' est que ca s' applique si il n' y avait pas de lt 
dans le but.

thery:
oops it is not my day today, sorry for the previous comment in french. 

My point is that I find it weird a theorem called ltngtP applies when there is no > or < in the goal 

-- 
Laurent

CohenCyril:
@thery that is true, it works even if there is no `_ < _` in the goal. Before the above mentioned commit, it used to work when there was a `_ == _` in the goal, and after it it also work when the goal contains `_ <= _`. So it does not apply only for `_ < _` but also for more situations, and indeed, a bit more than before.
What I mean, is that this theorem is not becoming misnamed by this commit. And it was not misnamed before, as it reflects the three case that are generated by a case analysis on this lemma.

thery:
> What I mean, is that this theorem is not becoming misnamed by this 
> commit. And it was not misnamed before, as it reflects the three case 
> that are generated by a case analysis on this lemma.
>
> —
>


[...]

PR #270, opened 2019-01-18
Dispatching order and norm, and anticipating normed modules.

CohenCyril:
### Motivation and consequences

The motivation for this PR is to bring to mathcomp dedicated order structures inside the hierarchy, hence facilitating developments using ordered structures which would be currently using the `coq-mathcomp-finmap` package (e.g. multinomials, which we plan to integrate into core mathcomp). Such a refactoring also make it possible to define algebraic structures where the norm and order are not entangled, such as normed modules, which would otherwise need a duplicate of the norm, its notations, and part of its theory (which is a problem we have in mathcomp/analysis).

Hence this PR is to integrate `order.v` from of http://github.com/math-comp/finmap to mathcomp and propagate the changes to `ssrnum`. This consists in the introductions of several new structures in the hierarchy (order, partial orders, lattices and normed abelian groups), all inheriting at least `choiceType` and some inherited by `numDomainType` which now takes its `norm` and `order` from two dedicated structures. A huge part of this PR is the introduction of order and lattice structures and their theory. The other part is the refactoring consiting in reducing the former `Num.mixin_of` using the new mixins, classes and structures and adapting the library (adding "trivial" `Canonical` instance declaration here and there) and performing some renamings (which could be avoided by importing the compatibility module `ssrnum.mc_1_9`).

I (@CohenCyril) think this work has converged pretty close to its final shape and might still contain a few bugs that can be fixed later, and several improvements can be envisioned after the merge of this PR.

### Possible things to review

Please review _as much as you can_ contents of `order.v` and `ssrnum`, and the modifications it entailed in many files in the library, including but not limited to,
- the shape and consistency of notation,
- the naming of lemmas (clarity, consistency, etc...),
- if the statements of theorems looks ok,
- check for obviously missing lemmas,
- the readability, soundness and (sufficient-)completeness of header or in-file documentation.

### Test

The only non technical solution I know is to download and compile a local copy of this branch by hand, using make and make install (possibly in a dedicated opam switch). The branch https://github.com/math-comp/analysis/tree/analysis_270 contains a experiment rebasing and refactoring all of mathcomp/analyis on top of this branch.

### Summary of changes
(by @CohenCyril  and by @pi8027 cf [this commit log](https://github.com/math-comp/math-comp/pull/270/commits/ba95a079160a8e87ba7ea01efeb9274b4c2007ee))

ssrnum related changes:
- Redefine the intermediate structure between `idomainType` and `numDomainType`,
  which is `normedDomainType` (normed integral domain without an order).
- Generalize (by using `normedDomainType` or the order structures), relocate
  (to order.v), and rename ssrnum related definitions and lemmas.
- Add a compatibility module `Num.mc_1_9` and export it to check compilation.
- Remove the use of the deprecated definitions and lemmas from entire theories.
- Implement factories mechanism to construct several ordered and num structures
  from fewer axioms.

Order related changes:
- Included order library from finmap, which contained partial and total order and distributive lattices structures (with or without top, bottom and complement) with an extra `unit` flag in structures, to toggle different printings of notations depending on which instances are used.
- Reorganize the hierarchy of finite lattice structures. Finite lattices have
  top and bottom elements except for empty set. Therefore we removed finite
  lattice structures without top and bottom.
- Reorganize the theory modules in order.v:
  + `LTheory` (lattice and partial order, without complement and totality)
  + `CTheory` (`LTheory` + complement)
  + `TTheory` (`LTheory` + total)
  + `Theory` (all)
- Give a unique head symbol for `Total.mixin_of`.
- Replace `reverse` and `^r` with `converse` and `^c` respectively.
- Fix packing and cloning functions and notations.
- Provide more ordered type instances:
  Products and lists can be ordered in two different ways: the lexicographical
  ordering and the pointwise ordering. Now their canonical instances are not
  exported to make the users choose them.
- Export `Order.*.Exports` modules by default.
- Specify the core hint database explicitly in order.v. (see #252)
- Apply 80 chars per line restriction.

General changes:
- Give consistency to shape of formulae and namings of `lt_def` and `lt_neqAle`
  like lemmas:
  `lt_def    x y : (x < y) = (y != x) && (x <= y)`,
  `lt_neqAle x y : (x < y) = (x != y) && (x <= y)`.
- Enable notation overloading by using scopes and displays:
  + Define `min` and `max` notations (`minr` and `maxr` for `ring_display`) as
    aliases of `meet` and `join` specialized for `total_display`.
  + Provide the `ring_display` version of `le`, `lt`, `ge`, `gt`, `leif`, and
    `comparable` notations and their explicit variants in `Num.Def`.
  + Define 3 variants of `[arg min_(i < n | P) F]` and `[arg max_(i < n | P) F]`
    notations in `nat_scope` (specialized for nat), `order_scope` (general
    version), and `ring_scope` (specialized for `ring_display`).
- Updated documentations and `CHANGELOG_UNRELEASED.md` entries.

### Automatic note to reviewers

EDIT @CohenCyril 2019-06-13


Zimmi48:
@erikmd or @CohenCyril to avoid the duplicate status checks, you can disable the GitHub integration in "Settings" / "Integration" in the GitLab repo.

erikmd:
Hi @Zimmi48, indeed I realized a couple of minutes ago that I had forgotten to disable the "native GitLab/GitHub integration" for branches reporting as it is now unnecessary given @coqbot's help. This is now done.
Thanks!

erikmd:
Finally regarding Travis, [the script run by Travis CI](https://github.com/math-comp/math-comp/blob/b0ef24be9fe47abbcc2c82e668091d4b32283c8b/.travis.yml#L41-L43) is very similar as [the one](https://github.com/math-comp/math-comp/blob/b0ef24be9fe47abbcc2c82e668091d4b32283c8b/Dockerfile.make#L26-L29) run by [this GitLab CI job](https://github.com/math-comp/math-comp/blob/b0ef24be9fe47abbcc2c82e668091d4b32283c8b/.gitlab-ci.yml#L85), so maybe it's not worth it to also rely on Travis for each CI build?  @CohenCyril what do you think?
− I guess the URL to configure this is [that one](https://travis-ci.org/math-comp/math-comp/settings).
Kind regards, Erik

pi8027:
Hi @CohenCyril, I had some progress on this PR. I need your help to fix `realDomainType`. Is `real_axiom` equivalent to the totality of `ler` and `ltr` for any `numDomainType`? If it is, should `realDomainType` be coercible to `orderType`? https://github.com/math-comp/math-comp/compare/experiment/order...pi8027:b7ff2ca

pi8027:
Progress report: I have fixed all the structures in `ssrnum` and added all the missing coercions and canonical projections. I haven't ported `normedModType` and `NumDomain.pack` is still broken. https://github.com/math-comp/math-comp/compare/experiment/order...pi8027:978f25a

pi8027:
@CohenCyril I finished to port `normedModType` but the following theorem doesn't seem to be held.
```
Lemma normr0 (R : numDomainType) (V : normedModType R) : `|0 : V| = 0.
```

A change for the 3rd axiom of `NormedModule.mixin_of` as follows would be needed. How do you think? https://github.com/math-comp/math-comp/compare/experiment/order...70250a5

```
  _ : forall x : T, (norm x == 0) = (x == 0)}.
```

thery:
just a naive question, is (norm x == 0) = (x == 0) not a consequence of (le_op x y) = (norm_op (y - x) == y - x)} for x= y?

CohenCyril:
@pi8027 no, unless I'm mistaken, since `0 <= 0`, we have that `'|0 : R| = 0` (by the last axiom of `numDomainType`) and then `'|0 : V| = '|0 *: 0| = '|0 : R| * '|0 *: V|` using the 2nd axiom of `normedModType` and you should be done.

pi8027:
@CohenCyril I did it. Thank you!

pi8027:
How can I prove this, or is this not needed for `normedModType`?

```
Lemma normrN (V : normedModType R) (x : V) : `|- x| = `|x|.
```

pi8027:
@CohenCyril I have a problem with canonical structures. I think that `F : numFieldType` should be coercible to `normedModType F` but it isn't. So `rewrite normr_eq0` and `rewrite (@normr_eq0 F F)` fail for `numFieldType`s. https://github.com/pi8027/math-comp/blob/ecb6dc515f9e0def5d541738270c1ba0c79d0afb/mathcomp/algebra/ssrnum.v#L3453

I might have made a mistake on the definition of the canonical `normedModtype` instance of `numDomainType`. https://github.com/pi8027/math-comp/blob/ecb6dc515f9e0def5d541738270c1ba0c79d0afb/mathcomp/algebra/ssrnum.v#L1071

CohenCyril:
@pi8027 In fact, there should be both coercions and canonical structures to each of the following structures: `numDomainType`, `numFieldType`, `realDomainType`, `realFieldType`, `archiRealType`, `rcfType` and `numClosedFieldType`.

pi8027:
I have written a compatibility module for `ssrnum` and attempted to fix `ssrint` with use of that module, but failed at [here](https://github.com/pi8027/math-comp/blob/1a89b19fd8ba5719c8a0563c616f32d5b08b89f4/mathcomp/algebra/ssrint.v#L1556) with the following error message:

```
Error:
In nested Ltac calls to "by (ssrhintarg)", "case (ssrcasearg) (ssrclauses)" and
"(normrN m)", last term evaluation failed.
In environment
m : int
The term "m" has type "int" while it is expected to have type "Num.NormedModule.sort ?V".
```

thery:
@pi8027  
Looks like the canonical mechanism is not able to lift an int into a Num.NormedModule.sort
If you  do 
Print Canonical Projections.
can you see your int_normedModType?

pi8027:
@thery You seem to be correct. I saw the following line in the result of `Print Canonical Projections`. There is no other line including `int_normedModType`.
```
Num.NumDomain.sort <- Num.NormedModule.sort ( int_normedModType )
```

CohenCyril:
@pi8027 you need to use the `NormedModType` packager with explicit argument `int`, otherwise you won't get the right head symbol for the projection `Num.NormedModule.sort`. (Btw you need canonicals for normedType and `lmodType`

thery:
@pi8027 canonical structures are some kind of black magic. It would be nice to automate in some way all this boilerplate.

pi8027:
@CohenCyril Thank you. It was solved!

pi8027:
@thery Yes, indeed. I need such the mechanism. :D

pi8027:
I found the following 2 problems.

- The lemma `normr_ge0` is added to the hint database `core`. But it was sometimes not applied by `done` tactic when I need it because the goal doesn't syntactically match to the conclusion of the hint. This has appeared in #283.
  ```
  normr_ge0
       : forall (R : numDomainType) (V : normedModType R) (v : V), 0 <= `|v|
  ```
- This was failed:
  ```
  Lemma normr_ge0_numDomain (R : numDomainType) (v : R) : 0 <= `|v|.
  Proof. Fail apply: normr_ge0. Abort.
  ```

CohenCyril:
> The lemma `normr_ge0` is added to the hint database `core`. But it was sometimes not applied by `done` tactic when I need it because the goal doesn't syntactically match to the conclusion of the hint. This has appeared in #283.

Hi! do you still have this problem?

pi8027:
@CohenCyril Yes. Sometimes I need to put `?normr_ge0` in rewrite tactic.

CohenCyril:
> @CohenCyril Yes. Sometimes I need to put `?normr_ge0` before `//` in rewrite tactic.

Try `Hint Extern 0 (is_true (_ <= _)) => now apply: normr_ge0`

pi8027:
@CohenCyril It does work well. Thank you!

CohenCyril:
Could you export the compatibility module, just to check whether it suffices to make fourdolor and odd-order compile?

pi8027:
@CohenCyril OK. I'm checking it in #286.

pi8027:
We have almost finished this PR and so is it time to check some missing things and TODOs? My concern is as follows:
- `min` and `max` notations for `orderType`s would be needed (`meet` and `join` is a bit confusing in this case...),
- naming of the following lemmas needs review: `(le|lt|lte)(xI|Ix|xU|Ux)`, `leIx_total`, and `lexU_total`,
- the subterm selection problem in rewriting (see also #280),
- I haven't understood the intents of some comments in ssrnum...

CohenCyril:
> * `min` and `max` notations for `orderType`s would be needed (`meet` and `join` is a bit confusing in this case...),

I totally agree! Let us put them both in the `total_display` (to be defined)

> * naming of the following lemmas needs review: `(le|lt|lte)(xI|Ix|xU|Ux)`, `leIx_total`, and `lexU_total` in order.v,

I suggest:
```coq
Lemma gel_leIx x y z : y <= x -> y `&` z <= x.
Lemma ger_leIx x y z : z <= x -> y `&` z <= x.
Lemma ge_leIx x y z : (y <= x) || (z <= x) -> y `&` z <= x.
Lemma leIx x y z : (y `&` z <= x) = (y <= x) || (z <= x).
```
and the "same" for `U`

CohenCyril:
> I suggest:
> 
> ```coq
> Lemma gel_leIx x y z : y <= x -> y `&` z <= x.
> Lemma ger_leIx x y z : z <= x -> y `&` z <= x.
> Lemma ge_leIx x y z : (y <= x) || (z <= x) -> y `&` z <= x.
> ```

or maybe 

```coq
Lemma leIxl x y z : y <= x -> y `&` z <= x.
Lemma leIxr x y z : z <= x -> y `&` z <= x.
Lemma leIx2 x y z : (y <= x) || (z <= x) -> y `&` z <= x.
```

pi8027:
@CohenCyril I prefer `leIx(l|r|2)`. Thanks!

pi8027:
@CohenCyril `finPOrderType` and `finLatticeType` are undocumented. Should `finLatticeType`s have canonical `tblatticeType` instances? I'm not sure.

pi8027:
The remaining work would be only min and max notations if PR #288 is successfully merged.

mathcomp/character/character.v
```diff
@@ -1502,7 +1501,7 @@ Qed.
 Lemma max_cfRepr_mx1 n (rG : mx_representation algCF G n) x :
    x \in G -> cfRepr rG x = cfRepr rG 1%g -> rG x = 1%:M.
 Proof.
-move=> Gx kerGx; have [|c _ def_x] := @max_cfRepr_norm_scalar n rG x Gx.
+move=> Gx kerGx; have [|c _ def_x] := @ max_cfRepr_norm_scalar n rG x Gx.
```
pi8027: I needed to put a space here to distinguish from `@max` notation defined in order.v. It looks so bad, but I have no solution for this.

ggonthier: I object to this. The `@` symbol is reserved in the Coq syntax to mean explicit application, and the MathComp library should respect this and never create tokens of the form `@ident`. The only acceptable use of this kind of notation is to provide an explicit variant of an abbreviation, and it should be given as `"@ 'ident'"` at level 0 or `"@ 'ident' X Y..."` at level 10 so as to factor with the Coq grammar - see `ssrfun.v` for example. Caveat: I can't tell what the notation was meant for since it's not in the PR version of `order.v`.

pi8027: `@max` and `@min` are the explicit variants of overloaded max/min operators for totally ordered sets (`orderType`). Therefore as you pointed out they should be given as `"@ ''max'"` and `@ "min"`. I didn't know this fact. Thank you so much.

pi8027: Fixed in #288

mathcomp/algebra/ssrnum.v
```diff
@@ -5143,12 +4862,501 @@ End RealMixin.
...
+Notation RealLtPoMixin := Num.RealMixin.LtPo.
 Notation RealLeMixin := Num.RealMixin.Le.
 Notation RealLtMixin := Num.RealMixin.Lt.
 Notation RealLeAxiom R := (Num.RealMixin.Real (Phant R) (erefl _)).
+Notation RealLeTotal R := (Num.RealMixin.Total (Phant R) (erefl _)).
 Notation ImaginaryMixin := Num.ClosedField.ImaginaryMixin.
+
+(* compatibility module *)
+Module mc_1_7.
+Module OrigNum := Num.
```
pi8027: To export the contents of the `ssrnum.Num` module in the `ssrnum.mc_1_7.Num` compatibility module without disabling interactive edit in PG and CoqIDE (see https://github.com/coq/coq/issues/8388), I have putted an alias of `ssrnum.Num` here. But this alias sometimes pollutes Coq responses. For example:
```
Coq < Check (0 <= 1)%R.
0 <= 1
     : bool
where
?t : [ |- mc_1_7.OrigNum.NumDomain.Exports.numDomainType]
```

```
Coq < Print Canonical Projections.
...
int <- mc_1_7.OrigNum.NumDomain.sort ( int_numDomainType )
...
```

Removing this line and replacing all the `OrigNum` with `ssrnum.Num` solve this problem but disable interactive edit.

pi8027:
The problem mentioned in the above review breaks my tool for drawing the hierarchy diagram (#290). But somehow I succeed to generate the correct hierarchy diagram. This diagram would be useful to make this development comprehensible: [hierarchy.pdf](https://github.com/math-comp/math-comp/files/2901758/hierarchy.pdf).

mathcomp/algebra/ssrnum.v
pi8027: I found that some solutions have been provided for https://github.com/coq/coq/issues/8388, but we need to wait for the release of Coq 8.10 for making everything working well...

mathcomp/algebra/ssrnum.v
```diff
@@ -156,45 +157,87 @@ Reserved Notation ">= y :> T" (at level 35, y at next level).
...
+Fact ring_display : unit. Proof. exact: tt. Qed.
+
 Module Num.
 
-(* Principal mixin; further classes add axioms rather than operations. *)
-Record mixin_of (R : ringType) := Mixin {
-  norm_op : R -> R;
-  le_op : rel R;
-  lt_op : rel R;
+Record mixin_of (R : ringType) (Rorder : porderMixin R)
```
CohenCyril: @ggonthier suggestion: package on the fly to use nice symbols.

pi8027: Is that really possible? There are no structure sharing ring operations, order, and norm before this declaration.

CohenCyril: Yes, one could create a section with a variable `R : ringType` and axioms `pom : porderMixin R` and `norm_op : R -> R` do a `Let T := (R : Type)` and provide T with eq, zmod, ring, order and norm canonical structures.

pi8027:
Why `NumClosedField` appears twice in the documentation of ssrnum? Which one is correct?

CohenCyril:
> I'm trying to update the documentation of order and ssrnum. Why `NumClosedField` appears twice in the documentation of ssrnum? Which one is correct?

:laughing: the one line 31 is incorrect.

mathcomp/algebra/ssrnum.v
pi8027: I attempted, but failed because of this issue: https://github.com/coq/coq/issues/9764.

pi8027: I also needed to put `Notation "0" := (0 : T).` because `@lt _ _ 0%R` cannot be typed before declaring a join structure of `porderType` and `zmodType`.

pi8027: As a conclusion, I prefer to keep this as it is and improve later.

mathcomp/ssreflect/order.v
```diff
@@ -0,0 +1,3391 @@
...
+(* (at your option) any later version.                                   *)
+(*                                                                       *)
+(* This program is distributed in the hope that it will be useful,       *)
+(* but WITHOUT ANY WARRANTY; without even the implied warranty of        *)
+(* MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the         *)
+(* GNU General Public License for more details.                          *)
+(*                                                                       *)
+(* You should have received a copy of the GNU General Public License     *)
+(* along with this program.  If not, see <http://www.gnu.org/licenses/>. *)
+(*************************************************************************)
```
pi8027: This header would be needed to be changed because 1. order.v is not a draft anymore and 2. MathComp is distributed under CeCILL-B. @CohenCyril Could you please correct this as you wish?

pi8027:
I have no more things to do on this PR now; it is ready for review.

mathcomp/algebra/ssrnum.v
```diff
@@ -5082,73 +4861,690 @@ rewrite -normN ge0_norm //; have [hxy|hxy] := leN_total (x + y).
...
+Arguments nmono_in_lerif [R A f C].
+Arguments mono_lerif [R f C].
+Arguments nmono_lerif [R f C].
+
+End Theory.
+
+End Num.
+End mc_1_7.
+
+Export mc_1_7 mc_1_7.Num.Syntax.
```
pi8027: This line should be removed before merging.

pi8027:
I will rebase this tomorrow. Some inheritance bugs might be found. (see #318)

pi8027:
I have attempted to add `unitAlgType -> numDomainType` inheritance by following some discussions with @CohenCyril . Now `algC.v` has a performance problem (compilation is very slow), and the join of `algType` and `comRingType` isn't unique: `numDomainType` and `fieldExtType`. [hierarchy.pdf](https://github.com/math-comp/math-comp/files/3069257/hierarchy.pdf)

CohenCyril:
> Redefining `idomainType` as the join of `comUnitRing` and `unitAlgType` might solve the second issue.

this is one is not desirable

ggonthier:
None of these are desirable; there's a reason the regular module structure on rings is only inferred for a specific tag, and the linear structure does _not_ require a module structure on its codomain.

CohenCyril:
> None of these are desirable; there's a reason the regular module structure on rings is only inferred for a specific tag, and the linear structure does _not_ require a module structure on its codomain.

Yes, yes, we decided to backtrack on that, and @pi8027 is now removing the inheritance of Lmodules everywhere, to makes the paths shorter. The common superstructure of `numDomainType` and `normedModule` (required in math-comp/analysis) will be a `NormedDomain`.

CohenCyril:
@thery 
> @pi8027 canonical structures are some kind of black magic. It would be nice to automate in some way all this boilerplate.

we are on it, and we have some ideas to make sure our users are less impacted by hierarchy refactorings that before, by providing systematic higher level blocks (factories, independent of the hierarchy, **public**) that are independent of basic blocks (mixins, dependent on the specific way the hierarchy is generated, **private**).

pi8027:
Removing the `norm_scale` axiom from `NormedDomain` breaks some properties (e.g., `normr0`, `normrN`, `normr_id`). The following axioms might be needed.
```
  _ : forall n x, norm (x *+ n) = norm x *+ n;
  _ : forall x, norm (- x) = norm x;
```

CohenCyril:
> Removing the `norm_scale` axiom from `NormedDomain` breaks some properties (e.g., `normr0`, `normrN`, `normr_id`). The following axioms might be needed.
> 
> ```
>   _ : forall n x, norm (x *+ n) = norm x *+ n;
>   _ : forall x, norm (- x) = norm x;
> ```

yes!

pi8027:
Remarks:
- Now `[arg min_( i < n | P ) F]` and `[arg max_( i < n | P ) F]` notations are duplicated in `fintype.v` and `order.v`. We would need to make `fintype.v` and `bigop.v` depend on `order.v` and reshuffle some definitions and theorems to solve this duplication. This probably should be done in a separate PR.
- Applying order related theorems (including `arg_minP` and `arg_maxP`) for `nat` requires rewriting by `leEnat` and `ltEnat`. Using overloaded ordering relation everywhere would solve this, but this may cause another difficulty because many proofs depend on that `(_ <= _)%N` and `(_ < _)%N` have the same head symbol `leq`...
  ```coq
  Lemma leEnat: le = leq.
  Lemma ltEnat (n m : nat): (n < m) = (n < m)%N.
  ```

CohenCyril:
> Remarks:
> 
> * Now `[arg min_( i < n | P ) F]` and `[arg max_( i < n | P ) F]` notations are duplicated in `fintype.v` and `order.v`. We would need to make `fintype.v` and `bigop.v` depend on `order.v` and reshuffle some definitions and theorems to solve this duplication. This probably should be done in a separate PR.
> * Applying order related theorems (including `arg_minP` and `arg_maxP`) for `nat` requires rewriting by `leEnat` and `ltEnat`. Using overloaded ordering relation everywhere would solve this, but this may cause another difficulty because many proofs depend on that `(_ <= _)%N` and `(_ < _)%N` have the same head symbol `leq`...
>   ```coq
>   Lemma leEnat: le = leq.
>   Lemma ltEnat (n m : nat): (n < m) = (n < m)%N.
>   ```
>   
>   
>   I prefer to keep this as it is and to use `leEnat` and `ltEnat`.

@pi8027 I think the theories of generic order on one side and `leq` on the other should be kept separate anyway. Hence we should/must have all the necessary duplicates, in the right scopes so as never to get a generic `le` or `lt` by using a theorem on `leq` or `_ < _`.

pi8027:
@CohenCyril Both polymorphic and monomorphic `arg_min`/`arg_max` notations should be defined in the `form_scope` (isn't it?). So it's not possible to use both notations selectively unless defining them with different keywords.

pi8027:
Using the polymorphic definitions and notations of `arg_min`/`arg_max` in both cases, deprecating monomorphic ones, and using the following lemmas in the cases of `nat` might be a satisfactory solution.

```coq
arg_minnP
     : forall (I : finType) (i0 : I) (P : pred I) (F : I -> nat),
       P i0 -> extremum_spec leq P F [arg min_(i < i0 | P i) F i]
arg_maxnP
     : forall (I : finType) (i0 : I) (P : pred I) (F : I -> nat),
       P i0 -> extremum_spec geq P F [arg max_(i > i0 | P i) F i]
```

CohenCyril:
@ggonthier do you see any objection for us to put `[arg min_(i < i0 | P i) F i]` both in the `nat_scope` and `order_scope` (depending on whether it is a `leq` or generic `le`), rather than only once in `form_scope`? The alternative is to rename the one for nat like this: `[arg minn_(i < i0 | P i) F i]`.

ggonthier:
No objections, @CohenCyril  - that seems a fairly natural overloading.

pi8027:
OK. So I will put `[arg (min|max)_...]` notations in `nat_scope`, `order_scope`, and `ring_scope`.

pi8027:
@CohenCyril Is this change fine? If it is, I will rebase and squash this and make use of `deprecate` notation in the compatibility module. https://github.com/pi8027/math-comp/commit/68c697174082268bbfe9c81d7bcdc9752c5fd088

CohenCyril:
> @CohenCyril Is this change fine? If it is, I will rebase and squash this and make use of `deprecate` notation in the compatibility module. [pi8027@68c6971](https://github.com/pi8027/math-comp/commit/68c697174082268bbfe9c81d7bcdc9752c5fd088)

Looks good to me.

CohenCyril:
@amahboubi @ggonthier @gares this PR is ready for review.

CohenCyril:
@amahboubi @ggonthier @gares I am currently fixing several issues in `order`, it should be finished by tomorrow.

CohenCyril:
Ok I fixed almost everything I spotted.

But, @ggonthier, @amahboubi, @thery, @gares, @LaurenceRideau, @pi8027  there is an issue for which I cannot take a light decision: `ltgtP` ... do not follow the same structure as the `ltngtP` lemma from ssrnat:
in `ltgtP` (and `ltrgtP`) lemma, the operators come in the following order `==`, `<=`, `<` while in `ltngtP` it is `<=`, `<` and `==`. For consistency they should be ordered in the same way, and thus there are now two options:
1. For backward compatibility they should be in the order  [`<=`, `<`, `==`] (as in `ltngtP`),
2. but for the sake of intuition as noted by @thery in #233 it should rather be [`==`, `<=`, `<`] everywhere (as in `ltgtP`) because the last item is matched first. I think I a missing a part of the history of the lemma in order to take a decision.

Which of 1. and 2. would be best?

Additionally
- in `ltgtP` and `ltrgtP` the forms `y <= x` should come before the forms `x <= y` (to match the last one first), I will fix this as soon as we take a decision on the former.
- in `leqP`, `leP` ..., operators are in the order `<=` first, and `<` while the same intuition as in the paragraph above would lead to think it should be the converse.

gares:
IMO 2 is best

pi8027:
@CohenCyril It's worth trying 2.

pi8027:
I found that option 2 (reordering the type arguments of `ltngtP` to `[==, <=, <]`) doesn't break any libraries other than fourcolor. We can provide a compatibility module so this is not so hard thing to do. (see #377)

pi8027:
@amahboubi @ggonthier @gares @CohenCyril I think all the problems we found have been addressed. I have also patched fourcolor (math-comp/fourcolor#19), odd-order (math-comp/odd-order#17), and real-closed (math-comp/real-closed#16) without any use of the compatibility module. CI checks with these overlays have passed. So this PR is ready for review.

CohenCyril:
@amahboubi @ggonthier @gares @thery @LaurenceRideau @strub because of the size of the PR, the time it took to develop and the potential impact on the analysis library and other users, I would suggest of all mathcomp PR we give this one the highest priority for reviewing and integration. What do you think? If you have any question on how to proceed, you can tag, mail or call me to discuss it.
Best wishes!

amahboubi:
Hi @CohenCyril @pi8027! It is great to see this PR converging. In order to ease review and subsequent usage/documentation of the library, I suggest to have this PR as an item of our first mathcomp dev meeting of the year (I will send the message+poll today). It would be great to have a short presentation summarizing the previous issues and the solutions.

amahboubi:
@CohenCyril, I have unassigned you this PR, as it's no good practice to be at the same time main author and assignee.

pi8027:
There are two name conflicts of lemmas in `order.v`.
```coq
Locate lexUl.
Locate lexUr.
```
```
Constant mathcomp.ssreflect.order.Order.BLatticeTheory.lexUl
Constant mathcomp.ssreflect.order.Order.LatticeTheoryJoin.lexUl
  (shorter name to refer to it in current context is Order.LatticeTheoryJoin.lexUl)

Constant mathcomp.ssreflect.order.Order.BLatticeTheory.lexUr
Constant mathcomp.ssreflect.order.Order.LatticeTheoryJoin.lexUr
  (shorter name to refer to it in current context is Order.LatticeTheoryJoin.lexUr)
```
Symmetric ones of `Order.LatticeTheoryJoin.lexU(l|r)` are named `leIx(l|r)` so we probably should rename them to `leUx(l|r)`. I also wish to have a Coq plugin to detect name conflicts of imported unqualified names.

affeldt-aist:
TODO: in minr_l, minr_r, maxr_r, maxr_l (ssrnum.v), introduce lemmas to avoid (elimT ???) definitions

pi8027:
[Note] I will do the following renamings to avoid the name conflicts of https://github.com/math-comp/math-comp/pull/270#issuecomment-534985132:
- `Order.BLatticeTheory.lexUl` -> `disjoint_lexUl`,
- `Order.BLatticeTheory.lexUr` -> `disjoint_lexUr`,
- `Order.TBLatticeTheory.lexIl` -> `cover_leIxl`,
- `Order.TBLatticeTheory.lexIr` -> `cover_leIxr`.

In a discussion happened today, I discovered that it would be nice to have the canonical `fin(C?)latticeType` instance of finsets (or fsets?).

pi8027:
> Hi @CohenCyril @pi8027! It is great to see this PR converging. In order to ease review and subsequent usage/documentation of the library, I suggest to have this PR as an item of our first mathcomp dev meeting of the year (I will send the message+poll today). It would be great to have a short presentation summarizing the previous issues and the solutions.

Here Is a summary:
- The subterm selection problem (typically in rewriting).
  - Currently, we sometimes need to write patterns explicitly.
- `Hint Resolve` doesn't work to declare `normr_ge0` as a hint.
  - Use `Hint Extern` instead.
- Since the type of norm function is now `forall (R : numDomainType) (V : normedZmodType R), V -> R`,
  sometimes the type `V` cannot be inferred from context, e.g., `|0| = 0 :> R`.
  - We need to put type annotations differently, e.g., `|0 : R| = 0`.
- Hierarchy related issues
  - Nonempty finite lattices always have bottom and top.
    So we equated `fin(C)LatticeType`, `fin(C)BLatticeType`, and `fin(C)TBLatticeType`.
  - Defining `normedModType` to inherit `lmodType` introduces an ambiguous join.
    Now we redefined it as `normedZmodType` that inherit `zmodType` and is parameterized by `numDomainType`.
    We cannot state some interactions between the norm and scale functions on that structure.
  - Other several inheritance issues are solved by `hierarchy.ml`, which is now a part of CI.
- Arguments ordering problem of comparison predicates, e.g., `ltngtP`.
  - Will be separately addressed by #378.
- Defining factories sometimes requires explicit record-eta expansion.

CohenCyril:
In addition, I edited the orignal post to try and input more high level explanations.

amahboubi:
Can you elaborate on what
> (which could be avoided by importing the compatibility module ssrnum.mc_1_9).
means? Where is this compatibility module?

amahboubi:
> This work started by @CohenCyril and continued by @pi8027.

I suggest removing this line. This is not fair to the other contributors, in locs or in discussions. E.g. the finmap math-comp repo acknowledges several influences, and in that repo, a file like [order.v](https://github.com/math-comp/finmap/blob/master/order.v) has received commits from more contributors.

CohenCyril:
> I suggest removing this line. This is not fair to the other contributors, in locs or in discussions. E.g. the finmap math-comp repo acknowledges several influences, and in that repo, a file like [order.v](https://github.com/math-comp/finmap/blob/master/order.v) has received commits from more contributors.

done

affeldt-aist:
For information, it was possible to use this PR to rebase mathcomp-analysis on top of it, which means essentially to use the norm provided by ssrnum.v to subsume the several notions of norms that were used in the current version of mathcomp-analysis. This went well except for problems that have been reported and could be quickly fixed. This rebasing led to substantial simplifications in mathcomp-analysis. See https://github.com/math-comp/analysis/tree/analysis_270.

amahboubi:
TODO list on order.v:
- header:  remove misleading sentences, list the notations and definitions introduced by the library
- header: document better the tagging mechanism allowing to work with different order structures on a same type. This could be in the header and/or in later in the file
- remove the no-op last Import
- header: improve the documentation of what you get by importing selected of sub-modules 
- BTW do we need Order.Def?
- What about importing Order.Syntax in all of the subsequent Order.LTheory, Order.TTheory, etc. Note: a Syntax module is only useful to solve potential grammar conflicts. Which does not seem to be the case here.
- ensure all notations do have a scope
- declare instances for division on nat inside this file, and use this as a documentation/template

amahboubi:
Timline:
- TODO list addressed for Oct. 25th
- Merge <= Oct. 31st.

pi8027:
I found one more TODO. Factories in `ssrnum` are undocumented.

pi8027:
> * BTW do we need Order.Def?

I think `Order.Def` is used only for this kind of explicit application: https://github.com/math-comp/math-comp/pull/270/files#diff-f9ae64f5e242216de4bf5a674b57fd36R371-R394. Having only *qualified* names such as `Order.le`, `Order.lt`, `Order.ge`, `Order.gt` might be enough for this.

> * What about importing Order.Syntax in all of the subsequent Order.LTheory, Order.TTheory, etc. Note: a Syntax module is only useful to solve potential grammar conflicts. Which does not seem to be the case here.

Since `Num.Syntax` is exported by default, doing the same for `Order.Syntax` might be better. Actually it works.

CohenCyril:
> I think `Order.Def` is used only for this kind of explicit application:

This can be done using scopes rather than modules and it should...
All of these notations should be in ring_scope.

CohenCyril:
> > I think `Order.Def` is used only for this kind of explicit application:
> 
> This can be done using scopes rather than modules and it should...
> All of these notations should be in ring_scope.

Oh no, it cannot... but simple notations are prefixed with their module and we maybe can "hack" the other ones to include it `"@ 'Num.le' R"`

pi8027:
> Oh no, it cannot... but simple notations are prefixed with their module and we maybe can "hack" the other ones to include it `"@ 'Num.le' R"`

Anyway, we need to make symbols like `Order.Def.lt` qualified or disambiguated with a suffix. Just making them qualified can be done by turning `Def` modules such as `Order.POrderDef` to sections and removing the `Order.Def` module. In this way, we don't need that hack.

amahboubi:
Naive question : what is this block of notation without a scope?

pi8027:
@amahboubi `Num.Def.(ler|ltr|...)` were `Definition`s before this change. We need to redefine them to make the new `ssrnum` library compatible to a degree with the old one.
https://github.com/math-comp/math-comp/blob/3ad8a9418dc5d39ef553033f5c8f581525a3aae3/mathcomp/algebra/ssrnum.v#L241-L263 In order to use order theories to reason about num structures, `<=%R` (`Num.Def.ler`) should have the same head symbol as `<=%O` without any reduction; thus, `Num.Def.(ler|ltr|...)` should be redefined as notations rather than definitions. Indeed `Notation ler := ...` things cannot be defined in a specific scope. `Notation "@ 'ler' R" := ...` things are declared for explicit application but their displays are fixed to `ring_display` (see https://github.com/math-comp/math-comp/pull/270#discussion_r259755442).

hivert:
@CohenCyril asked me offline to comment here the naming convention. Currently `latticeType` are actually distributive lattice. It would be very uncommon in my community (combinatorics) to assume that all lattice are distributive.  So I'd rather have them called `distrLatticeType` or maybe `dlatticeType`. Note: since it's not something you'll write every line, my vote goes for the long name.

pi8027:
@hivert Could you confirm this renaming? Currently we don't have complemented non-distributive lattice structures in #388 and I'm not sure how interesting it is and that we should add them later.
```
(*           porderType == the type of partially ordered types                *)
(*     distrLatticeType == the type of distributive lattices                  *)
(*    bDistrLatticeType == latticeType with a bottom element                  *)
(*   tbDistrLatticeType == latticeType with both a top and a bottom           *)
(*   cbDistrLatticeType == the type of sectionally complemented distributive  *)
(*                         lattices                                           *)
(*                         (lattices with bottom and a difference operation)  *)
(*  ctbDistrLatticeType == the type of complemented distributive lattices     *)
(*                         (lattices with top, bottom, difference, complement)*)
(*            orderType == the type of totally ordered types                  *)
(*        finPOrderType == the type of partially ordered finite types         *)
(*  finDistrLatticeType == the type of nonempty finite distributive lattices  *)
(* finCDistrLatticeType == the type of nonempty finite complemented           *)
(*                         distributive lattices                              *)
(*         finOrderType == the type of nonempty totally ordered finite types  *)
```

hivert:
That would be indeed my favorite naming convention. Note that I'm note sure that `sectionally complemented lattice` is a standard name. It looks that it the same as `relatively complemented distributive lattice` according to wikipedia. It moreover mention that "A distributive lattice is complemented if and only if it is bounded and relatively complemented."...

pi8027:
@hivert Sure. But currently, we don't have relative complement operation "the complement of `x` in the interval `[y, z]`" in the library, which can easily be constructed from sectional complement operation "the complement of `x` in the interval `[0, y]`".

pi8027:
@CohenCyril I think you lost my last commit during the last rebasing: https://github.com/pi8027/math-comp/commit/1a2998ff78007bb508d216024864559d9882fc32. I rebased math-comp/fourcolor#19 so the CI failure on fourcolor will be disappeared.

pi8027:
I think we fixed all the problems discovered recently. I hope we can merge this after the next MathComp meeting!

amahboubi:
This had been written long ago, but somehow I forgot to publish my review, sorry...

mathcomp/ssreflect/order.v
```diff
@@ -0,0 +1,5974 @@
...
+(*   Order.CTheory: complemented lattices including Order.LTheory.            *)
+(*   Order.TTheory: totally ordered types including Order.LTheory.            *)
+(*    Order.Theory: ordered types including all of the above theory modules   *)
+(*                                                                            *)
+(* To access the definitions, notations, and the theory from, say,            *)
+(* "Order.Xyz", insert "Import Order.Xyz." at the top of your scripts.        *)
+(* Notations are accessible by opening the scope "order_scope" bound to the   *)
+(* delimiting key "O".                                                        *)
+(*                                                                            *)
+(* We provide the following structures of ordered types                       *)
```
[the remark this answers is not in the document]
CohenCyril: addressed

mathcomp/ssreflect/order.v
```diff
@@ -0,0 +1,5974 @@
...
+(*           x `|` y == the join of x and y.                                  *)
+(*                 0 == the bottom element.                                   *)
+(*                 1 == the top element.                                      *)
+(*           x `\` y == the (sectional) complement of y in [0, x].            *)
+(*              ~` x == the complement of x in [0, 1].                        *)
+(*   \meet_<range> e == iterated meet of a lattice with a top.                *)
+(*   \join_<range> e == iterated join of a lattice with a bottom.             *)
+(*                                                                            *)
+(* There are three distinct uses of the symbols                               *)
+(* <, <=, >, >=, _ <= _ ?= iff _, >=<, and ><:                                *)
```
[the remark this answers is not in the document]
CohenCyril: fixed

mathcomp/ssreflect/order.v
```diff
@@ -0,0 +1,5974 @@
...
+(* 2. using this symbol to tag canoincal porderType structures using          *)
+(*    `Canonical my_porderType := POrderType my_display my_type my_mixin`,    *)
+(* 3. declaring notations for the main operations of this library, by         *)
+(*    setting the first argument of the definition to the display, e.g.       *)
+(*    `Notation my_syndef_le x y := @le my_display _ x y` or                  *)
+(*    `Notation "x <<< y" := @lt my_display _ x y (at level ...)`             *)
+(*    Non overloaded notations will default to the default display.           *)
+(*                                                                            *)
+(* One may use displays either for convenience or to desambiguate between     *)
+(* different structures defined on copies of one. Two examples of this are    *)
```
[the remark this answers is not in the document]
CohenCyril: I tried to address this.

mathcomp/ssreflect/order.v
```diff
@@ -0,0 +1,5974 @@
...
+(* given in this library:                                                     *)
+(* - the converse ordered structures. If T is an ordered type (and in         *)
+(*   particular porderType d), then T^c is a copy of T with the same ordered  *)
```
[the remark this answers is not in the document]
CohenCyril: The suggestion was not the right explanation. I tried to rephrase.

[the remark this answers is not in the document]
CohenCyril: I tried to explain better.

mathcomp/ssreflect/order.v
```diff
@@ -0,0 +1,5974 @@
...
+(*            orderType == the type of totally ordered types                  *)
+(*        finPOrderType == the type of partially ordered finite types         *)
+(*  finDistrLatticeType == the type of nonempty finite distributive lattices  *)
+(* finCDistrLatticeType == the type of nonempty finite complemented           *)
+(*                         distributive lattices                              *)
+(*         finOrderType == the type of nonempty totally ordered finite types  *)
+(*                                                                            *)
+(* Each of these structures take a first argument named display, of type      *)
+(* unit. Instantiating it with tt or an unknown key will lead to a default    *)
+(* display for notations, i.e. over these structures, we have:                *)
```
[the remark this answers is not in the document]
CohenCyril: fixed

CohenCyril:
@amahboubi are you happy with the rephrasings?

CHANGELOG_UNRELEASED.md
```diff
@@ -9,10 +9,112 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
...
+  + Generalized the following lemmas as properties of `normedDomainType`:
+    `normr0`, `normr0P`, `normr_eq0`, `distrC`, `normr_id`, `normr_ge0`,
+    `normr_le0`, `normr_lt0`, `normr_gt0`, `normrE`, `normr_real`,
+    `ler_norm_sum`, `ler_norm_sub`, `ler_dist_add`, `ler_sub_norm_add`,
+    `ler_sub_dist`, `ler_dist_dist`, `ler_dist_norm_add`, `ler_nnorml`,
+    `ltr_nnorml`, `lter_nnormr`.
+  + The compatibility layer for the version 1.9 is provided as the
+    `ssrnum.mc_1_9` module. One may compile proofs compatible with the version
+    1.9 in newer versions by using the `mc_1_9.Num` module instead of the `Num`
+    module. However, instances of the number structures may require changes.
```
pi8027: `1.9` and `mc_1_9.Num` should be `1.10` and `mc_1_10.Num`. I fix them now.

mathcomp/algebra/ssrnum.v
```diff
@@ -238,36 +255,167 @@ End Exports.
...
+  Pack phR (@Class T b0 m0).
+
+Definition eqType := @Equality.Pack cT xclass.
+Definition choiceType := @Choice.Pack cT xclass.
+Definition zmodType := @GRing.Zmodule.Pack cT xclass.
+
+End ClassDef.
+
+(* TODO: Ideally,`numDomain_normedZmodType` should be located in              *)
+(* `NumDomain_joins`. Currently, it's located here to make `hierarchy.ml` can *)
```
amahboubi: `hierarchy.ml` really?

CohenCyril: Yes cf etc/utils/hierarchy.ml (do you wish there were more explainations here?)

pi8027: This is because of `hierarchy.ml` constructs the inheritance graph from the output of `Print Canonical Projections` and filters out all the join of A and B where A and B do not inherit each other. If A inherits B, we usually declare the join of A and B in the `Exports` module of A, so `hierarchy.ml` filters out all the join of A and B that is not declared in the `Exports` module of A or B. However, this does not apply to `numDomainType` and `normedZmodType` if we get rid of this workaround.

Another possible solution is recognizing constants as inheritances that are declared as both canonical projections and coercions, but this does not apply to `finGroupType`.

CohenCyril:
@amahboubi I straightened the history. You can merge using "Create a merge commit" after the CI passes.

PR #266, opened 2018-12-20
Setup Docker-based configuration for GitLab CI

erikmd:
Closes #243 
Closes #256  
Related: #159, #245

Dear MathComp developers,

I've just finished to polish the configuration mentioned in https://github.com/math-comp/math-comp/issues/243#issuecomment-447613466.

Beyond what was already mentioned in this comment, this PR provides several other features:

* **opam-based build:** unlike the current Travis CI config that relies on make, the main `Dockerfile` in this PR relies on opam path pinning so (1) the opam specifications are tested, and (2) the mathcomp-dev Docker images obtained in the end allow installing other opam packages based on `coq-mathcomp-character:dev`. To implement this approach, it proved necessary to move the opam files in the parent folder, as opam 2.0 fails when a path-pinned package refers to files in the parent folders (e.g. `../Makefile.common`)

    * lightweight images: the main `Dockefile` in this PR only builds mathcomp with one switch (4.07.0+flambda if applicable), and relies on the so-called *multi-stage build* feature of Docker to significantly reduce the size of the images for Coq ≥ 8.7, given that many such images will be pushed to Docker Hub and/or the GitLab registry. Indeed without this feature, a Docker image that "inherits" from another image and just removes many files won't be smaller.  
    (Thus, the uncompressed size of coqorg/coq:dev (containing two switches) is around 2.86GB on my machine, while the size of a mathcomp-dev image built this way from coqorg/coq:dev and coqorg/base:bare is around 1.57GB.)

    * all branch builds of MathComp will pushed to the GitLab registry with a naming convention like this     (where 1 is the value of CI_PIPELINE_IID = unique id of the current pipeline scoped to the project):
        * `registry.gitlab.com/math-comp/math-comp:1_master_coq-8.6`
        * `registry.gitlab.com/math-comp/math-comp:1_master_coq-8.7`
        * `registry.gitlab.com/math-comp/math-comp:1_master_coq-8.8`
        * `registry.gitlab.com/math-comp/math-comp:1_master_coq-dev`

    * additionally, the master branch builds will be pushed to Docker Hub with the following naming convention:
        * `mathcomp/mathcomp-dev:coq-8.6`
        * `mathcomp/mathcomp-dev:coq-8.7`
        * `mathcomp/mathcomp-dev:coq-8.8`
        * `mathcomp/mathcomp-dev:coq-dev`

* **make-based build:** an extra job also builds MathComp using make with `coqorg/coq:latest` (latest stable version), essentially to test the "Makefile build specification".

[...]

PR #272, opened 2019-01-28
Add more libraries to CI & Update local opam doc

### LINKED ISSUE #251: Unable to install mathcomp as local opam packages
When trying to follow these instructions: https://github.com/math-comp/math-comp/blob/master/INSTALL.md#compilation-and-installation-of-a-custom-version-using-opam, I get the error given below.
As far as I understand the error message, opam copies the files (only) from the `ssreflect` directory into a sandbox directory and invokes `make`, but `Makefile` in `ssreflect` refers to `../Makefile.common`, which has not been copied and hence compilation fails.

```
$ opam pin -n add ssreflect
[coq-mathcomp-ssreflect.dev] synchronised from file:///Users/anton/prj/coq/mathcomp/mathcomp/ssreflect
coq-mathcomp-ssreflect is now pinned to file:///Users/anton/prj/coq/mathcomp/mathcomp/ssreflect (version dev)
$ opam install coq-mathcomp-ssreflect

<><> Synchronising pinned packages ><><><><><><><><><><><><><><><><><><><><>  🐫
[coq-mathcomp-ssreflect.dev] no changes from file:///Users/anton/prj/coq/mathcomp/mathcomp/ssreflect

The following actions will be performed:
  ∗ install coq-mathcomp-ssreflect dev*

<><> Processing actions <><><><><><><><><><><><><><><><><><><><><><><><><><>  🐫
[ERROR] The compilation of coq-mathcomp-ssreflect failed at "/Users/anton/.opam/opam-init/hooks/sandbox.sh build make -j 4".

#=== ERROR while compiling coq-mathcomp-ssreflect.dev =========================#
# context     2.0.1 | macos/x86_64 | ocaml-base-compiler.4.07.1 | pinned(file:///Users/anton/prj/coq/mathcomp/mathcomp/ssreflect)
# path        ~/.opam/coq.dev/.opam-switch/build/coq-mathcomp-ssreflect.dev
# command     ~/.opam/opam-init/hooks/sandbox.sh build make -j 4
# exit-code   2
# env-file    ~/.opam/log/coq-mathcomp-ssreflect-11185-5f1023.env
# output-file ~/.opam/log/coq-mathcomp-ssreflect-11185-5f1023.out
### output ###
# Makefile:8: ../Makefile.common: No such file or directory
# make: *** No rule to make target `../Makefile.common'.  Stop.



<><> Error report <><><><><><><><><><><><><><><><><><><><><><><><><><><><><>  🐫
┌─ The following actions failed
│ λ build coq-mathcomp-ssreflect dev
└─
╶─ No changes have been performed
```

erikmd:
Dear math-comp developers,

This PR is a follow-up of #266 and of @anton-trunov's suggestions in #245 and #256.

I basically selected among the 5 proposed libraries, the repos that had a `.travis.yml` configuration and the `.gitlab-ci.yml` is based on the corresponding commands; so the automated tests would include for the moment:

* https://github.com/math-comp/odd-order with coq-8.7
* https://github.com/math-comp/fourcolor with coq-{8.6, 8.7, 8.8, dev}
* https://github.com/coq-community/lemma-overloading with coq-{8.8, 8.9, dev}

while https://github.com/imdea-software/fcsl-pcm and https://github.com/math-comp/analysis don't have a Travis config in their upstream repo for the moment IINM.

Feel free to comment in this PR or in #245 if you'd want to add a few extra libraries to the CI, or if the suggested versions of Coq taken into account above should be changed.

Regarding the dependency on coq-8.9, since it is not yet merged in opam, it automatically relies on the `coqorg/coq:8.9-beta1` image for the moment.
Also, I needed to update the `coq-mathcomp-ssreflect.opam` file so the pinned package `coq-mathcomp-ssreflect.dev` could compile with Coq 8.9.
If this works, will you agree to update https://github.com/coq/opam-coq-archive/blob/master/extra-dev/packages/coq-mathcomp-ssreflect/coq-mathcomp-ssreflect.dev/opam likewise?

Finally, this PR also closes #251 (documentation issue)

Kind regards, Erik

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -4,7 +4,7 @@
 #     (Dockerfile containing: "opam switch set $compiler && eval $(opam env)")
 #   - master (protected branch) => push on GitLab registry and Docker Hub
 #   - other branches (not tags) => push on GitLab registry
-#   - Todo: GitHub PRs => push on GitLab
+#   - GitHub PRs => push on GitLab and report back thanks to @coqbot
 # - test stage (image: mathcomp-dev:$IID_$SLUG_coq-8.6)
 #   - script template foreach project (custom CONTRIB_URL, script)
 #   - jobs foreach project and Coq version (custom COQ_VERSION, CONTRIB_VERSION)
@@ -67,6 +67,9 @@ coq-8.7:
 coq-8.8:
   extends: .opam-build
 
+coq-8.9:
+  extends: .opam-build
+
 coq-dev:
   extends: .opam-build
 
@@ -117,6 +120,7 @@ make-coq-latest:
 # - Add 1 job per Coq version to test, that extends the previous hidden job,
 #   and sets vars COQ_VERSION, CONTRIB_VERSION (compatible Git branch/tag)
 
+# The Four Color Theorem
 .ci-fourcolor:
   extends: .ci
   variables:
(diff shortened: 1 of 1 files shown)


erikmd:
Hi @Zimmi48, it seems that **coqbot**'s report is working only for the pipeline itself, not for the individual jobs: IINM the report of the jobs should appear only if there is a job failing (?) and as soon as a job has failed; [which is the case](https://gitlab.com/math-comp/math-comp/-/jobs/152381492) − because I have forgotten the code below in the `.yml`:
```
coq-8.9:
  extends: .opam-build
```

This jobs-reporting feature was working in #270 however… do you have an idea of the source of that issue?
FYI here is a screenshot of the webhooks integration config:

2019-01-28_19-17-04_screenshot_webhooks

…I don't know if this a know bug, or if I should just remove the job webhook, and add it again?

Zimmi48:
@erikmd There are a few exceptions to the failures that are reported. Docker image not found is among them because it often corresponds to an outdated job whose image has been deleted. If you find this problematic, I could change that...

erikmd:
@Zimmi48 OK thanks for your reply! I was not aware that **coqbot** was filtering-out the job reports depending on the failures' type.

On the one hand, I find it very convenient that individual job successes don't lead to a report (in order to avoid cluttering the PR summary) but indeed on the other hand, the fact that only `script` failures are reported is not that intuitive IMHO: adding a report for all job failures could be handy to always have a nice *shortcut* to the corresponding log… (if it's not too much work for implementing this).

Anyway, if you think that changing this would have a negative impact on the workflow you are relying on in `coq` CI, forget about it!

Zimmi48:
>if it's not too much work for implementing this

It's not. In fact, it's more work to do the filtering. The issue, as shown by the message of the commit which introduced it, was rather to avoid creating noise (https://github.com/coq/bot/commit/e09800677c99cd15417a31d8fb19b8ab36faefe0). But there is no reason why this should impact you negatively. I've just made this condition specific to Coq (https://github.com/coq/bot/commit/488385c4172771ba5ac27296e33b2725649b844e).

erikmd:
> there is no reason why this should impact you negatively. I've just made this condition specific to Coq.

Great! Thank you @Zimmi48! and thanks @CohenCyril for the merge.

PR #275, opened 2019-01-30
we silence warnings that just pollute our logs so that we can see the relevant warnings

gares:
Namely:
-projection-no-head-constant
-redundant-canonical-projection
-notation-overridden

### DIFF
--- a/mathcomp/Make
+++ b/mathcomp/Make
@@ -90,3 +90,7 @@ ssreflect/tuple.v
 
 -I .
 -R . mathcomp
+
+-arg -w -arg -projection-no-head-constant
+-arg -w -arg -redundant-canonical-projection
+-arg -w -arg -notation-overridden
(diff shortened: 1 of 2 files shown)


gares:
I don't know what went wrong in CI and I don't have the rights to restart jobs. (can I get these)?

CohenCyril:
@gares what is you gitlab.com username?

erikmd:
@gares, indeed the `make-coq-latest` job had failed and it seems a spurious failure as the "main" jobs `coq-{8.6,8.7,8.8,8.9,dev}` were ok.
@CohenCyril will add you to the gitlab, and in the meantime I have re-triggered the pipeline from the page https://gitlab.com/math-comp/math-comp/pipelines (hopefully this will not only retry the failing job but also the second stage of the pipeline)

PR #283, opened 2019-02-11
Fix all the proof scripts by using the compatibility module (part of #270)

pi8027:


pi8027:
@CohenCyril This is ready to merge. Maybe I made algC.v slower than the original one.

PR #286, opened 2019-02-14
Export the compatibility module to check compilation (part of #270)

pi8027:


pi8027:
I think that CI will fail but wait for the result...

PR #288, opened 2019-02-19
Some progress on #270

pi8027:
- Renaming of `leIx*` and `lexU*` lemmas
- Factories (`LePOrderMixin`, `LtPOrderMixin`, `MeetJoinMixin`, `LeOrderMixin`, `LtOrderMixin` in order.v, and `RealLeMixin`, `RealLtMixin` in ssrnum.v)
- Adding `total_display` and `min`/`max` notations (`minr`/`maxr` for `ring_display`)
- Reorganize the structures of finite ordered sets (Some structures were useless and have been removed because nonempty finite lattices have top and bottom)


pi8027:
I found some fields of the above mixins are not necessary.

pi8027:
I have removed the unnecessary fields and added `min`/`max` notations.

PR #290, opened 2019-02-22
A tool to draw the hierarchy diagram

pi8027:
This is a small utility to draw the hierarchy diagram of the mathematical structures. This utility uses the unambiguous direct coercion paths between `class_of` types. Indirect inheritances are eliminated thanks to the property of coercion paths. However, some edges are missing in the output graph because the corresponding coercions between `class_of` types are not defined or the `Equality` module has no `class_of` record (it is defined as a notation).

If I use the information of coercion paths between `<module>.type` types instead of that between `<module>.class_of` types, more complicated method would be needed to eliminate edges of indirect inheritances.

Usage (graphviz is required):
```
$ ./etc/utils/hierarchy-diagram | dot -Tpdf -o hierarchy.pdf
```

Output: [hierarchy.pdf](https://github.com/math-comp/math-comp/files/2894433/hierarchy.pdf)

### DIFF
--- a/etc/utils/hierarchy-diagram
+++ b/etc/utils/hierarchy-diagram
@@ -1,55 +1,135 @@
 #!/bin/bash
 
+usage(){
+cat <<EOT >&2
+Usage : hierarchy-diagram [OPTIONS]
+
+Description:
+    hierarchy-diagram is a small utility to draw hierarchy diagrams of
+    mathematical structures.  This utility uses the coercion paths and the
+    canonical projections between <module>.type types (typically in the MathComp
+    library) to draw the diagram.  Indirect edges which can be composed of other
+    edges by transitivity are eliminated automatically for each kind of edges.
+    A diagram appears on the standard output in the DOT format which can be
+    converted to several image formats by Graphviz.
+
+Options:
+    -h, -help:
+        Output a usage message and exit.
+
+    -canonicals (off|on|color):
+        Configure output of edges of canonical projections.  The default value
+        is "on".
+
+    -coercions (off|on|color):
+        Configure output of edges of coercions.  The default value is "off".
+        The value given by this option must be different from that by
+        -canonical soption.
+
+    -R dir coqdir:
+        This option is passed to coqtop: "recursively map physical dir to
+        logical coqdir".
+
+    -lib library:
+        Specify a Coq library used to draw a diagram.  This option can appear
+        repetitively.  If not specified, all.all will be used.
+EOT
+}
+
 raw_coercions=$(tempfile -s .out | sed s/\.out$//)
 raw_canonicals=$(tempfile -s .out | sed s/\.out$//)
 parsed_coercions=$(tempfile)
 parsed_canonicals=$(tempfile)
+opt_canonicals=on
+opt_coercions=off
+opt_libs=()
+opt_coqtopargs=()
 
-coqtop -w none -R mathcomp mathcomp >/dev/null 2>&1 <<EOT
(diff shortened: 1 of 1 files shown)


pi8027:
If I use the information of coercion paths between `<module>.type` types instead of that between `class_of` types, more complicated method would be needed to eliminate edges of indirect inheritances.

pi8027:
I have reimplemented this by using coercion paths between `<module>.type` types. This one works fine.

Output: [hierarchy.pdf](https://github.com/math-comp/math-comp/files/2894903/hierarchy.pdf)

pi8027:
I found that some inheritances are missing (e.g., `FinGroup` -> `Finite`, `FinRing.Zmodule` -> `CountRing.Zmodule`), or... did I misunderstand something?

pi8027:
The missing edge from `FinGroup` to `Finite` was because of my mistake. I found a way to generate a similar diagram from canonical projections. So I will try to improve this as a checking tool detecting missing coercions and canonical projections. @CohenCyril

pi8027:
Red and blue edges are coercions and canonical projections respectively. Join canonical projections and indirect inheritances are eliminated. Output: [hierarchy.pdf](https://github.com/math-comp/math-comp/files/2901208/hierarchy.pdf)

amahboubi:
This is a great tool, thanks a lot! I guess a short documentation would be useful (e.g. your initial comment). But I do not know what is the appropriate place for this. May be a README.dev ? Or a wiki pages for the tools in the util directory? This is more a question for us I guess, @gares ? @CohenCyril ?

gares:
I agree with @amahboubi it needs some doc.
IMHO you could put a `usage { cat <<EOT .... EOT }` function at the beginning of the file, and if one passes `-h|--help` call this function and exit. The body of the function could contain the PR header, roughly. In this way the doc is on top of the file, if you open it with an editor, and also printed if you pass `-h`.

pi8027:
I have resumed this work today. Are there any desirable command line options to control output, e.g., direction, color, disabling coercions edge?

amahboubi:
I would say that control of the list of imported libraries would be an important feature. How would you be able to filter coercion edges? by full name? name pattern? type?

pi8027:
@amahboubi I have no good idea on how to filter out some structures and edges. But an option to specify libraries instead of `all.all` can be easily implemented.

amahboubi:
That would be great! It would also make the tool of a more general interest than just Mathcomp.

pi8027:
I finished to put doc and some options to control output. I changed the direction of edges to make them the same as Fig. 3 of "Packaging Mathematical Structures". Examples:

<pre>
$ ./etc/utils/hierarchy-diagram -canonicals blue -coercions off -R mathcomp mathcomp -lib ssreflect.all_ssreflect -lib fingroup.all_fingroup -lib algebra.all_algebra \
  | dot -Tpdf -o <a href="https://github.com/math-comp/math-comp/files/2942171/algebra-hierarchy.pdf">algebra-hierarchy.pdf</a>
$ ./etc/utils/hierarchy-diagram -canonicals blue -coercions red -R mathcomp mathcomp \
  | dot -Tpdf -o <a href="https://github.com/math-comp/math-comp/files/2942172/all-hierarchy.pdf">all-hierarchy.pdf</a>
</pre>

CohenCyril:
@gares and @amahboubi happy with the current state?

amahboubi:
I will not have time to review this thoroughly before next week. But I think that merging as is will not hurt so I am fine with it.

PR #291, opened 2019-02-27
Fix inheritances from countalg to finalg

pi8027:
The inheritances from countalg to finalg are broken. This has happened in #209 and found in #290.


CohenCyril:
This is blocking for the next release, @CohenCyril  when do you think you will have time to look at it?

mathcomp/algebra/finalg.v
```diff
@@ -94,18 +96,24 @@ Definition choiceType := @Choice.Pack cT xclass.
...
+Definition countZmod_baseFinGroupType :=
+  base_group countZmodType zmodType finType.
+Definition countZmod_finGroupType :=
+  fin_group countZmod_baseFinGroupType zmodType.
 
 End ClassDef.
 
 Module Exports.
 Coercion base : class_of >-> GRing.Zmodule.class_of.
+Coercion base2 : class_of >-> CountRing.Zmodule.class_of.
```
pi8027: We wouldn't need to export `base` as coercion thanks to this `base2`. Ideally, the `base` class of `finZmodType` is `countZmodType`, but actually it is `zmodType` because `Finite.mixin_of` contains `Countable.mixin_of`. I will keep this as it is and wait for some input.

CohenCyril: @pi8027 we should keep it like this for now, and open an issue, about putting the `Countable.class_of` in the `base` of `Fin.class_of` and check if math-comp still compiles, despite past timeouts...

CohenCyril:
@pi8027 I will review this 1st of April.

pi8027:
@CohenCyril Fortunately, the first 2 cases seems to be fixed in #270.

pi8027:
@CohenCyril I have fixed all the other inheritances you indicated. I prefer to merge this without fixing ssrnum, rebase #270 on top of it, and see which inheritances are broken.

CohenCyril:
Dear @pi8027, PR #270 will probably not make it to 1.8.0 but rather 1.9.0... so I would rather see the two joins in ssrnum fixed in this PR (#291) so that we can merge and release 1.8.0 asap.

pi8027:
@CohenCyril Fixed. I made a commit separately to make it easy to revert 7ab5c99 and rebase #270.

CohenCyril:
> @CohenCyril Fixed. I made a commit separately to make it easy to revert [7ab5c99](https://github.com/math-comp/math-comp/commit/7ab5c99ab4c2ecfd55702a4279392f067652e357) and rebase #270 later.

very good idea!

PR #294, opened 2019-03-04
Dependent positive finfun

ggonthier:
Generalise the finite function library to support dependently typed functions with a structurally positive representation and a structurally decreasing coercion to `Funclass`.
   Removes the `subType` instance for `finfun_of`, which is no longer used in inheriting combinatorial structures of the return type. The function graph `fgraph` projector is however retained for plain functions, and supplemented with a total function graph projector `tfgraph` in the general case.
   This PR runs into issue coq/coq#9663 (partial solution: https://github.com/coq/coq/pull/9690) which prevents type inference from matching generic dependent and non-dependent function types, and thereby includes some workarounds for this shortcoming. There is a companion PR math-comp/odd-order#11 that fixes one residual instance of this issue in the Odd Order proof.
  This PR also fixes a performance regression in `integral_char.v` that appears to have gone undetected - see the commit comment for a more detailed analysis.

### DIFF
--- a/mathcomp/ssreflect/finfun.v
+++ b/mathcomp/ssreflect/finfun.v
@@ -239,34 +239,35 @@ Notation family F := (family_mem (fmem F)).
 Section InheritedStructures.
 
 Variable aT : finType.
+Notation dffun_aT rT rS := {dffun forall x : aT, rT x : rS}.
 
-Definition finfun_eqMixin (rT : _ -> eqType) :=
-  @PcanEqMixin {dffun finPi aT rT} _ _ _ tfgraphK.
+Local Remark eqMixin rT : Equality.mixin_of (dffun_aT rT eqType).
+Proof. exact: PcanEqMixin tfgraphK. Qed.
 Canonical finfun_eqType (rT : eqType) :=
-  EqType {ffun aT -> rT} (finfun_eqMixin _).
-Canonical dfinfun_eqType (rT : _ -> eqType) :=
-  EqType {dffun finPi aT rT} (finfun_eqMixin _).
+  EqType {ffun aT -> rT} (eqMixin (fun=> rT)).
+Canonical dfinfun_eqType rT :=
+  EqType (dffun_aT rT eqType) (eqMixin rT).
 
-Definition finfun_choiceMixin (rT : _ -> choiceType) :=
-  @PcanChoiceMixin _ {dffun finPi aT rT} _ _ tfgraphK.
+Local Remark choiceMixin rT : Choice.mixin_of (dffun_aT rT choiceType).
+Proof. exact: PcanChoiceMixin tfgraphK. Qed.
 Canonical finfun_choiceType (rT : choiceType) :=
-  ChoiceType {ffun aT -> rT} (finfun_choiceMixin _).
-Canonical dfinfun_choiceType (rT : _ -> choiceType) :=
-  ChoiceType {dffun finPi aT _} (finfun_choiceMixin rT).
+  ChoiceType {ffun aT -> rT} (choiceMixin (fun=> rT)).
+Canonical dfinfun_choiceType rT :=
+  ChoiceType (dffun_aT rT choiceType) (choiceMixin rT).
 
-Definition finfun_countMixin (rT : _ -> countType) :=
-  @PcanCountMixin _ {dffun finPi aT rT} _ _ tfgraphK.
+Local Remark countMixin rT : Countable.mixin_of (dffun_aT rT countType).
+Proof. exact: PcanCountMixin tfgraphK. Qed.
 Canonical finfun_countType (rT : countType) :=
-  CountType {ffun aT -> rT} (finfun_countMixin _).
-Canonical dfinfun_countType (rT : _ -> countType) :=
-  CountType {dffun finPi aT rT} (finfun_countMixin _).
+  CountType {ffun aT -> rT} (countMixin (fun => rT)).
+Canonical dfinfun_countType rT :=
+  CountType (dffun_aT rT countType) (countMixin rT).
 
-Definition finfun_finMixin (rT : _ -> finType) :=
-  @PcanFinMixin (dfinfun_countType rT) _ _ _ tfgraphK.
+Local Definition finMixin rT :=
+  PcanFinMixin (tfgraphK : @pcancel _ (dffun_aT rT finType) _ _).
 Canonical finfun_finType (rT : finType) :=
-  FinType {ffun aT -> rT} (finfun_finMixin _).
-Canonical dfinfun_finType (rT : _ -> finType) :=
-  FinType {dffun finPi aT rT} (finfun_finMixin _).
+  FinType {ffun aT -> rT} (finMixin (fun=> rT)).
+Canonical dfinfun_finType rT :=
+  FinType (dffun_aT rT finType) (finMixin rT).
 
 End InheritedStructures.
 
--- a/mathcomp/ssreflect/finfun.v
+++ b/mathcomp/ssreflect/finfun.v
@@ -13,7 +13,8 @@ Require Import ssrfun ssrbool eqtype ssrnat seq choice fintype tuple.
 (* {ffun aT -> rT} as Leibnitz equality and extensional equalities coincide.  *)
 (*     (T ^ n)%type is notation for {ffun 'I_n -> T}, which is isomorphic     *)
 (*   to n.-tuple T, but is structurally positive and thus can be used to      *)
-(*   define inductive types, e.g., Inductive tree := node n of tree ^ n.      *)
+(*   define inductive types, e.g., Inductive tree := node n of tree ^ n (see  *)
+(*   mid-file for an expanded example).                                       *)
 (* --> More generally, {ffun fT} is always structurally positive.             *)
 (*  For f : {ffun fT} with fT := forall x : aT -> rT x, we define             *)
 (*              f x == the image of x under f (f coerces to a CiC function)   *)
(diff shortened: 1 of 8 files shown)


ggonthier:
There may be a problem with the way the CI pipeline is set up, since math-comp/odd-order#11, which fixes the Miller-pattern issue in odd-order, does not seem to be taken into account by the CI pipeline, even though it was merged five hours before the pipeline was restarted.

ejgallego:
With the current setup things are a hard to debug, I'd suggest we improve two things in the pipeline:
- output configuration variables `printenv -0 | sort -z | tr '\0' '\n'`
- output the commit hash of the contrib after checkout.

erikmd:
@ejgallego 

> output the commit hash of the contrib after checkout.

Indeed! @ggonthier could you add a commit in this PR with the following change?

Adding the line `- git rev-parse --verify HEAD` after that line:
https://github.com/math-comp/math-comp/blob/e5ff94165ab722bfb77d4437a58d49aacc81683d/.gitlab-ci.yml#L112

However I'd vote against the other suggestion by Emilio:

> output configuration variables `printenv -0 | sort -z | tr '\0' '\n'`

because there is some sensitive information (e.g. private tokens) among the environment variables of the build.

However, we could of course add something like this:

```
- echo COQ_VERSION=${COQ_VERSION}
- echo CONTRIB_URL=${CONTRIB_URL}
- echo CONTRIB_VERSION=${CONTRIB_VERSION}  
```

ejgallego:
> there is some sensitive information (e.g. private tokens) among the environment variables of the build.

Oh indeed, we had this problem in the Coq setup, however note that you are somehow vulnerable if regular users can trigger that.

What we do now in Coq is to have specific deploy jobs with a restricted `environment`, so deploy will only run in under very controlled conditions. I do indeed suggest that jobs which run "user code" should never ever run under an environment with protected variables.

erikmd:
Hi, I was not enough specific in my previous comment, yes the math-comp CI setup relies on protected variables, so if some user specifically opens a PR with `printenv`, no private info will be disclosed (unless the PR is merged in master of course)

Le 5 mars 2019 12:42:14 GMT+01:00, "Emilio Jesús Gallego Arias" <notifications@github.com> a écrit :
>> there is some sensitive information (e.g. private tokens) among the
>environment variables of the build.
>
>Oh indeed, we had this problem in the Coq setup, however note that you
>are somehow vulnerable if regular users can trigger that.
>
>What we do now in Coq is to have specific deploy jobs with a restricted
>`environment`, so deploy will only run in under very controlled
>conditions. I do indeed suggest that jobs which run "user code" should
>never ever run under an environment with protected variables.
>
>-- 
>You are receiving this because you commented.
>Reply to this email directly or view it on GitHub:
>https://github.com/math-comp/math-comp/pull/294#issuecomment-469649603

-- 
Envoyé de mon appareil Android avec K-9 Mail. Veuillez excuser ma brièveté.

ejgallego:
Hi @erikmd ,

> unless the PR is merged in master of course

indeed I was aware of this part of the setup, but even so, if jobs inheriting from `.ci` do have private variables this is still a big security risk as they are running external code.

For example I could take over lemmas overloading and dump the vars from there.

So I suggest indeed you do restrict the private variables using `environment` and then adding `printenv` to `.ci` jobs should not be a problem. Jobs that run in the protected env should never ever run any user code [only push artifacts].

erikmd:
Makes sense indeed, thanks @ejgallego! The math-comp CI jobs that clone external repos need not deploying anything...

Le 5 mars 2019 13:01:28 GMT+01:00, "Emilio Jesús Gallego Arias" <notifications@github.com> a écrit :
>Hi @erikmd ,
>
>> unless the PR is merged in master of course
>
>indeed I was aware of this part of the setup, but even so, if jobs
>inheriting from `.ci` do have private variables this is still a big
>security risk as they are running external code.
>
>For example I could take over lemmas overloading and dump the vars from
>there.
>
>So I suggest indeed you do restrict the private variables using
>`environment` and then adding `printenv` to `.ci` jobs should not be a
>problem. Jobs that run in the protected env should never ever run any
>user code [only push artifacts].
>
>-- 
>You are receiving this because you were mentioned.
>Reply to this email directly or view it on GitHub:
>https://github.com/math-comp/math-comp/pull/294#issuecomment-469654667

-- 
Envoyé de mon appareil Android avec K-9 Mail. Veuillez excuser ma brièveté.

ejgallego:
Cool! Let us know if you need help with that, in Coq we had great problem with the secret variables but I think we got the thing right.

The main invariant is that jobs that declare `enviroment` are to be treated as trusted code, and thus need careful review. In jobs that don't set the restricted environment up, people can add whatever they want.

mathcomp/ssreflect/finfun.v
```diff
@@ -36,128 +64,242 @@ Unset Printing Implicit Defensive.
...
+  (at level 0, x ident, format "[ 'ffun'  x  =>  E ]") : fun_scope.
 
-Notation "[ 'ffun' : aT => F ]" := (finfun (fun _ : aT => F))
-  (at level 0, only parsing) : fun_scope.
+Notation "[ 'ffun' => E ]" := [ffun _ => E]
+  (at level 0, format "[ 'ffun' =>  E ]") : fun_scope.
 
-Notation "[ 'ffun' x => F ]" := [ffun x : _ => F]
-  (at level 0, x ident, format "[ 'ffun'  x  =>  F ]") : fun_scope.
+(* Test that the type and accessor are structurally positive and decreasing,
```
CohenCyril: This test should be moved to a test suite (which disapeared with the integration of ssr in coq, but we should restore it for this purpose, and a few other)

CohenCyril:
@ggonthier I plan to finish the review by March 29th.

mathcomp/ssreflect/finfun.v
ggonthier: I'd rather keep this in the file, because it's a working demonstration of the use case mentioned in the header - I ought to have written "example" rather than "test" in the comment.
  We could legitimately include this code in a test suite, however, as the MathComp library will probably not make use of this feature in the near future, and we ought to be alerted if some change in the Coq positivity or guard rules breaks it.

CohenCyril: Do you mean we should push it to Coq test-suite instead?

ggonthier: Not really, because you'd need to copy all the `finfun` code and documentation to that test, so Coq devs would understand it. Plus, they might be tempted to fix the test without fixing `finfun`, if there is some string reason for changing the guard rules. If it's in `mathcomp`, which is in the CI anyway, at least we'll be notified.

CohenCyril: I guess I did not understand the "however" in
> We could legitimately include this code in a test suite, however, [...]

pi8027:
Note: my modification of fintype and finfun have been successfully rebased on top of this PR without losing positivity and structurally decreasing property. So I can use finfuns as arrays whose type of elements depends on indices. https://github.com/pi8027/math-comp/tree/modified-fintype-experimental
> Georges : the proposed modification may not be compatible with the way the new version of finfun works now.

ggonthier:
@pi8027 : you are correct, there will be a way to combine this implementation with constant-time indexing of `finTypes`. I think that there are better ways of doing this that what you've experimented with, however, and also that we should put some thought in making the transport reduce, if we want this to be useful for computing directly with matrices. All in all, that change should be a different PR.

CohenCyril:
Dear @ggonthier could you please rebase this branch on top of master?

CohenCyril:
Dear @ggonthier there are unresolved merge remaining in the PR hence the failure...
Also, the best practice is to rebase, I did it with my own copy and resolved the conflict I can push if you want. Rebasing was made harder because successive commits of this PR rewrite the same piece of code... I believe this would be a justification to squash them, unless there is a specific reason why you would keep them separate?

ggonthier:
@CohenCyril I've corrected the merge - I'd somehow missed the spurious conflicts in `finfun` for the one in `ChangeLog`. I do plan to squash and merge, but since several of the commits either explore different options, or work around Coq issues, it seemed preferable to keep them distinct during the discussion, and for future reference when the Coq issues are settled and we move to eliminate the work arounds.

CohenCyril:
@ggonthier I finally read it completely. I believe the final result would be good to merge.

One should:
- [ ] open an issue on mathcomp to remind ourselves that once https://github.com/coq/coq/pull/9690 to reaches two released versions of Coq, we should remove the duplication between `dffun` and `ffun`.
- [ ] open an issue or a PR about generalizing `_ =1 _` to the dependent case.
- [ ] at least squash 162cb3f67c20d4587a2e57738ffb5b3dbb63565c with the previous commit.

But ideally:
- [ ] squash/fixup at least 00cadb6aad408447739cddb86dddaf1d98e7b1be, 47254118ac8a28eb1108bcf85cca9c1a1bfa9571 and 6e853635171990d48bed1f129423019860ac3102
- [ ] and rebase so that the history looks like (roughly, with more squashable things inside brackets) :
850862dc6475bd48524a294651400df4b5b7ecf3 (master) -- df346ace8c0b5eb61f44b09abcebc193d8a0d6ac -- [ a9edbe27928858540c375c972d709df094c65d69 -- 1296cbd3d39ae04ab1ade2383f2396febb934c13 -- e64978dea1b796df8c4d827bca0fae0279b63724 ] -- 80846bf9d528e378fc2e6d809273aae676ecf964 -- cac497b28f88e8f95a42ee60e3593defcc196bad.
(if so 162cb3f67c20d4587a2e57738ffb5b3dbb63565c should be dealt with during the rebase)

Please tell me if you think this goes too far...

CohenCyril:
@ggonthier I rebased and did the squashing I mentionned in the last part of the commit. Please tell me if you are ok with it: https://github.com/CohenCyril/math-comp/tree/dependent-positive-finfun (the git diff is empty except for some spurious spaces and one typo)
If you agree I will force-push it here and merge.

ggonthier:
@CohenCyril everything looks fine to me now; you're clear to push, and merge after CI passes.

ggonthier:
@ejgallego this is a documented incompatibility: due to the change of representation `fgraph` is no longer be a field accessor, nor a `subType` projector. The failing lemma at the end of `heap.v` can no longer be proved by brute force `destruct`, nor as an instance of `val_inj` (as would have been the proper way to prove it). It can now easily be proved as `can_inj fgraphK`, but if one a proof compatible with all versions, this should be used:
```
by move=> eq_r12; apply/ffunP=> x; rewrite -[x]enum_rankK -!tnth_fgraph eq_r12.
```
@anton-trunov would either be acceptable? Alternatively, we could reinstate the `fgraph` `subType` structure on non-dependent finfuns, so `val_inj` could be used throughout. This would have to be deprecated, though, as the `eqType` structure will not be compatible.

PR #295, opened 2019-03-05
Export addrKA and subrKA from GRing.Theory

pi8027:
These lemmas are added by #160 but cannot be accessed in the unqualified form because they are not included in `GRing.Theory`.

Declaring all the lemmas in `GRing.Theory` first and ~~exporting~~ including them in `GRing` would be a better way to do this. But I haven't tried yet.


pi8027:
> Declaring all the lemmas in GRing.Theory first and exporting them in GRing would be a better way to do this. But I haven't tried yet.

`GRing.*` lemmas should be used in the qualified form. So the `Export` command wouldn't work in this case. The `Include` command may work but a little bit scary.

pi8027:
@CohenCyril This one might require your review.

PR #301, opened 2019-03-21
missing lemma in seq.v

CohenCyril:


amahboubi:
Fine with me.

mathcomp/ssreflect/seq.v
```diff
@@ -1926,6 +1926,9 @@ Qed.
 Lemma mem_rem_uniq s : uniq s -> rem s =i [predD1 s & x].
 Proof. by move/rem_filter=> -> y; rewrite mem_filter. Qed.
 
+Lemma mem_remF s : uniq s -> x \in rem s = false.
```
CohenCyril: I am having afterthough. I think a better name would be `uniq_mem_remF`

mathcomp/ssreflect/seq.v
```diff
@@ -1926,6 +1926,9 @@ Qed.
...
+Proof. by move=> us; rewrite mem_rem_uniq // inE eqxx. Qed.
```
ggonthier: Perhaps `by move/mem_rem_uniq->; rewrite inE eqxx.` ?

mathcomp/ssreflect/seq.v
ggonthier: To be consistent with the lemma immediately above, it should be `mem_rem_uniqF`.

gares:
@amahboubi you happen to be the assignee, please merge at your convenience

CohenCyril:
@gares ok now?

PR #236, opened 2018-10-29
Revert "Adding allsigs, the dependent version of allpairs"

CohenCyril:
This reverts commit 044634dcc2f645e3afbdad6cb8dcc66f3eb4a87e.
See issue https://github.com/math-comp/odd-order/pull/5

PR #304, opened 2019-03-21
Refactoring allpairs to handle the dependent version as well

CohenCyril:
The return of `allsigs` cf #236, math-comp/odd-order#5 and #217: refactoring with `allpairs` this time.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -105,6 +105,14 @@
 	* Added lemmas `foldrE`, `foldlE`, `foldl_idx` and `sumnE`
 	  to turn "seq statements" into "bigop statements"
 
+	* Generalized the notation `[seq E | i <- s, j <- t]` to the case
+	  where `t` may depend on `i`. The notation is now primitive and
+	  expressed using `flatten` and `map` (see documentation of seq).
+	  `allpairs` now expands to this notation when fully applied.
+	  Added `allpairs_dep` and made it self-expanding as well.
+	  Generalized some lemmas in a backward compatible way.
+	  Some strictly more general lemmas now have suffix `_dep`.
+
 24/04/2018 - compatibility with Coq 8.8 and several small fixes - version 1.7
 
 	* Added compatibility with Coq 8.8 and lost compatibility with
--- a/mathcomp/ssreflect/bigop.v
+++ b/mathcomp/ssreflect/bigop.v
@@ -1193,13 +1193,19 @@ rewrite !(big_mkcond _ P) unlock.
 by elim: r1 => /= [|i r1 ->]; rewrite (mul1m, mulmA).
 Qed.
 
-Lemma big_allpairs I1 I2 (r1 : seq I1) (r2 : seq I2) F :
-  \big[*%M/1]_(i <- [seq (i1, i2) | i1 <- r1, i2 <- r2]) F i =
-    \big[*%M/1]_(i1 <- r1) \big[op/idx]_(i2 <- r2) F (i1, i2).
+Lemma big_allpairs_dep I1 (I2 : I1 -> Type) J (h : forall i1, I2 i1 -> J)
+    (r1 : seq I1) (r2 : forall i1, seq (I2 i1)) (F : J -> R) :
+  \big[*%M/1]_(i <- [seq h i1 i2 | i1 <- r1, i2 <- r2 i1]) F i =
+    \big[*%M/1]_(i1 <- r1) \big[*%M/1]_(i2 <- r2 i1) F (h i1 i2).
 Proof.
 elim: r1 => [|i1 r1 IHr1]; first by rewrite !big_nil.
 by rewrite big_cat IHr1 big_cons big_map.
 Qed.
+
+Lemma big_allpairs I1 I2 (r1 : seq I1) (r2 : seq I2) F :
+  \big[*%M/1]_(i <- [seq (i1, i2) | i1 <- r1, i2 <- r2]) F i =
+    \big[*%M/1]_(i1 <- r1) \big[op/idx]_(i2 <- r2) F (i1, i2).
+Proof. exact: big_allpairs_dep. Qed.
 
 Lemma big_pred1_eq (I : finType) (i : I) F :
   \big[*%M/1]_(j | j == i) F j = F i.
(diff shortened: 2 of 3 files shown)


CohenCyril:
@ggonthier I am advocate for leaving the duplications for `allpairs` for backward compatibility purposes.

ggonthier:
You have a point, but since the theory no longer concerns the actual `allpairs` constant, perhaps a better solution would be to repurpose it as the general tag for lemmas about the two-variable comprehension notation, using a '_dep' suffix to identify lemmas that cover the dependent case when there is a specific lemma for the non-dependent case, consistent with the rest of MathComp. That means `allsigs` would not be used, so you'll have to rename the PR...

CohenCyril:
@ggonthier happy with the result?

mathcomp/ssreflect/bigop.v
```diff
@@ -1159,6 +1159,15 @@ rewrite !(big_mkcond _ P) unlock.
...
+Lemma big_allpairs_dep  (I1 : Type)  (I2 : I1 -> Type)
+  (r1 : seq I1) (r2 : forall i1, seq (I2 i1)) (F : sigT I2 -> R) :
+  \big[op/idx]_(i <- [seq Tagged I2 i2 | i1 <- r1, i2 <- r2 i1]) F i =
+  \big[op/idx]_(i1 <- r1) \big[op/idx]_(i2 <- r2 i1) F (Tagged I2 i2).
+Proof.
+elim: r1 => [|i1 r1 IHr1]; rewrite !(big_nil, big_cons)//= big_cat {}IHr1.
+by case: (r2 i1) => [|i2 r21]; rewrite /= !(big_nil, big_cons)//= big_map.
+Qed.
+
 Lemma big_allpairs I1 I2 (r1 : seq I1) (r2 : seq I2) F :
```
ggonthier: Hadn't we agreed to generalise `sigT I2` to an arbitrary intermediate type, and `Tagged I2` to an arbitrary function?

mathcomp/ssreflect/seq.v
```diff
@@ -108,9 +108,15 @@ Require Import ssrfun ssrbool eqtype ssrnat.
...
 (*  ** Iterators: for s == [:: x_1, ..., x_n], t == [:: y_1, ..., y_m],       *)
 (*        map f s == the sequence [:: f x_1, ..., f x_n].                     *)
-(* allpairs f s t == the sequence of all the f x y, with x and y drawn from   *)
-(*                  s and t, respectively, in row-major order.                *)
+(* [seq E | x <- s, y <- t] := (flatten [seq [seq E | x <- t] | y <- s])      *)
+(*                == the sequence of all the f x y, with x and y drawn from   *)
+(*                   s and t, respectively, in row-major order,               *)
+(*                   and where t is possibly dependent in elements of s       *)
 (*               := [:: f x_1 y_1; ...; f x_1 y_m; f x_2 y_1; ...; f x_n y_m] *)
+(* allpairs_dep f s t := self expanding definition for                        *)
```
ggonthier: the `m` here now depends on the `x` index.

mathcomp/ssreflect/seq.v
```diff
@@ -2733,31 +2738,106 @@ rewrite /N /= count_cat -/(N x _) Nx0 ?mem_rem_uniq ?undup_uniq ?inE ?eqxx //.
...
+Proof. by move=> xs yt; apply/allpairsPdep; exists x, y. Qed.
+
+Lemma mem_allpairs_dep R (f : forall x, T x -> R) s1 t1 s2 t2 :
+    s1 =i s2 -> (forall x, x \in s1 -> t1 x =i t2 x) ->
+  [seq f x y | x <- s1, y <- t1 x] =i [seq f x y | x <- s2, y <- t2 x].
+Proof.
+move=> eq_s eq_t z; apply/allpairsPdep/allpairsPdep=> -[x [y [xs ys ->]]];
+by exists x, y; rewrite ?eq_s ?eq_t// -?eq_s -?eq_t// eq_s.
+Qed.
+
```
ggonthier: This should be `eq_allpairs_dep`.

CohenCyril: I hesitated, but the non dependent version is (and was) `mem_allpairs` and the LHS and RHS are indeed have head symbol `_ \in _`.

mathcomp/ssreflect/bigop.v
CohenCyril: Oh, yes, I forgot about this file, thanks for the reminder!

mathcomp/ssreflect/seq.v
ggonthier: Sorry, I later agreed with you and meant to erase that comment.

mathcomp/ssreflect/seq.v
```diff
@@ -2733,31 +2738,106 @@ rewrite /N /= count_cat -/(N x _) Nx0 ?mem_rem_uniq ?undup_uniq ?inE ?eqxx //.
...
+Lemma size_allpairs s t : size [seq f x y | x <- s, y <- t] = size s * size t.
 Proof. by elim: s => //= x s IHs; rewrite size_cat size_map IHs. Qed.
 
-Lemma allpairs_cat s1 s2 t :
-  allpairs (s1 ++ s2) t = allpairs s1 t ++ allpairs s2 t.
-Proof. by elim: s1 => //= x s1 ->; rewrite catA. Qed.
+End AllPairsNonDep.
 
-End AllPairs.
+Arguments allpairs {S T R} / f s.
```
ggonthier: The `/` should come after the `s` _and_ `t` arguments: we only want to auto-expand the fully applied definition.

mathcomp/ssreflect/seq.v
CohenCyril: actually, not only `m`, but the whole sequence...

CohenCyril: thus I rephrased once more not to give to much details...

mathcomp/ssreflect/seq.v
CohenCyril: I must have been very distracted when I did this...

ChangeLog
```diff
@@ -105,6 +105,11 @@
 	* Added lemmas `foldrE`, `foldlE`, `foldl_idx` and `sumnE`
 	  to turn "seq statements" into "bigop statements"
 
+	* Added allpairs_dep, the dependent version of allpairs,
+	  with the same notation `[seq E | i <- s, j <- t]`
```
ggonthier: As `all pairs_dep` is now a secondary definition, so this should be rephrased to emphasize that the notation is now central, and `allpairs` is a self-expanding definition.

CohenCyril: done

PR #305, opened 2019-03-21
compat sumn with bigop

CohenCyril:


CohenCyril:
I think `sumnE : sumn r = \sum_(i <- r) i` is missing.

ggonthier:
I'd say `sumnE` is the only addition that can really be justified; its proof is `rewrite unlock`. The other lemmas can be readily derived from it, perhaps with a specialisation of `big_map` to `F := id` to help with rewriting in the right-to-left direction. In the same vein we could provide a lemma to rewrite a `foldr` into a `bigop`.
  Unless there's a strong use case I'm missing, it hardly seems reasonable to specialise `sumnE` to every kind of list comprehension notation under the sun. On the subject of comprehensions, the current library is missing lemmas for the `image` seq comprehension, though `finset` does have some for `imset`, so perhaps we should provide those.

CohenCyril:
@ggonthier I updated the PR to reflect your comment, and I agree with you about not multiplying specializations. I also added elimination of `foldr`, and also of `foldl` for completeness.

CohenCyril:
@ggonthier happy with the current result?

Issue #299, opened 2019-03-19
Generation of permutations


MrSet:
I have been using the definition of permutational equivalence `perm_eq` from `seq.v`for a while and have a few suggestions for extensions of the library. I will like to hear opinions on what extensions you think will be relevant for the mathcomp library.
For this issue please only consider **if** things are relevant for mathcomp. Phrasing and styling of the lemmas/definitions can be postponed to a possible pull request.  

## Misc. extensions of `seq.v`
Standard lemma in the style of `perm_map`
```coq
Lemma perm_flatten {A : eqType} :
   forall (l l' : seq (seq A)),
     perm_eq l l' -> perm_eq (flatten l) (flatten l').
```
Lemmas which I have found useful for reasoning about `perm_eq`. Could be rephrased using `reflect`. 
```coq 
  Lemma count_mem_geq1 {A : eqType} (hd : A) :
    forall (l : seq A),
      count_mem hd l >= 1 ->
      exists (l' l'' : seq A), l = l' ++ hd :: l''.

  Lemma perm_eq_cons {A : eqType} (hd : A) tl l:
    perm_eq (hd :: tl) l ->
    exists l' l'', l = l' ++ hd :: l'' /\ perm_eq tl (l' ++ l'').
```

Substitution lemma
```coq
  Lemma perm_subst_help {T : eqType} : forall (a b c d : seq T)
  , perm_eq b d ->  perm_eq (a ++ b) c -> perm_eq (a ++ d) c.
```

## Reflection of `perm_eq` to Permutation from the standard library
I have found this useful whenever reasoning about an assumption on `perm_eq`.
```
  Lemma perm_eq_permP {A : eqType} :
    forall (l l' : seq A), reflect (Permutation l l') (perm_eq l l').
```
## Generation of permutations
The following two functions can create all permutations of a list. 

```coq
  Fixpoint interleave {A : Type} (x : A) (l : seq A) : seq (seq A) :=
    match l with
    | hd :: tl => (x :: l) :: [seq hd :: y | y  <- interleave x tl]
    | [:: ] => [:: [:: x]]
    end.

  Fixpoint permutations {A : Type} (l : list A) : list (list A) :=
    match l with
    | hd :: tl => flatten [seq interleave hd x | x <- permutations tl]
    | [::] => [:: [::]]
    end.
```
And quite a few lemmas of the correctness of these two functions, about sizes of the generated list of permutations, and that all permutations are generated and only permutations are generated. The most important ones are:

```coq
[...]

PR #314, opened 2019-03-27
Add allpairs_dep_map(l|r) and eq_in_allpairs

pi8027:
I'm using non-dependent variants of them in my WIP branch.

### DIFF
--- a/mathcomp/ssreflect/seq.v
+++ b/mathcomp/ssreflect/seq.v
@@ -2777,6 +2777,21 @@ End AllPairsDep.
 
 Arguments allpairs_dep {S T R} f s t /.
 
+Section AllPairsDepMap.
+
+Variables (S S' : Type) (T T' : S -> Type) (R : Type) (f : forall x, T x -> R).
+
+Lemma allpairs_dep_mapl (g : S' -> S) s (t : forall x : S, seq (T x)) :
+  [seq f y | x <- map g s, y <- t x] = [seq f y | x <- s, y <- t (g x)].
+Proof. by elim: s => //= x s ->. Qed.
+
+Lemma allpairs_dep_mapr (g : forall x, T' x -> T x) s
+      (t : forall x : S, seq (T' x)) :
+  [seq f y | x <- s, y <- map (g x) (t x)] = [seq f (g x y) | x <- s, y <- t x].
+Proof. by elim: s => //= x s ->; rewrite -map_comp. Qed.
+
+End AllPairsDepMap.
+
 Section AllPairsNonDep.
 
 Variables (S T R : Type) (f : S -> T -> R).
@@ -2795,6 +2810,15 @@ Section EqAllPairsDep.
 
 Variables (S : eqType) (T : S -> eqType).
 Implicit Types (R : eqType) (s : seq S) (t : forall x, seq (T x)).
+
+Lemma eq_in_allpairs R (f g : forall x, T x -> R) s t :
+  (forall x, x \in s -> {in t x, f x =1 g x}) ->
+  [seq f x y | x <- s, y <- t x] = [seq g x y | x <- s, y <- t x].
+Proof.
+elim: s => //= x s ihs e; rewrite ihs.
+by congr cat; apply/eq_in_map => y; apply: (e x); rewrite inE eqxx.
+by move=> x' x'in y'; apply: (e x' _ y'); rewrite inE x'in orbT.
+Qed.
 
 Lemma allpairsPdep R f s t (z : R) :
   reflect (exists x y, [/\ x \in s, y \in t x & z = f x y])


mathcomp/ssreflect/seq.v
```diff
@@ -2777,6 +2777,21 @@ End AllPairsDep.
 
 Arguments allpairs_dep {S T R} f s t /.
 
+Section AllPairsDepMap.
+
+Variables (S S' : Type) (T T' : S -> Type) (R : Type) (f : forall x, T x -> R).
+
+Lemma allpairs_dep_mapl (g : S' -> S) s (t : forall x : S, seq (T x)) :
```
CohenCyril: I think one can omit the `_dep_` here since high order unification will fall in Miller pattern unification while rewriting both ways... @ggonthier am I right?

mathcomp/ssreflect/seq.v
```diff
@@ -2777,6 +2777,21 @@ End AllPairsDep.
...
+  [seq f y | x <- map g s, y <- t x] = [seq f y | x <- s, y <- t (g x)].
+Proof. by elim: s => //= x s ->. Qed.
+
+Lemma allpairs_dep_mapr (g : forall x, T' x -> T x) s
```
CohenCyril: same here

mathcomp/ssreflect/seq.v
```diff
@@ -2796,6 +2811,15 @@ Section EqAllPairsDep.
 Variables (S : eqType) (T : S -> eqType).
 Implicit Types (R : eqType) (s : seq S) (t : forall x, seq (T x)).
 
+Lemma eq_in_allpairs R (f g : forall x, T x -> R) s t :
```
CohenCyril: the proof is a one liner but the statement can be generalized:
```coq
Lemma eq_in_allpairs R (f g : forall x, T x -> R) s t :
  (forall x, x \in s -> {in t x, f x =1 g x}) <->
  [seq f x y | x <- s, y <- t x] = [seq g x y | x <- s, y <- t x].
Proof.
split=> fg.
  by congr (flatten _); apply/eq_in_map=> ??; apply/eq_in_map=> ??; apply: fg.
move=> x xs; apply/eq_in_map; move: x xs; apply/eq_in_map.
apply: eq_from_flatten_shape => //.
by rewrite /shape -!map_comp/=; apply: eq_map => x/=; rewrite !size_map.
Qed.
```
and specialized to extensional equality everywhere
```coq
Lemma eq_allpairs R (f g : forall x, T x -> R) s t : (forall x, f x =1 g x) ->
  [seq f x y | x <- s, y <- t x] = [seq g x y | x <- s, y <- t x].
Proof. by move=> fg; apply/eq_in_allpairs => ????; apply: fg. Qed.
```

mathcomp/ssreflect/seq.v
ggonthier: Correct; also, this should go straight in the `AllPairsDep` section, and the proof should be `rewrite -map_comp`.

mathcomp/ssreflect/seq.v
ggonthier: Likewise, and the proof should be `rewrite -(eq_map (fun=> map_comp _ _))`
Also, the `allpairs_comp` lemma should be taken out of the section, proved with `rewrite map_flatten allpairs_mapl allpairs_mapr.`, replaced by its converse (exchanging LHS and RHS), and renamed `map_allpairs`: it cannot be used left-to-right, because the `funcomp` constant does not appear.

mathcomp/ssreflect/seq.v
ggonthier: That's not right, because `eq_allpairs` should not require `T : eqType`.
Also, there is a point to have a non-dependent version of `eq_in_allpairs`, because the function equality can be stated differently in this case as `{in s & t, f =2 g}`, so the lemma above should be called `eq_in_allpairs_dep`.

ggonthier: Also, I strongly object to your practice of juxtaposing `?` patterns as in `????` above:
(a) This is inviting trouble should either Coq or user libraries define a `??` symbol.
(b) It forgoes a basically free opportunity to better document the proof script.

mathcomp/ssreflect/seq.v
CohenCyril: > renamed `map_allpairs`:

because of this, I suggest this goes in 1.8.0 in order not to introduce backward incompatibilities with a future 1.8.1 version

pi8027:
@CohenCyril If you need to fix and merge this PR asap, I'd like to give you ownership of it. I may not able to work on this in this week.

CohenCyril:
> @CohenCyril If you need to fix and merge this PR asap, I'd like to give you ownership of it. I may not be able to work on this in this week.

ok taking ownership now

ggonthier:
Actually, I've been working on a bit of refactoring of `seq` while looking into implementing some of @MrSet's suggestions #299, which has uncovered a few loose threads in `seq`: missing implicits and documentation. I've also got @pi8027 's `allpairs` extension in, so perhaps I can do the hijacking (leaving out the `permutations` development, which requires a bit more work).

CohenCyril:
> Actually, I've been working on a bit of refactoring of `seq` while looking into implementing some of @MrSet's suggestions #299, which has uncovered a few loose threads in `seq`: missing implicits and documentation. I've also got @pi8027 's `allpairs` extension in, so perhaps I can do the hijacking (leaving out the `permutations` development, which requires a bit more work).

Do you think it will be ready soon, with @gares we planned to release this afternoon...

CohenCyril:
PS: I did not see your comment, so I did this: https://github.com/CohenCyril/math-comp/tree/allpairs-lemma

ggonthier:
I've been working on a separate branch, which I've just published (sew-permutations-allpairs). I'm just checking the MC compile, then you can either push it here or turn it into a PR that replaces this one.

ggonthier:
There are still improvements I want to explore with `permutations`, but that can become a later PR.

CohenCyril:
> I'm just checking the MC compile

It does not compile.
Could you please submit a PR rebased on master?

ggonthier:
I did notice, and I'm digging into it because I am merged with master, and things break in an unrelated file.

ggonthier:
Fine, @CohenCyril, it's rebased and compiles. Not sure what happened in `primitive_action`; could be an issue with `trivial` rather than `intro`.
   On a separate issue, would it be possible for `hierarchy_test` _NOT_ to dump tons of diagnostics on successful compilations?

CohenCyril:
> On a separate issue, would it be possible for `hierarchy_test` _NOT_ to dump tons of diagnostics on successful compilations?

yes!

CohenCyril:
Dear @ggonthier, the branch https://github.com/math-comp/math-comp/compare/seq-permutations-allpairs looks like it is **merged** with master and not **rebased** on it.
Additionnally it is failing because of the use of the new ssreflect clear behaviour, which was not yet available in 8.7 and 8.8 (we are dropping support for 8.6 in PR #321).

CohenCyril:
> > On a separate issue, would it be possible for `hierarchy_test` _NOT_ to dump tons of diagnostics on successful compilations?
> 
> yes!

I addressed it in #323

ggonthier:
I've rebased (turns out git had inserted a no-op back merge of the remote, which threw off any further rebasing) and turned a branch into a PR, so I can tell where the offending clear is...
I'll probably en up using the auto squash, since my git -i install on macOS is broken (sic).

CohenCyril:
> I'll probably en up using the auto squash, since my git -i install on macOS is broken (sic).

I can squash it for you. (I will also chase spurious spaces (some remained in your previous commits))

Also, I think you did not take into account my last remark:

> Additionally it is failing because of the use of the new ssreflect clear behavior, which was not yet available in 8.7 and 8.8 (we are dropping support for 8.6 in PR #321).

so the CI will fail again on 8.7 and 8.8...

ggonthier:
I did - I'm waiting for the CI to fail so I know what to fix.

CohenCyril:
> I did - I'm waiting for the CI to fail so I know what to fix.

I know what to fix:

```coq
-elim: s => //= x s IHs /andP[s'x {}/IHs-IHs]; rewrite factS -IHs mulnC.
+elim: s => //= x s IHs /andP[s'x /IHs-{IHs}IHs]; rewrite factS -IHs mulnC.
```
 
```coq
-  by move=> m n /subnK<-; elim: {n}(n - m) => // n IHn {}/IHn; apply: PsS.
+  by move=> m n /subnK<-; elim: {n}(n - m) => // n IHn /IHn; apply: PsS.
```

CohenCyril:
maybe you want to fix it in a slightly different way so that no *duplication of clear* warning are triggered on Coq 8.9 :laughing:

ggonthier:
Ok, fixed (blindly).

ggonthier:
Please do go ahead, then, by all means.

PR #316, opened 2019-03-30
[build] Preliminary support for building with Dune.

ejgallego:
This PR provides preliminary support for building math-comp with Dune
[support in Dune's master currently]

As of today we do consider math-comp as a monolithic library due to
Dune's Coq support being unable to compose libraries yet, this will be
fixed in the future.


CohenCyril:
@ejgallego and @gares should this go to 1.8.0 ?

ejgallego:
This is experimental so people interested in the Dune build can run it, I don't think it affects the regular workflow otherwise yet.

gares:
This is mostly fine to me (we could merge it) but for the latter opam file. I think know why it is there, but that opam package corresponds to no real opam package. Couldn't we use the real opam packages instead (and use dune's compositional build stuff?).

ejgallego:
> Couldn't we use the real opam packages instead (and use dune's compositional build stuff?).

Yes, in a few days, the compositional stuff is undergoing testing as we speak; so indeed I am fine to wait until I update this PR with the split package support.

CohenCyril:
Let's target 1.8.1 then! Thanks

CohenCyril:
Hi @ejgallego, could you please write some comments to document the build files, add tests to the CI, and maybe explain how it changes the development process (with coqide? proofgeneral?...)

ejgallego:
Noted, will do @CohenCyril .

gares:
ping. Also, I think that the `synopsis` field should be more precise than that.

ejgallego:
This is still waiting on a new Dune release for composition support.

ejgallego:
Ok, this is finally working, but needs an unreleased version of dune, hopefully we do a dune release soon.

Note the large amount of warnings that we have to disable in order to get a clean build. A missing feature for the Coq mode is to support build profiles [that would allow a common set of flags, but I'm unsure we want it here instead of per-lib. A warning cleanup would be nice tho.

> Hi @ejgallego, could you please write some comments to document the build files, add tests to the CI, and maybe explain how it changes the development process (with coqide? proofgeneral?...)

I will add a CI build as soon as we get a release version of dune, I'm unsure what to write about the documentation of the build files, as far as I know these are standard and covered properly in both the Coq and dune manual.

We will add support to CoqIDE / PG very soon, that will be mostly automatic.

affeldt-aist:
The release managers for mathcomp 1.11.0 think that, given the fact that it depends on an unreleased version of dune (as of March 7), this PR could be postponed to milestone 1.12.0. Agreed?

ejgallego:
Hi @affeldt-aist , sure, this is not urgent. Note that this PR just adds support for the dune build, I think there is still a long way before it could be considered to be the default.

CohenCyril:
@gares @ejgallego I do not understand the pros and cons. Do you think we could converge towards either a milestone or closure until we have a milestone?

ejgallego:
No particular opinion, I guess that's a matter of preference.

CohenCyril:
Then I guess the switch to dune should be discussed at the next mathcomp meeting, and then possibly scheduled for some release.

.gitlab-ci.yml
```diff
@@ -79,6 +79,18 @@ make-coq-latest:
     - tags
     - merge_requests
 
+.dune-build:
+  stage: build
+  image: coqorg/coq:dev
+  script:
+    - opam switch "$COMPILER_EDGE"
+    - opam pin -y add dune https://github.com/ejgallego/dune.git#coq+compose_local
```
gares, line 87: Do we really need an unreleased version of Dune?

ejgallego: It should work fine with recent versions, but the PR needs updating.

ejgallego:
Ok, I've updated the PR. @corwin-of-amber did some testing and found some bugs in Dune, but in principle things should work fine.

Pending stuff here is meta-data for the opam file generation, and reorg / cleanup of files if desired.

.gitlab-ci.yml
ejgallego: Updated.

mathcomp/algebra/dune
```diff
@@ -0,0 +1,19 @@
+(coq.theory
+ (name mathcomp.algebra)
+ (package coq-mathcomp-algebra)
+ (modules :standard)
+ (synopsis "Mathematical Components")
```
gares, line 5: The metadata is sketchy, the one in the opam files is much better. Example:
```
synopsis: "Mathematical Components Library on Algebra"
description: """
This library contains definitions and theorems about discrete
(i.e. with decidable equality) algebraic structures : ring, fields,
ordered fields, real fields,  modules, algebras, integers, rational
numbers, polynomials, matrices, vector spaces...
"""
```

gares (changes requested):

mathcomp/algebra/dune
gares: I guess this metadata is used in order to generate the opam packages, but I may be wrong.

ejgallego: Oh indeed I didn't go the effort to update the metadata until you folks discussed if this would be accepted. Will fix.

ejgallego: Ok, meta-data update, noted that I didn't enable the opam file generation yet as it would overwrite the opam files in the repos.

ejgallego: Actually the meta-data here is for the META file, which is not relevant here. If we have something similar to META files for Coq one day it would become relevant.

gares:
As per https://github.com/coq/ceps/pull/48#issuecomment-727585541 this PR is blocked by ocaml/dune#3210 IMO

.gitlab-ci.yml
```diff
@@ -79,6 +79,19 @@ make-coq-latest:
       - tags
       - merge_requests
 
+.dune-build:
+  stage: build
+  image: coqorg/coq:dev-ocaml-4.13.1-flambda
+  script:
+    - opam switch "$COMPILER_EDGE"
+    - opam install -y dune
+    - opam pin -y add dune git+https://github.com/Alizter/dune#coq-compose.git
```
gares, line 88: Is this branch of dune a requirement?

Alizter: To be clear, detection of libraries in user-contrib hasn't been implemented yet, only inter-project composition of coq dune projects. Adding deps on installed libs is the next step, but I am not certain if this will make 3.3.

proux01 (changes requested):
This shouldn't be merged until we get a release of dune supporting coq-native.
The alternative would be to add configure scripts to edit the dune files but that proved a nightmare to get reasonably right in multinomials.

PR #317, opened 2019-04-01
Make the finTypes to be equipped with indexing functions

pi8027:
depends on: #294

This PR redefine the `finType` structure by adding the following bijection to `Finite.mixin_of T`.
```
fin_encode : T -> 'I_#|T|, fin_decode : 'I_#|T| -> T.
```
I have used this variant of `finType` to obtain efficient programs manipulating finfuns by program extraction. Finfuns can be extracted to arrays and a finfun application of `f : {ffun T -> A}` can be done by `@fin_encode T` and `Array.get` (in OCaml). More details can be found in [my FLOPS 2018 paper](http://logic.cs.tsukuba.ac.jp/~sakaguchi/papers/flops2018.pdf).

Former implementations have small incompatibility problem with the original one and proofs using `rewrite [Finite.enum]unlock` might be broken. I have fixed this problem by putting both `enum` and indexing functions pair to `Finite.mixin_of`. I'm still not sure that this should be a part of MathComp, but I think that effort to maintain these changes (at least ones in fintype.v) is reasonably small.


pi8027:
@CohenCyril Thank you for the explanation.

@ggonthier 
> I think that there are better ways of doing this that what you've experimented with,

I have fixed an incompatibility issue of this. So it is better than the previous one. Do you still see any possible improvements?

> however, and also that we should put some thought in making the transport reduce, if we want this to be useful for computing directly with matrices.

- I don't understand "making the transport reduce" in this sentence.
- It is experimentally confirmed that the multiplication of n×n matrices of integers can be optimized and its execution time has been improved from O(n^5) to O(n^3) in extracted code. But we would have no improvements on computations inside Coq. (CoqEAL is a better way to do it.)

Issue #293, opened 2019-03-01
Note on how to edit .gitlab-ci.yml to put an overlay for non-backward-compatible changes


erikmd:
Following yesterday's meeting, I've taken a look at the GitLab CI doc https://docs.gitlab.com/ce/ci/yaml/ and it should be possible to slightly edit the `.gitlab-ci.yml` file (i.e. no need for separate overlay scripts) in case a PR breaks a "child CI project", to act as an overlay.

**Note:** I did not tested this on a realistic example nor automated the edition of `.gitlab-ci.yml` below, but I open this issue to document the steps, in case we want to automate this later on, if this need for overlay occurs in math-comp.

To sum up:
* assuming a math-comp PR 1234 breaks the `ci-fourcolor-dev` job
    ```yml
    ci-fourcolor-dev:
      extends: .ci-fourcolor
      variables:
        COQ_VERSION: "dev"
    ```
* assuming a non-backward-compatible, downstream patch has been pushed to a separated branch `fix-foo` in a fork `https://github.com/user/fourcolor.git`
* then one would only need to add a commit in the PR, replacing the above four lines with:
    ```yml
    ci-fourcolor-dev:
      extends: .ci-fourcolor
      variables:
        COQ_VERSION: "dev"
      # lines below have been added
      except:
        - /^pr-1234$/
        - /^github-branch$/

    ci-fourcolor-dev-1234:
      extends: .ci-fourcolor
      variables:
        COQ_VERSION: "dev"
        CONTRIB_URL: "https://github.com/user/fourcolor.git"
        CONTRIB_VERSION: "fix-foo"
      only:
        - /^pr-1234$/
        - /^github-branch$/
    ```
    where `github-branch` denotes the **name of the GitHub source branch of the PR, assuming it is part of the upstream (math-comp) repo**.  
    If the PR is opened from a math-comp **fork**, the two `- /^github-branch$/` lines should be skipped.
* so the upcoming build CI in `pr-1234` will use the specified branch, while after the merge in `master`, the regular branch of fourcolor will be used (assuming it has integrated the downstream patch).

erikmd:
@Zimmi48 I've just edited the description here following your suggestion in #324, thanks again!
@CohenCyril feel free to refine the prose if need be.

CohenCyril:
@erikmd could you put that as a mathcomp wiki page, and reference it in the isssue?

CohenCyril:
[...]

PR #324, opened 2019-04-04
Seq refactoring; adding `permutations`, extending  `allpairs`

ggonthier:
This covers extensions proposed by @MrSet and @pi8027 in #299 and #314, respectively, as well as some additional fixes for `seq`, `fintype` and `primitive_action`.
There is additional ongoing work on implementing `permutations` more efficiently (removing a linear time and space factor); this will be the object of another PR.
EDIT by Cyril: closes #314


CohenCyril:
@ggonthier I have a squashed and rebased version with resolved conflicts + cleansed spurious whitespaces. May I force-push?

ggonthier:
Did you also rebase wrt the ChangeLog upheaval? Otherwise it would make sense to squash wrt that before...

CohenCyril:
> Did you also rebase wrt the ChangeLog upheaval? Otherwise it would make sense to squash wrt that before...

Yes I did

CohenCyril:
@ggonthier the result it here: https://github.com/CohenCyril/math-comp/commits/seq-permutations-allpairs

CohenCyril:
I realize some comments of the commit log do not make sense anymore now it is squashed... so I am taking the liberty to rephrase slightly

ggonthier:
Looks great - good to merge, I'd say; thanks!

CohenCyril:
@erikmd I tried to follow your tutorial #293 but I must have missed something. Could you have a look?

erikmd:
@CohenCyril I just took a look and it seems:
* that your implementation of #293 was very fine: in particular:
    * in https://gitlab.com/math-comp/math-comp/pipelines/55269045 there is the good number of jobs
    * in https://gitlab.com/math-comp/math-comp/-/jobs/190933399 I indeed see
        `$ git rev-parse --verify HEAD`  
        `6fe341242be1d4e55ac23fa6f022e6d115b89bf1`
* that **coqbot** reporting was imprecise because it aggregated failures from different pipelines (this is maybe a very minor issue, but I Cc @Zimmi48 just in case)
* that your patch in the external branch is maybe incomplete: at first sight it should at least modify this line: https://github.com/math-comp/fourcolor/blob/master/theories/configurations.v#L2078

CohenCyril:
Failures are spurious now, one can merge.

ggonthier:
Shouldn't we remove the patch commit before merging?

CohenCyril:
> Shouldn't we remove the patch commit before merging?

We do not have to, but let's do the cleanup right now indeed.

CohenCyril:
Dear @ggonthier, I do not understand what happened here. Did you commit to this pr by accident?

ggonthier:
Yes, and then no: my local sync didn't react too well to all those forced pushes, and then I decided to tweak the RHS of `mem_permutations` while I was at it. I'm looking into another refactoring bit (consistency renaming or `perm` lemmas) which we should do while we're at it. I'll signal when you can straighten things out again...

CohenCyril:
Checks have passed. I will rebase and merge later

mathcomp/ssreflect/.#nseq.v
```diff
@@ -0,0 +1 @@
+gonthier@siatic.saclay.inria.fr.46535
```
erikmd: @ggonthier I guess this file should not be committed

CohenCyril: I will filter it out when I rebase

Zimmi48:
@erikmd Cyril pushed the branch `seq-permutations-allpairs` manually on GitLab. Thus, it ran without the overlays (since they are supposed to run only for the branch named `pr-324`). Because the recommendation is to use alternative names for building math-comp, the reported status from the CI run triggered by @coqbot did not override the status reported for the branch. Overall my recommendation is to just not push to GitLab directly unless you have a good reason to.

CohenCyril:
@Zimmi48 I do not even know how to push directly on gitlab, so I did do it on purpose... or at all?

erikmd:
@Zimmi48 thanks for your reply!

I guess the behavior is the same whether one directly push a branch to GitLab, or to GitHub because branches are mirrored to GitLab… so are you suggesting that all PRs should be opened from forks, rather than from internal, upstream branches on GitHub?
(this PR #324 was indeed in the latter case, as `seq-permutations-allpairs` is an upstream branch)

Zimmi48:
Thanks for the clarification. I did not realize that this branch was mirrored automatically. @erikmd You could modify your instructions in #293 to test for the branch name `pr-xxx` and also the GitHub branch name (similar to what we do in Coq's user overlays https://github.com/coq/coq/tree/master/dev/ci/user-overlays).

PR #330, opened 2019-04-17
Fix inheritance bugs of order and ssrnum (part of #270)

pi8027:
- Rebased on top of the current `master`.
- Remove the axiomless `normedType` structure.
- Redefine the intermediate structure between `idomainType` and `numDomainType`, which is `normedDomainType` (normed integral domain without an order) and have no inheritances of `lmodType`.
- Rename the compatibility module: `mc_1_7` -> `mc_1_8`.
- Remove some entries from the compatibility module which are already removed in MathComp 1.8.0.


CohenCyril:
I forced-push in math-comp experiment/order in order to trigger the CI.

PR #332, opened 2019-04-18
Remove unused `Require`s and a hint directive from interval.v

pi8027:
`Hint Rewrite intP` in interval.v would be a mistake. I have replaced it with `itvP`.

BTW, I guess that making the library dependencies sparse helps to increase the parallelizability of the compilation. Is there any way to detect/remove unused `Require`s systematically?

### DIFF
--- a/mathcomp/algebra/interval.v
+++ b/mathcomp/algebra/interval.v
@@ -4,7 +4,7 @@ Require Import mathcomp.ssreflect.ssreflect.
 From mathcomp
 Require Import ssrfun ssrbool eqtype ssrnat seq div choice fintype.
 From mathcomp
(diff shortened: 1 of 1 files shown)


mathcomp/algebra/interval.v
```diff
@@ -471,7 +471,7 @@ move=> x [[[] a|] [[] b|]] /itv_dec // [? ?];
      | rewrite (@ler_lt_trans _ x) // 1?ltrW ].
 Qed.
 
-Hint Rewrite intP : core.
+Hint Rewrite itvP : core.
```
CohenCyril: We should remove this directive. It was useless and it will remain so.

PR #337, opened 2019-04-28
Generalise use of `{pred T}` from coq/coq#9995

ggonthier:
Use `{pred T}` systematically for generic _collective_ boolean
predicate.
Use `PredType` to construct `predType` instances.
Instrument core `ssreflect` files to replicate these and other new
features introduces by coq/coq#9995 (`nonPropType` interface,
`simpl_rel` that simplifies with `inE`).

### DIFF
--- a/mathcomp/algebra/interval.v
+++ b/mathcomp/algebra/interval.v
@@ -331,7 +331,7 @@ Definition pred_of_itv (i : interval R) : pred R :=
         | BOpen_if b ub => x <= ub ?< if b
         | BInfty => true
       end].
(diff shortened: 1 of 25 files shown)


ggonthier:
@CohenCyril : failure of `finmap` seem to be due to the latter inlining the low-level representation of `predType` to construct a series of undocumented sub- packed classes, and directly using the low-level `predType` constructor. This is incompatible with coq/coq#9995, which changes that representation. I'm going to submit a patch for `finmap`, but in the meantime, unless you object, I'm going to forge ahead and merge this PR, as it it holding up the final commit in coq/coq#9995.

PR #340, opened 2019-04-30
Reimplement the hierarchy related tools in OCaml

### LINKED ISSUE #308: Clean, refactor and document the regression test for the structure hierarchy
Since the library is compiling before the merge, such a test would materialize the change operated by #291.

pi8027:
Closes #308

- The functionalities of the structure hierarchy related tools `hierarchy-diagram` and `hierarchy_test.py` are provided by an OCaml script `hierarchy.ml`.
- `test_suite/hierarchy_test.v` is deleted. Now make can generate it.


pi8027:
Hmm. I may need to remove `test_suite/hierarchy_test.v` from `Make` to generate `.coqdeps.d`...

CohenCyril:
Hi, isn't it better/possible to use Coq API rather than `coqtop` + string processing to generate the graph?

pi8027:
@CohenCyril Yes, I think so. I would have time to attempt it in June or July.

CohenCyril:
@gares do you think it is a good idea?

gares:
Yes, it is doable but it is not so trivial. I see two options:
- make a plugin defining a new command and then compiler a script that says
  ```coq
  Declare ML Module "my_plugin".
  Require Import all.v.
  MyCommand.
  ```
- build a new toplevel by linking coq (as a library). Initializing Coq is not trivial either (and you need to initialize it and require all .vo files). Something like serapi (we may even use it directly, but we may need to extend it with the queryes we want).

So for now I'd use this code and live with the text based interaction with Coq.

PR #342, opened 2019-05-05
add `deprecate` helper notation; no `perm` in non-`perm_eq` lemma names

ggonthier:
- add notation support for lemma renaming PRs, helping clients adjust
to the name change by emitting warning or raising errors when the old
name is used. The default is to emit warnings, but clients can change
this to silence (if electing to delay migration) or errors (to help
with actual migration). Usage:
Notation old_id := (deprecate old_id new_id) (only parsing).
—> Caveat 1: only prenex maximal implicit of `new_id` are preserved, so,
as `Notation` does not support on-demand implicits, the latter should
be added explicitly as in `(deprecate old new _ _)`.
—> Caveat 2: the warnings are emitted by a tactic-in-term, which
is interpreted during type elaboration. As the SSReflect elaborator may
re-infer type in arguments multiple times (notably, views and arguments
to `apply` and `rewrite`), clients may see duplicate warnings.
- use the `deprecate` facility to introduce the first part of the
refactoring of `seq` permutation lemmas : only lemmas concerning
`perm_eq` should have `perm` as a component of their name.
- document local additions in `ssreflect.v` and `ssrbool.v`

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -159,9 +159,14 @@ ci-fourcolor-dev:
     - make install
 
 ci-odd-order-8.7:
-  extends: .ci-odd-order
-  variables:
-    COQ_VERSION: "8.7"
+ extends: .ci-odd-order
+ variables:
+   COQ_VERSION: "8.7"
+
+ci-odd-order-8.8:
+ extends: .ci-odd-order
+ variables:
+   COQ_VERSION: "8.8"
 
 ci-odd-order-dev:
   extends: .ci-odd-order
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,9 +9,23 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Added
 
+- `nonPropType`, an interface for non-`Prop` types, and `{pred T}` and
+   `relpre f r`, all of which will be in the Coq 8.11 core SSreflect library.
+
+- `deprecate old_id new_id`, notation for `new_id` that prints a deprecation
+  warning for `old_id`; `Import Deprecation.Silent` turns off those warnings,
+  `Import Deprecation.Reject` raises errors instead of only warning.
+
 ### Changed
 
+- definition of `PredType` which now takes only a `P -> pred T` function;
+  definition of `simpl_rel` to improve simplification by `inE`. Both these
+  changes will be in the Coq 8.11 SSReflect core library.
+
 ### Renamed
+
+- `leq_size_perm` -> `uniq_min_size` (permuting conclusions)
+- `perm_uniq` -> `eq_uniq` (permuting assumptions)
 
 ### Misc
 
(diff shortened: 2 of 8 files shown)


ggonthier:
@gares: will try, but as this is a Coq < 8.9 only error it may be a diminishing return effort. I suspect it has to do with the ssr view abstraction code not purging delayed unification constraints before reabstracting evars, and Coq delaying the `return` predicate constraints for all `match`es in those old versions. It'll take me a while to check, as I'm not set up right now to debug with Coq 8.7.

gares:
Oh, btw the code dealing with views was mostly rewritten in the new tactic monad in 8.9 to support the new intro patterns including the tactic one, so it may be the case that it changed behavior in a subtle way. If you say the new behavior is better or even just ok, then It may not be worth spending much time on it (8.10beta is to be released this week).

ggonthier:
Ok, @gares, I somewhat tracked down the issue. It is specific to coq v8.7, so on the one hand it could go away as we drop 8.7 support. On the other hand, it's _not_ something that was fixed by the SSReflect view code refactoring as that was only in 8.9, and the characteristics of the issue are fairly alarming, so it might still be worth looking into, to make sure the root cause has indeed been fixed.
```
Require Import ssreflect.
Lemma foo U (l1 l2 : list U) : l1 = nil -> (l1 = nil -> l2 = nil) -> True. Proof. split. Qed.
Lemma bar T (t S : list T) : (t = nil -> S = nil) -> True.
move/(fun s1_0 ss21 => foo _ _ _ s1_0 ss21).
```
generates
```
Error:
Ltac call to "move (ssrmovearg) (ssrclauses)" failed.
Illegal application: 
The term
 "fun (s1_0 : t = nil) (ss21 : t = nil -> l = nil) => foo T t l s1_0 ss21"
of type "t = nil -> (t = nil -> l = nil) -> True"
cannot be applied to the terms
 "e" : "t = nil"
 "_top_assumption_" : "t = nil -> S = nil"
The 2nd term has type "t = nil -> S = nil" which should be coercible to
 "t = nil -> l = nil".
```
The most alarming is that this bug is _name-sensitive_; the error does not occur unless the name of the third parameter of `bar` _starts with an_ `S` !!! Also, replacing the abstract `T` with any fixed type makes the issue disappear. I'm quite at a loss as to what might cause such mischief.
I've not raised an issue Coq-side because of this being tied to both an old Coq version and to an SSReflect feature, but you may want to do so if the bug turns out to be on the Coq side.

ggonthier:
Unless someone objects, I'll merge this as soon as CI completes, as this PR is holding up a more substantial revision of the `perm_eq` lemmas. It was split off to allow reusing the (misnamed, and little used) `perm_uniq` lemma name without requiring an overlay for `fourcolor`: the plan is to first deprecate (here) the old `perm_uniq`, then update `fourcolor`, submit the main `perm_eq` PR, and finally update all clients to the new lemma names.

PR #344, opened 2019-05-06
remove dependence on Arith lemmas

soraros:
Ensure `by ...` does not invoke arithmetic hints from the Coq prelude.

### DIFF
--- a/mathcomp/ssreflect/ssrnat.v
+++ b/mathcomp/ssreflect/ssrnat.v
@@ -106,6 +106,10 @@ Set Implicit Arguments.
 Unset Strict Implicit.
 Unset Printing Implicit Defensive.
 
+(* Disable Coq prelude hints to improve proof script robustness. *)
+
+Remove Hints plus_n_O plus_n_Sm mult_n_O mult_n_Sm : core.
+
 (* Declare legacy Arith operators in new scope. *)
 
 Delimit Scope coq_nat_scope with coq_nat.
@@ -200,7 +204,7 @@ Lemma add1n n : 1 + n = n.+1.            Proof. by []. Qed.
 
 Lemma addn0 : right_id 0 addn. Proof. by move=> n; apply/eqP; elim: n. Qed.
 
-Lemma addnS m n : m + n.+1 = (m + n).+1. Proof. by elim: m. Qed.
+Lemma addnS m n : m + n.+1 = (m + n).+1. Proof. by apply/eqP; elim: m. Qed.
 
 Lemma addSnnS m n : m.+1 + n = m + n.+1. Proof. by rewrite addnS. Qed.
 
@@ -831,7 +835,7 @@ Proof. by move=> eq_op x; apply: eq_iteri; case. Qed.
 End Iteration.
 
 Lemma iter_succn m n : iter n succn m = m + n.
-Proof. by elim: n => //= n ->. Qed.
+Proof. by rewrite addnC; elim: n => //= n ->. Qed.
 
 Lemma iter_succn_0 n : iter n succn 0 = n.
 Proof. exact: iter_succn. Qed.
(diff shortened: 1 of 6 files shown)


soraros:
> By "check" I just meant trying to see how much of rest of the library compiles with your Remove Hints command, but we can leave that experiment for another PR.

@ggonthier this is easier than I thought

mathcomp/ssreflect/ssrnat.v
```diff
@@ -1647,14 +1647,17 @@ elim: p q => [p IHp|p IHp|] [q|q|] //=; rewrite !natTrecE //;
   by rewrite ?IHp ?nat_of_succ_gt0 ?(doubleS, doubleD, addn1, addnS).
 Qed.
 
+Lemma nat_of_muln_gt0 p q : (p * q)%positive = p * q :> nat.
+Proof.
```
soraros: I add this lemma so the proof script for `nat_of_mul_bin` is similar to `nat_of_add_bin`

mathcomp/ssreflect/ssreflect.v
```diff
@@ -4,6 +4,7 @@ From Coq Require Export ssreflect.
 Global Set SsrOldRewriteGoalsOrder.
 Global Set Asymmetric Patterns.
 Global Set Bullet Behavior "None".
+Remove Hints plus_n_O plus_n_Sm mult_n_O mult_n_Sm : core.
```
ggonthier: Please put this in `ssrnat.v` rather than `ssreflect.v`. The latter is meant to be a proxy for `Coq/plugins/ssr/ssreflect`, so apart from the `Require Import` it should only contain parameter settings and local advance copies of upcoming features (as we angle for backward compatibility with 2 prior versions of Coq). An conceivably some MathComp users might choose not to use `ssrnat` (e.g., to make it easier to use some of the not-so-parametrisable arithmetic automation), and those would expect to have these hints available.

mathcomp/ssreflect/ssrnat.v
```diff
@@ -831,7 +831,7 @@ Proof. by move=> eq_op x; apply: eq_iteri; case. Qed.
 End Iteration.
 
 Lemma iter_succn m n : iter n succn m = m + n.
-Proof. by elim: n => //= n ->. Qed.
+Proof. by elim: n => /= [| n ->]; rewrite ?addn0 // addnS. Qed.
```
ggonthier: Perhaps `rewrite addnC; elim: n => //= n ->` or `rewrite addnC; apply/eqP; elim: n` would be more elegant.

mathcomp/ssreflect/ssrnat.v
ggonthier: Well, `nat_of_addn_gt0` was not meant as a user-facing lemma, and was only there to factor a subproof common to `nat_of_add_bin` and `nat_of_add_mul`; Coq lacked the facilities (`Local Lemma`) to express that at the time this was written.
  I'm fine with switching to export `nat_of_[succ|addn|muln]_gt0`, but they need better names.
`nat_of_[succ|add|mul]_pos` seems the obvious choice.

mathcomp/ssreflect/ssreflect.v
soraros: This is never meant to stay at merge time. I just add it here to make sure the we are really clean from arith lemmas.
Now moved.

mathcomp/ssreflect/ssrnat.v
soraros: Wow!

mathcomp/ssreflect/ssrnat.v
```diff
@@ -105,6 +105,7 @@ Require Export Ring.
 Set Implicit Arguments.
 Unset Strict Implicit.
 Unset Printing Implicit Defensive.
+Remove Hints plus_n_O plus_n_Sm mult_n_O mult_n_Sm : core.
```
soraros: Not so sure where to place this line in ssrnat.

mathcomp/ssreflect/ssrnat.v
soraros: Changed to `nat_of_[succ|add|mul]_pos`.

mathcomp/ssreflect/ssrnat.v
ggonthier: The top of the file seems as good a place as any. Do separate this from the boilerplate flags with at least an empty line, and preferably a short rationale comment such as "Disable Coq prelude hints to improve proof script robustness."

soraros: @ggonthier, Is this what you mean?

ggonthier: Yes, that's perfect.

soraros:
@ggonthier Thank you for your guidance!

RalfJung:
This affects all proofs using `auto`/`eauto` as well in any development where ssreflect is (transitively) imported, right? So, e.g. anything using Iris now will also not have these hints in the `core` database any more?

It would be good if such global side-effects of ssreflect could at least be documented, similar to [what std++ does](https://gitlab.mpi-sws.org/iris/stdpp/#side-effects).

PR #345, opened 2019-05-07
refactor `seq` permutation theory

ggonthier:
- Change the naming of permutation lemmas so they conform to a
consistent policy: `perm_eq` lemmas have a `perm_` (_not_ `perm_eq`)
prefix, or sometimes a `_perm` suffix for lemmas that _prove_ `perm_eq`
using a property when there is also a lemma _using_ `perm_eq` for the
same property. Lemmas that do not concern `perm_eq` do _not_ have
`perm` in their name.
- Change the definition of `permutations` for a time- and space-
back-to-front generation algorithm.
- Add frequency tally operations `tally`, `incr_tally`, `wf_tally` and
`tally_seq`, used by the improved `permutation` algorithm.
- add deprecated aliases for renamed lemmas


ggonthier:
For the record: the latest (and hopefully last) CI failure appears to have a similar cause to the one discussed in #342, namely incomplete processing of unification constraints by the SSReflect plugin code; thus the problem may still be present in the current version of the code. Here it was solved by changing the `eq_big_perm` legacy notation from
```
Notation eq_big_perm :=
  ((fun P F r2 => @perm_big _ _ _ _ _ r2 P F) _ _) (only parsing).
```
to
```
Notation eq_big_perm :=
  ((fun R idx op I r1 P F r2 => @perm_big R idx op r1 r2 P F)
           _    _   _   _  _ _ _) (only parsing).
```
which are nearly identical: the one difference is that the evars for the holes inside the `fun ... => ...` expression in the first version have additional `(P := nil)` and `(F := nil)` in their context, and
perhaps this causes the unification algorithm to delay some constraints.

PR #346, opened 2019-05-10
add lemma on merge_sort_push invariant

palmskog:
As per an earlier discussion with @ggonthier and @ejgallego, a currently unstated key invariant of the `ss` argument to `merge_sort_push` is that its `n`th item has size `2^n` if it is not empty, so that the function only performs perfectly balanced merges (and does not collapse to insertion sort). Here is a lemma stating this fact.

### DIFF
--- a/mathcomp/ssreflect/path.v
+++ b/mathcomp/ssreflect/path.v
@@ -503,6 +503,30 @@ Fixpoint merge_sort_push s1 ss :=
   | s2 :: ss' => [::] :: merge_sort_push (merge s1 s2) ss'
   end.
 
+Definition size2s (s : seq T) n :=
+  if s isn't [::] then size s == 2^n else true.
+
+Definition nth_size2s ss n := forall m, size2s (nth [::] ss m) (n + m).
+
+Lemma nth_size2s_cons ss s n : nth_size2s (s :: ss) n -> nth_size2s ss n.+1.
+Proof. by move=> nsize2s m; rewrite addSn -addnS; apply: nsize2s. Qed.
+
+Lemma nth_size2s0 ss n : nth_size2s ss n.+1 -> nth_size2s ([::] :: ss) n.
+Proof. by move=> nsize2s [] //= m; rewrite addnS -addSn. Qed.
+
+Lemma merge_sort_push_inv ss s1 n : size s1 = 2^n ->
+  nth_size2s ss n -> nth_size2s (merge_sort_push s1 ss) n.
+Proof.
+elim: ss s1 n => [|[|t s] ss ihss] s1 n sizen.
+- move=> _ [] /=; last by move=> m; rewrite nth_nil.
+  by rewrite addn0; case: s1 sizen => //= _ s ->.
+- move=> nsize2s m; move: m (nsize2s m); case=> //= _; rewrite addn0.
+  by case: s1 sizen => //= _ s ->.
+- move=> nsize2s; move: (nsize2s 0); rewrite addn0 => /eqP /= => ssize2s.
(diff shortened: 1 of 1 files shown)


mathcomp/ssreflect/path.v
```diff
@@ -503,6 +503,31 @@ Fixpoint merge_sort_push s1 ss :=
...
+Definition size2s (s : seq T) n :=
+  if s isn't [::] then size s == 2^n else true.
+
+Definition nth_size2s ss n := forall m, size2s (nth [::] ss m) (n + m).
+
+Lemma nth_size2s_cons ss s n : nth_size2s (s :: ss) n -> nth_size2s ss n.+1.
+Proof. by move=> nsize2s m; rewrite addSn -addnS; apply: nsize2s. Qed.
+
+Lemma nth_size2s0 ss k : nth_size2s ss k.+1 -> nth_size2s ([::] :: ss) k.
+Proof. by move=> nsize2s; case=> //= n; rewrite addnS -addSn. Qed.
```
ejgallego: Maybe you could fold the case inside the intro pattern `move=> nsize2s [] //= n`.

mathcomp/ssreflect/path.v
```diff
@@ -503,6 +503,31 @@ Fixpoint merge_sort_push s1 ss :=
...
+
+Lemma merge_sort_push_inv ss s1 n : size s1 = 2^n ->
+  nth_size2s ss n -> nth_size2s (merge_sort_push s1 ss) n.
+Proof.
+elim: ss s1 n => //=.
```
ejgallego: You may want to intro the pattern a bit more: `elim: ss s1 n => [|s2 ss ihss] s1 n sizen`

mathcomp/ssreflect/path.v
```diff
@@ -503,6 +503,31 @@ Fixpoint merge_sort_push s1 ss :=
...
+  move=> s1 n sizen _; case => //=; last by move=> m; rewrite nth_nil.
+  by rewrite addn0; case: s1 sizen => //= _; move=> s ->.
```
ejgallego: Last `move` seems unnecessary.

ejgallego: In the sense that you are already on the intro pattern `case: s1 sizen => //= _ s ->`

ggonthier:
My main issue with this PR is that it introduces a fair amount of material for a result that is ultimately only useful as documentation of the algorithm (as MathComp does not provide facilities for reasoning about algorithmic complexity), while still not providing any actual natural language of the algorithm it purports to document.
  At the very least there needs to be an explanatory comment describing the algorithm (along the lines of what Hugo wrote in the Coq standard library implementation of the very same), and the proof itself should be clearly marked as a documentary result and be self-contained -- see my suggestion above.
   I'd actually written this review last week, but somehow failed to post it... sorry.

ybertot:
Dear @palmskog , we are currently looking at this PR for potential inclusion in the next release.  I just performed a rebase locally and we would mostly need to add the documentation that @ggonthier mentioned.  Will you have time to work on this on the next few days?

ybertot:
For the record, the result of the rebase is visible at https://github.com/ybertot/math-comp/tree/merge-sort-push-invariant
We could work together on adding the right comment to document the behavior of `merge_sort_push` that is illustrated by your added lemma.

CohenCyril:
> For the record, the result of the rebase is visible at https://github.com/ybertot/math-comp/tree/merge-sort-push-invariant

I think this `Remark` by @ggonthier should be merged, regardless of which other additions are planned.

PR #351, opened 2019-05-28
Replace eqVneq to destruct not only x == y, but also y == x

anton-trunov:
This PR changes `eqVneq` lemma to be able to do case analysis on `x == y` and `y == x` simultaneously.
This change lets us simplify a pattern I see sometimes: `rewrite eq_sym; case: eqP ...`. I tried to demonstrate this by using `eqVneq` in mathcomp's libraries (I found those places by grepping and made the changes rather formally).
Any feedback would be highly appreciated!


mathcomp/ssreflect/eqtype.v
```diff
@@ -196,6 +196,15 @@ Proof. exact/eqP/eqP. Qed.
 
 Hint Resolve eq_refl eq_sym : core.
 
+Variant eq_xor_neq_sym (T : eqType) (x y : T) : bool -> bool -> Set :=
+  | EqNotNeqSym of x = y : eq_xor_neq_sym x y true true
+  | NeqNotEqSym of x <> y : eq_xor_neq_sym x y false false.
+
+Lemma eqsP (T : eqType) (x y : T) : eq_xor_neq_sym x y (x == y) (y == x).
```
pi8027: `case: eqsP` finds the first occurrence of `?y == ?x` and generates a hypothesis `?x = ?y` or `?x <> ?y`. `eq_xor_neq_sym x y (y == x) (x == y)` might be better.

anton-trunov: @pi8027 Good catch! Thank you very much.

CohenCyril:
I am in favor of adding this lemma, as it looks useful. But given how many times it is used I would prefer a longer name, that reflects contains the word `sym` rather than just `s`, as in `eqPsym` (which I prefer) or `eq_symP`.

anton-trunov:
Thank you @CohenCyril. I'm happy to change names to anything that make more sense than the current preliminary naming.
I've always thought that `P` should be the last character for views. Are there any exceptions in mathcomp?

CohenCyril:
> I've always thought that `P` should be the last character for views. Are there any exceptions in mathcomp?

Yes, suffixes `Pl` (for left), `Pr` (for right), `Pn` (for neg), and more recently `Pdep`

thery:
we should use machine learning to automatically generate names :smiley:

mathcomp/ssreflect/eqtype.v
```diff
@@ -196,6 +196,15 @@ Proof. exact/eqP/eqP. Qed.
 
 Hint Resolve eq_refl eq_sym : core.
 
+Variant eq_xor_neq_sym (T : eqType) (x y : T) : bool -> bool -> Set :=
+  | EqNotNeqSym of x = y : eq_xor_neq_sym x y true true
+  | NeqNotEqSym of x <> y : eq_xor_neq_sym x y false false.
```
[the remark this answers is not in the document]
anton-trunov: Thank you for the feedback, fixed now. I'm keeping recent changes as separate commits to make the review process easier. I'll squash the commits later.

CohenCyril:
> there is already an eqVneq lemma stating the x=y / x!=y alternative, so replacing that lemma would be preferable.

I agree, I think it would be great to rename `eqPsym` to `eqVneq` and remove the old `eqVneq` lemma
(and same for `eqsVneq`).

anton-trunov:
Great suggestions, thank you @ggonthier and @CohenCyril. Please take a look if I missed anything.

CHANGELOG.md
```diff
@@ -5,6 +5,16 @@ Last releases: [[1.9.0] - 2019-05-22](#190---2019-05-22) and [[1.8.0] - 2019-04-
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
+## [Unreleased]
+
+### Changed
+
+- `eqVneq` lemma is changed from `{x = y} + {x != y}` to
+  `eq_xor_neq x y (y == x) (x == y)` which allows to use it as
+   a view like `eqP`, in addition providing simultaneous destruction
```
CohenCyril: I would remove the comparison with `eqP` which is less relevant than the comparison to `eqVneq`.

mathcomp/algebra/polydiv.v
```diff
@@ -2239,8 +2238,7 @@ Proof.
 apply/eqP/idP=> [d0|]; last first.
   case/or3P; [by move/eqP->; rewrite div0p| by move/eqP->; rewrite divp0|].
   by move/divp_small.
-case: (eqVneq p 0) => [->|pn0]; first by rewrite eqxx.
-case: (eqVneq q 0) => [-> | qn0]; first by rewrite eqxx orbT.
+case: (eqVneq p 0) => [->//|pn0]; case: (eqVneq q 0) => [->//| qn0].
```
CohenCyril: this can even be shorted to `case: eqVneq => // pn0; case: eqVneq => // qn0`

mathcomp/ssreflect/eqtype.v
```diff
@@ -196,6 +196,15 @@ Proof. exact/eqP/eqP. Qed.
 
 Hint Resolve eq_refl eq_sym : core.
 
+Variant eq_xor_neq (T : eqType) (x y : T) : bool -> bool -> Set :=
+  | EqNotNeq of x = y : eq_xor_neq x y true true
+  | NeqNotEq of x != y : eq_xor_neq x y false false.
+
+Lemma eqVneq (T : eqType) (x y : T) : eq_xor_neq x y (y == x) (x == y).
+Proof. by rewrite eq_sym; case: eqP=> [|/eqP]; constructor. Qed.
```
CohenCyril: `by rewrite eq_sym; case: (altP eqP); constructor.`

mathcomp/solvable/abelian.v
```diff
@@ -836,9 +836,9 @@ Qed.
 
 Lemma rank_gt0 G : ('r(G) > 0) = (G :!=: 1).
 Proof.
-case: (eqsVneq G 1) => [-> |]; first by rewrite rank1 eqxx.
-case: (trivgVpdiv G) => [-> | [p p_pr]]; first by case/eqP.
-case/Cauchy=> // x Gx oxp ->; apply: leq_trans (p_rank_le_rank p G).
+case: (eqsVneq G 1) => [-> |]; first by rewrite rank1.
+case: (trivgVpdiv G) => [-> | [p p_pr]]; first by rewrite eqxx.
```
CohenCyril: `case: (trivgVpdiv G) => [/eqP->// | [p p_pr]].`

CHANGELOG.md
```diff
@@ -5,6 +5,16 @@ Last releases: [[1.9.0] - 2019-05-22](#190---2019-05-22) and [[1.8.0] - 2019-04-
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
+## [Unreleased]
```
CohenCyril: As per https://github.com/math-comp/math-comp/wiki/Agenda-of-the-April-23rd-2019-meeting-9h30-to-12h30#avoiding-issues-with-changelog new changelog entries now go to `CHANGELOG_UNRELEASED.md`

CHANGELOG.md
anton-trunov: done

mathcomp/algebra/polydiv.v
anton-trunov: much better, thanks

mathcomp/ssreflect/eqtype.v
anton-trunov: I keep forgetting about `altP`, thanks

mathcomp/solvable/abelian.v
anton-trunov: done

CHANGELOG.md
anton-trunov: I created the file, keeping `## [Unreleased]` section, which looks a bit weird since the whole file is devoted to unreleased stuff, but I think it should make it easier to move entries from `CHANGELOG_UNRELEASED.md` to `CHANGELOG.md`. I can change that if you'd like, though.

CohenCyril: I think this is fine

CHANGELOG_UNRELEASED.md
```diff
@@ -0,0 +1,17 @@
...
+[here](https://github.com/math-comp/math-comp/wiki/Agenda-of-the-April-23rd-2019-meeting-9h30-to-12h30#avoiding-issues-with-changelog).
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
+
+## [Unreleased]
+
+### Changed
+
+- `eqVneq` lemma is changed from `{x = y} + {x != y}` to
+  `eq_xor_neq x y (y == x) (x == y)` which allows to use as a view and provide
```
CohenCyril: One last comment: I do not think this counts as a view, so you could write:
- `eqVneq` lemma is changed from `{x = y} + {x != y}` to `eq_xor_neq x y (y == x) (x == y)`, on which a case analysis performs simultaneous replacement of expressions of the form `x == y` and `y == x` by `true` or `false`, while keeping the ability to use it in the way it was used before.

anton-trunov: done!

CohenCyril:
I will probably squash and merge.

anton-trunov:
Great! Thanks to everyone for your time, patience and the great suggestions!

PR #372, opened 2019-07-18
Taking advantage of Coq 8.10 ssreflect new features

CohenCyril:

Mostly testing ltac in notations in views:
- Proposing a few different syntax for rewriting in views, up to debate.
While doing so, I encountered a problem with parsing `! term` in ssr rewrite but only when followed by a clear. Cf https://github.com/coq/coq/issues/10531. The original intent was to be able to use `=> x /!inE x_mem` to perform simplification between two introductions.
- Proposing `=> /[->]`syntax to substitute with an equality everywhere in the goal and context. 

##### Things done/to do

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -77,15 +77,6 @@ make-coq-latest:
     - tags
     - merge_requests
 
-coq-8.7:
-  extends: .opam-build
-
-coq-8.8:
-  extends: .opam-build
-
-coq-8.9:
-  extends: .opam-build
-
 coq-dev:
   extends: .opam-build
 
@@ -133,16 +124,6 @@ coq-dev:
     - make -j "${NJOBS}"
     - make install
 
-ci-fourcolor-8.7:
-  extends: .ci-fourcolor
-  variables:
-    COQ_VERSION: "8.7"
-
-ci-fourcolor-8.8:
-  extends: .ci-fourcolor
-  variables:
-    COQ_VERSION: "8.8"
-
 ci-fourcolor-dev:
   extends: .ci-fourcolor
   variables:
@@ -158,15 +139,6 @@ ci-fourcolor-dev:
     - make -j "${NJOBS}"
     - make install
 
-ci-odd-order-8.7:
- extends: .ci-odd-order
- variables:
-   COQ_VERSION: "8.7"
-
-ci-odd-order-8.8:
- extends: .ci-odd-order
- variables:
-   COQ_VERSION: "8.8"
 
 ci-odd-order-dev:
   extends: .ci-odd-order
@@ -183,15 +155,6 @@ ci-odd-order-dev:
     - opam pin add -n -k path coq-lemma-overloading .
     - opam install -y -v -j "${NJOBS}" coq-lemma-overloading
 
-ci-lemma-overloading-8.8:
-  extends: .ci-lemma-overloading
-  variables:
-    COQ_VERSION: "8.8"
-
-ci-lemma-overloading-8.9:
-  extends: .ci-lemma-overloading
-  variables:
-    COQ_VERSION: "8.9"
 
 ci-lemma-overloading-dev:
   extends: .ci-lemma-overloading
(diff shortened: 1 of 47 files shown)


gares:
Can't you just say that `/[x]` stands for "rewrite ! x" and that `->` is special in that context and means `{}top in *`? It is just that `/!inE` is not much shorter than `/[inE]` and the latter is a bit more future proof, eg you may need to apply the rewrite rule to specialize it, or write a multi rule (assuming you can define a notation for `[x,y,..,z]` in there).

CohenCyril:
> Can't you just say that `/[x]` stands for "rewrite ! x" and that `->` is special in that context and means `{}top in *`? It is just that `/!inE` is not much shorter than `/[inE]` and the latter is a bit more future proof, eg you may need to apply the rewrite rule to specialize it, or write a multi rule (assuming you can define a notation for `[x,y,..,z]` in there).

Sometimes it is with `!` sometimes it is without, moreover a general design pattern of mathcomp is that bracketed notations start with a keyword/symbol as in `[&& ..]`, `[:: ...]`, `[seq ...]`, `[tuple ...]`, `[eqType ...]` ... so using `[ rule ]` for rewriting breaks this pattern. At the very least it should be `[! rule]` (for bang) `[1 rule]` for once... and maybe it is alright... let me see.

CohenCyril:
@gares 
> `[->]` means `rewrite {}top in *`

It's more like `elim: _ / top in *` except I found no stable way of generalizing all the local context :disappointed:

See my commit: the current version seems to work but is a bit broken for hypothesis that depend on others, where I have to default to Coq's less powerful `fold` tactic with a `in` and potentially screwing up section hypothesis... moreover it does not handle multi-rules and quantified rules...

I have tried more clever versions since then... and my last attempt, which is supposed to handle everything is:
```coq
Ltac discharge_all_but x := do ![match goal with 
  H : _ |- _ => (* discharges a let in which is not x *)
                (move: x @H => x)
                 (* otherwise discharges a n assumption which is not x *)
                || (move: x H => x);
                (* tests whether H is still there, which means it is a section hypothesis *)
                tryif move: H => _
                then fail else idtac
  end].

Lemma lock_goal A : locked A -> A.
Proof. by rewrite -lock. Qed.
Lemma unlock_goal A : A -> locked A.
Proof. by rewrite -lock. Qed.

(* implementation of `/[<-]` *)
Ltac subst_eq :=
  let top_eq := fresh "_top_eq_" in move=> top_eq;
  apply: lock_goal; discharge_all_but top_eq;
  (* try dependent elim, otherwise (if it's a multirule) rewriting *)
  ((elim: _ / top_eq) || (rewrite -!{}top_eq));
  (* the `intros` here assumes that `move:` discharged hyps with
  the names they used to have, which seems like a decent invariant
  unfortunately, it seems to break with hyps introduced by `=> +` *)
  intros; apply: unlock_goal.
```
which works very well, except in the presence of `=> +` (see comment)

gares:
I thought you were calling the primitive `subst` tactic for `/[->]`.
To my knowledge, any context manipulation in ltac1 is very hard to write, so I would not even try.

amahboubi:
This PR needs an assignee. @gares ?

gares:
The more I think about this, the more I've the impression that you try to be too generic.

Eg many of the use cases that seem very legit could be also covered by `=> /mxE /inE`, that is ad hoc calls to `rewrite ! lemma` (and at some extent they are unfolding lemmas, so `/e` reminds me that, but maybe it's too confusing for a newcomer). What do you think about this? It's a baby step but still useful I believe.

The `subst` thing seems ahead of time: with Ltac2 (or anything else than ltac1) you can surely code something sensible, if the already existing `subst` does not behave as you like. I'm not sure we want that ltac code around.

affeldt-aist:
The relaease managers propose to put this PR in the 1.12.0 miletone. Are you ok?

CohenCyril:
> The relaease managers propose to put this PR in the 1.12.0 miletone. Are you ok?

I agree

CohenCyril:
As a side note, it was discussed in a math-comp meeting that substitution should mention the context hypotheses  explicitly anyway.

CohenCyril:
I splitted this PR in bits. The first bit is available as #501

PR #377, opened 2019-08-27
Reorder the arguments of (comparable_)ltgtP and remove the use of the compatibility module `ssrnum.mc_1_9` (part of #270)

pi8027:

- The order of arguments in `compare` and `(comparable_)ltgtP` was wrong. This PR fixes it.
- This PR adds CI overlays which don't use the compatibility module `ssrnum.mc_1_9`. 

I have also fixed `ltngtP`, but it's independent of #270. So I separated it off to #378.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -132,6 +132,8 @@ coq-dev:
   script:
     - make -j "${NJOBS}"
     - make install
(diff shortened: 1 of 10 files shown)


pi8027:
@CohenCyril Could you please merge this one?

PR #378, opened 2019-08-30
Reorder the arguments in `compare_nat` and `ltngtP`

pi8027:

This PR attempts the option 2 of https://github.com/math-comp/math-comp/pull/270#issuecomment-503000130 and reorders the  arguments in `compare_nat` and `ltngtP`
from `ltngtP m n : compare_nat m n (m <= n) (n <= m) (m < n) (n < m) (n == m) (m == n)`
to `ltngtP m n : compare_nat m n (n == m) (m == n) (n <= m) (m <= n) (n < m) (m < n)`.

The latter one is more natural because the last type argument will be attempted first to match with a subterm in the goal by the `case:` tactic. This patch was a part of #377 that is independent of #270.

##### Things done/to do

### DIFF
--- a/CHANGELOG_UNRELEASED.md
+++ b/CHANGELOG_UNRELEASED.md
@@ -93,6 +93,15 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   (which statement was modified to remove the dependency in `eqType`),
   and `order_path_min`.
 
+- `compare_nat` type family and `ltngtP` comparison predicate are changed
+  from `compare_nat m n (m <= n) (n <= m) (m < n) (n < m) (n == m) (m == n)`
+  to `compare_nat m n (n == m) (m == n) (n <= m) (m <= n) (n < m) (m < n)`,
+  to make it tries to match subterms with `m < n` first, `m <= n`, then
+  `m == n`.
+  + The compatibility layer for the version 1.9 is provided as the
+    `ssrnat.mc_1_9` module. One may compile proofs compatible with the version
+    1.9 in newer versions by using this module.
+
 ### Infrastructure
 
 - `Makefile` now supports the `test-suite` and `only` targets. Currently,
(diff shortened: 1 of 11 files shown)


pi8027:
I think we can continue some discussion about the following [Cyril's remark](https://github.com/math-comp/math-comp/pull/270#issuecomment-503000130) here, if needed.

> But, @ggonthier, @amahboubi, @thery, @gares, @LaurenceRideau, @pi8027 there is an issue for which I cannot take a light decision: `ltgtP` and `ltrgtP` do not follow the same structure as the `ltngtP` lemma from ssrnat:
> in `ltgtP` (and `ltrgtP`) lemma, the operators come in the following order `==`, `<=`, `<` while in `ltngtP` it is `<=`, `<` and `==`. For consistency they should be ordered in the same way, and thus there are now two options:
> 
> 1. For (more) backward compatibility they should be in the order  [`<=`, `<`, `==`] (as in `ltngtP`),
> 2. but for the sake of intuition as noted by @thery and @gares in #233 it should rather be [`==`, `<=`, `<`] everywhere (as in `ltgtP`) because the last item is matched first. I think I a missing a part of the history of the lemma in order to take a decision.
> 
> Which of 1. and 2. would be best?
> 
> Also, in `leqP`, `leP` ..., operators are in the order `<=` first, and `<` while the same intuition as in the paragraph above would lead to think it should be the converse.

Actually this PR fixes `compare_nat` and `ltngtP` as its option 2 indicated. As a result, I also needed to patch fourcolor: math-comp/fourcolor#18. Any other libraries haven't been broken in CI.

CohenCyril:
I'm fine with your changes

pi8027:
BTW, having `minn` and `maxn` in the arguments of `leqP`, `ltnP`, and `ltngtP` could be nice. But we need to reorder definitions and lemmas in ssrnat.

amahboubi:
Hi. I did not review this PR. What is blocking the merge here?

pi8027:
@amahboubi This is a compatibility breaking change. There was a discussion on this change in #233, but it has never realized before. So we probably should have more discussion before merging. (But personally I'm fine with this change as it is.)

amahboubi:
@pi8027 Thanks for the answer. Could you please add this item to the agenda of the next dev meeting?

pi8027:
@amahboubi Yes, I will do that.

Issue #359, opened 2019-06-18
Theorem names in ssrnat (switching long suffixes to short suffixes)


thery:
Should theorems like 
``odd_add : forall m n : nat, odd (m + n) = odd m (+) odd n``
be named ``oddD``?

hivert:
For further reference: in `seq.v`, lemma `take_addn` and `rot_addn` should be renamed.

PR #381, opened 2019-09-03
More lemmas on seqs

hivert:

Here is a bunch of basic seq related lemmas I've been using from time to time. In case you find them worth being included.

##### Things done/to do

### DIFF
--- a/mathcomp/ssreflect/seq.v
+++ b/mathcomp/ssreflect/seq.v
@@ -277,6 +277,10 @@ Lemma cat_cons x s1 s2 : (x :: s1) ++ s2 = x :: s1 ++ s2. Proof. by []. Qed.
 Lemma cat_nseq n x s : nseq n x ++ s = ncons n x s.
 Proof. by elim: n => //= n ->. Qed.
 
+Lemma nseq_addn n1 n2 x :
+  nseq (n1 + n2) x = nseq n1 x ++ nseq n2 x.
+Proof. by rewrite cat_nseq /nseq /ncons iter_add. Qed.
+
 Lemma cats0 s : s ++ [::] = s.
 Proof. by elim: s => //= x s ->. Qed.
 
@@ -742,6 +746,40 @@ move=> lt_i_n0 s; case lt_n0_s: (n0 < size s).
 by rewrite -{1}[s]cats0 take_cat lt_n0_s /= cats0.
 Qed.
 
+Lemma take_take i j : i <= j -> forall s, take i (take j s) = take i s.
+Proof.
+move=> ij s; elim: s i j ij => [// | a s IHs] [|i] [|j] //=.
+by rewrite ltnS => /IHs ->.
+Qed.
+
+Lemma take_drop i j s : take i (drop j s) = drop j (take (i + j) s).
+Proof. by rewrite addnC; elim: s i j => // x s IHs [|i] [|j] /=. Qed.
+
+Lemma take_addn i j s : take (i + j) s = take i s ++ take j (drop i s).
+Proof.
+elim: i j s => [|i IHi] [|j] [|a s] //; first by rewrite take0 addn0 cats0.
+by rewrite addSn /= IHi.
+Qed.
+
+Lemma takeC i j s : take i (take j s) = take j (take i s).
+Proof.
+wlog i_le_j : i j / i <= j.
+  by move=> Hwlog; case: (leqP i j) => [|/ltnW] /Hwlog ->.
+rewrite take_take // [RHS]take_oversize // (leq_trans _ i_le_j) //.
+elim: i s {i_le_j} => [|i IHi] s; first by rewrite take0.
+by case: s => [|a s]//; rewrite /= ltnS.
+Qed.
+
+Lemma take_nseq i j x : i <= j -> take i (nseq j x) = nseq i x.
+Proof. by move=>/subnKC <-; rewrite nseq_addn take_size_cat // size_nseq. Qed.
+
+Lemma drop_nseq i j x : drop i (nseq j x) = nseq (j - i) x.
+Proof.
+case: (leqP i j) => [/subnKC {1}<-|/ltnW j_le_i].
+  by rewrite nseq_addn drop_size_cat // size_nseq.
+by rewrite drop_oversize ?size_nseq // (eqP j_le_i).
+Qed.
+
 (* drop_nth and take_nth below do NOT use the default n0, because the "n"  *)
 (* can be inferred from the condition, whereas the nth default value x0    *)
 (* will have to be given explicitly (and this will provide "d" as well).   *)
@@ -833,6 +871,11 @@ Proof. by rewrite !has_count count_rev. Qed.
 
 Lemma all_rev a s : all a (rev s) = all a s.
 Proof. by rewrite !all_count count_rev size_rev. Qed.
+
+Lemma rev_nseq n x : rev (nseq n x) = nseq n x.
+Proof.
+by elim: n => [// | n IHn]; rewrite -{1}(addn1 n) nseq_addn rev_cat IHn.
+Qed.
 
 End Sequences.
 
@@ -1183,7 +1226,7 @@ Proof. by rewrite -has_pred1 has_count -eqn0Ngt; apply: eqP. Qed.
 Lemma count_uniq_mem s x : uniq s -> count_mem x s = (x \in s).
 Proof.
 elim: s => //= y s IHs /andP[/negbTE s'y /IHs-> {IHs}].
-by rewrite in_cons; case: eqVneq => // <-; rewrite s'y.
+by rewrite in_cons; case: (eqVneq y x) => // <-; rewrite s'y.
 Qed.
 
 Lemma filter_pred1_uniq s x : uniq s -> x \in s -> filter (pred1 x) s = [:: x].
@@ -1321,7 +1364,7 @@ Section NthTheory.
 Lemma nthP (T : eqType) (s : seq T) x x0 :
   reflect (exists2 i, i < size s & nth x0 s i = x) (x \in s).
 Proof.
-apply: (iffP idP) => [|[n Hn <-]]; last by apply mem_nth.
+apply: (iffP idP) => [|[n Hn <-]]; last exact: mem_nth.
 by exists (index x s); [rewrite index_mem | apply nth_index].
 Qed.
 
@@ -2342,6 +2385,19 @@ Qed.
 
 Lemma iota_uniq m n : uniq (iota m n).
 Proof. by elim: n m => //= n IHn m; rewrite mem_iota ltnn /=. Qed.
+
+Lemma take_iota k m n : take k (iota m n) = iota m (minn k n).
+Proof.
+rewrite /minn; case: ltnP => [lt_k_n|le_n_k].
+  by elim: k n lt_k_n m => [|k IHk] [|n]//= H m; rewrite IHk.
+by apply: take_oversize; rewrite size_iota.
+Qed.
+
+Lemma drop_iota k m n : drop k (iota m n) = iota (m + k) (n - k).
+Proof.
+by elim: k m n => [|k IHk] m [|n]//=; rewrite ?addn0// IHk addSn addnS subSS.
+Qed.
+
 
 (* Making a sequence of a specific length, using indexes to compute items. *)
 
--- a/CHANGELOG_UNRELEASED.md
+++ b/CHANGELOG_UNRELEASED.md
@@ -85,6 +85,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Added lemmas on commutation with difference, big sum and prod:
   `commrB`, `commr_sum`, `commr_prod`.
 
+- Added a few basic seq lemmas about `nseq`, `take` and `drop`:
+  `nseq_addn`, `take_take`, `take_drop`, `take_addn`, `takeC`,
+  `take_nseq`, `drop_nseq`, `rev_nseq`, `take_iota`, `drop_iota`.
+
+
 ### Changed
 
 - `eqVneq` lemma is changed from `{x = y} + {x != y}` to


mathcomp/ssreflect/seq.v
```diff
@@ -718,6 +722,10 @@ have [le_sn | lt_ns] := leqP (size s) n0; first by rewrite take_oversize.
 by rewrite size_takel // ltnW.
 Qed.
 
+Lemma size_take_leq s :
```
pi8027: If `size_take` is changed as follows, we wouldn't need `size_take_leq`. (Indeed this change is backward incompatible.)
```coq
Lemma size_take s : size (take n0 s) = minn n0 (size s).
```

mathcomp/ssreflect/seq.v
```diff
@@ -750,6 +758,47 @@ move=> lt_i_n0 s; case lt_n0_s: (n0 < size s).
 by rewrite -{1}[s]cats0 take_cat lt_n0_s /= cats0.
 Qed.
 
+Lemma take_take i j : i <= j -> forall s, take i (take j s) = take i s.
```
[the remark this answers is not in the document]
hivert: In my opinion, my form of statement is slightly easier to use. Indeed, in the other case (i.e. when `i >= j`), one conclude instantaneously with `take_oversize`. Your form of statement is good when you don't know the relative order of `i` and `j` which is the case for  `takeC` but, in my experience, occur very rarely. Hence my choice of statement.

Anyway, I would gladly put the two statements, provided we agree on the names.

What do you think ?

mathcomp/ssreflect/seq.v
```diff
@@ -750,6 +758,47 @@ move=> lt_i_n0 s; case lt_n0_s: (n0 < size s).
...
+Proof.
+move=> ij s; elim: s i j ij => [// | a s IHs] [|i] [|j] //=.
+by rewrite ltnS => /IHs ->.
+Qed.
+
+Lemma take_drop i j s : take i (drop j s) = drop j (take (i + j) s).
```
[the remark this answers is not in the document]
hivert: Again, this second form is only useful when `j >= i` otherwise one conclude with `drop_oversize`. And when `j >= i`, passing from `drop_take` to `take_drop` is just a matter of rewriting with `subnK`...

Again, if there is an agreement, I'm ok to add both.

mathcomp/ssreflect/seq.v
hivert: In my code, I have dozens of occurrence of `size (take n0 s)` under the assumption that `le : n0 <= size s)`.  Currently we have to write 
```    rewrite size_map size_take -/(minn _ _) (minn_idPl le).``` 
which is not so bad, except that I can't manage to remember the name `minn_idPl`. I figured out that 
```    rewrite size_take_leq le.``` 
was the easiest and fastest way to do it. With `minn` you'll have to do 
```    rewrite size_take_minn (minn_idPl le).```

Again, we could include more than one statement if we agree on the name. If the maintainers don't want to have yet another lemma, I'll also fully understand.

Comment (ggonthier) [reply mathcomp/ssreflect/seq.v]: > In my code, I have dozens of occurrence of `size (take n0 s)` under the assumption that `le : n0 <= size s`.

You should be using `size_takel`, then.

Comment (hivert) [reply mathcomp/ssreflect/seq.v]: So true ! Ok ! Let's remove this one.

File: mathcomp/ssreflect/seq.v (hunk 2 above)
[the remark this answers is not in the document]
Comment (pi8027) [reply mathcomp/ssreflect/seq.v]: Now I see how useful your `take_take` is, and would like to have both. Currently, I have no idea how to name them.

File: mathcomp/ssreflect/seq.v (hunk 3 above)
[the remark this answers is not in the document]
Comment (pi8027) [reply mathcomp/ssreflect/seq.v]: Since there is no name conflict here, would you add `drop_take`? Otherwise, I would like to do it by myself.

File: mathcomp/ssreflect/seq.v [hunk 4]
```diff
@@ -750,6 +758,47 @@ move=> lt_i_n0 s; case lt_n0_s: (n0 < size s).
...
+Proof.
+elim: j s => [|j IHs] s /=; first by rewrite !drop0 addn0.
+by case: s => [// | a s]; rewrite addnS /=.
+Qed.
+
+Lemma take_addn i j s : take (i + j) s = take i s ++ take j (drop i s).
```
anton-trunov: Here is an alternative one-line proof of `take_addn` showing that the lemma currently named `take_take` is useful, because it does not enforce a particular form of one of `take`'s arguments:
```coq
by rewrite take_drop addnC -(take_take i (j + i)) ?leq_addl ?cat_take_drop.
```

mathcomp/ssreflect/seq.v
```diff
@@ -750,6 +758,47 @@ move=> lt_i_n0 s; case lt_n0_s: (n0 < size s).
...
+Lemma takeC i j s : take i (take j s) = take j (take i s).
+Proof.
+wlog i_le_j : i j / i <= j.
+  by move=> Hwlog; case: (leqP i j) => [|/ltnW] /Hwlog ->.
+rewrite take_take // [RHS]take_oversize //; apply: (leq_trans _ i_le_j).
+elim: i s {i_le_j} => [|i IHi] s; first by rewrite take0.
+by case: s => [|a s]//; rewrite /= ltnS.
+Qed.
+
+Lemma take_nseq i j x : i <= j -> take i (nseq j x) = nseq i x.
```
anton-trunov: I like this proof! By using `subnKC` it's possible to shave off `addnC` here:
```coq
move=>/subnKC <-; move: (j-i) => d.
by rewrite nseqD take_size_cat // size_nseq.
```

mathcomp/ssreflect/seq.v
```diff
@@ -750,6 +758,47 @@ move=> lt_i_n0 s; case lt_n0_s: (n0 < size s).
...
+Proof.
+move=>/subnK <-; move: (j-i) => d.
+by rewrite addnC nseqD take_size_cat // size_nseq.
+Qed.
+
+Lemma drop_nseq i j x : drop i (nseq j x) = nseq (j - i) x.
```
anton-trunov: Let me share a trick I sometimes use. Often times we'd like to weaken our hypothesis on the top of the goal stack in two different ways. We can define a duplicating lemma like this one:
```coq
Lemma duP {A : Type} : A -> A * A.
Proof. by []. Qed.
```
which can be later used like so:
```coq
case: (leqP i j) => [/subnKC {1}<- | /ltnW /duP[j_le_i /eqP->]].
- by rewrite nseqD drop_size_cat // size_nseq.
by rewrite drop_oversize // size_nseq.
```
This, admittedly, may result in less readable proofs, but it lets one make less context shuffling steps.

mathcomp/ssreflect/seq.v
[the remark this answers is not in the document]
pi8027: I found my old proofs which use my version of `take_take` and `drop_take`. But I really didn't use them so much, and I did know `i <=j` does hold when I use `take_take`, as you said. Now I think both my suggestions were useless, so please close them freely.

mathcomp/ssreflect/seq.v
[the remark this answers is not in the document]
pi8027: Now I think `drop_take` is useless, but there is a one-line proof of `take_drop` without using `drop_take`:
```
Proof. by rewrite addnC; elim: s i j => // x s IHs [|i] [|j] /=. Qed.
```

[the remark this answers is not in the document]
hivert: Thanks for the short proof ! I'm updated my PR.

mathcomp/ssreflect/seq.v
hivert: Unfortunately at the point of the script where `take_addn` is natural, I can't use `cat_take_drop` without closing and reopening the sequence section. Indeed  `cat_take_drop` only works with the section variable `n0`. Since the induction proof is not that complicated, I decided to stick with it.

mathcomp/ssreflect/seq.v
hivert: Yes ! Thanks .

mathcomp/ssreflect/seq.v
hivert: Good trick ! However, I'd rather not use such a trick in a single place of the `mathcomp` code. I'll gladly use it provided the `mathcomp` guys agree to make it standard (i.e.: decide where to put your `duP` lemma.

anton-trunov: Thank you. I'll leave a useful link for the future reference here: when PR #372 is merged we will get that duplicating view (although implemented in a different way) in `ssreflect` -- https://github.com/math-comp/math-comp/pull/372/files#diff-a6f6aa5ebb042f5ccd8936b292342547R172

amahboubi:
@hivert , thanks for contributing! @pi8027 @anton-trunov thanks for your comments! Are we ready to merge?

amahboubi:
If so,  @hivert could you add a line to the change log listing the names of the lemmas added by this PR?

mathcomp/ssreflect/seq.v
```diff
@@ -750,6 +754,44 @@ move=> lt_i_n0 s; case lt_n0_s: (n0 < size s).
...
+Proof.
+elim: i j s => [|i IHi] [|j] [|a s] //; first by rewrite take0 addn0 cats0.
+by rewrite addSn /= IHi.
+Qed.
+
+Lemma takeC i j s : take i (take j s) = take j (take i s).
+Proof.
+wlog i_le_j : i j / i <= j.
+  by move=> Hwlog; case: (leqP i j) => [|/ltnW] /Hwlog ->.
+rewrite take_take // [RHS]take_oversize //; apply: (leq_trans _ i_le_j).
```
CohenCyril: Just a small comment, you either write `apply: leq_trans i_le_j` (no parenthesis and no implicit) or else chain with the previous rewrite by rewriting `(leq_trans _ i_le_j)//` (untested)

mathcomp/ssreflect/seq.v
CohenCyril: please put spaces around the `-` symbol

CohenCyril: I am not sure why you need to generalize `(j - i)`, without running the code, it seems to me that you could just prove the thing with:Proof.
```coq
Proof. ￼move/subnKC<-; rewrite nseqD take_size_cat// size_nseq. Qed.
```

mathcomp/ssreflect/seq.v
```diff
@@ -750,6 +754,44 @@ move=> lt_i_n0 s; case lt_n0_s: (n0 < size s).
...
+Proof.
+move=>/subnKC <-; move: (j-i) => d.
+by rewrite nseqD take_size_cat // size_nseq.
+Qed.
+
+Lemma drop_nseq i j x : drop i (nseq j x) = nseq (j - i) x.
+Proof.
+case: (leqP i j) => [/subnKC {1}<-|/ltnW j_le_i].
+- by rewrite nseqD drop_size_cat // size_nseq.
+- rewrite drop_oversize ?size_nseq //.
```
CohenCyril: The preferred style with two subgoals is no bullet (indent the first subgoal and not the second). Additionally no need to make the detour by `-subn_eq0`, since comparison is convertible to the equation you want to get, so you can just do:
`by rewrite  drop_oversize ?size_nseq// (eqP j_le_i).`
(again not tested, sorry)

mathcomp/ssreflect/seq.v
```diff
@@ -2347,6 +2394,20 @@ Qed.
 Lemma iota_uniq m n : uniq (iota m n).
 Proof. by elim: n m => //= n IHn m; rewrite mem_iota ltnn /=. Qed.
 
+Lemma take_iota k m n : take k (iota m n) = iota m (minn k n).
+Proof.
+rewrite /minn; case: ltnP=> H; last by apply: take_oversize; rewrite size_iota.
+elim: k n H m => [//= | k IHk]; first by case.
```
CohenCyril: you could inline the cases in the elim by doing:
```coq
elim: k n H m => [[]//|k IHk []//= n H m]; rewrite IHk.
```

mathcomp/ssreflect/seq.v
```diff
@@ -2347,6 +2394,20 @@ Qed.
 Lemma iota_uniq m n : uniq (iota m n).
 Proof. by elim: n m => //= n IHn m; rewrite mem_iota ltnn /=. Qed.
 
+Lemma take_iota k m n : take k (iota m n) = iota m (minn k n).
+Proof.
+rewrite /minn; case: ltnP=> H; last by apply: take_oversize; rewrite size_iota.
```
CohenCyril: In mathcomp `H` is reserved for names of groups, hypothesis should bare relevant names (here I guess `lt_something_somthing`)

CohenCyril: also `by apply:` is abbreviated `exact:`

mathcomp/ssreflect/seq.v
```diff
@@ -2347,6 +2394,20 @@ Qed.
...
+elim: k n H m => [//= | k IHk]; first by case.
+by case=> //= n H m; rewrite IHk.
+Qed.
+
+Lemma drop_iota k m n : drop k (iota m n) = iota (m + k) (n - k).
+Proof.
+elim: k m n => [/= | k IHk] m n /=; first by rewrite addn0; case: n.
```
CohenCyril: Since you do a case on n in both cases I guess you could do:
```coq
elim: k m n => [|k IHk] m [|n]//=; rewrite ?addn0// IHk addSn addnS subSS.
```

mathcomp/ssreflect/seq.v
```diff
@@ -750,6 +754,44 @@ move=> lt_i_n0 s; case lt_n0_s: (n0 < size s).
 by rewrite -{1}[s]cats0 take_cat lt_n0_s /= cats0.
 Qed.
 
+Lemma take_take i j : i <= j -> forall s, take i (take j s) = take i s.
+Proof.
+move=> ij s; elim: s i j ij => [// | a s IHs] [|i] [|j] //=.
+by rewrite ltnS => /IHs ->.
```
CohenCyril: This `ltnS` is maybe not necessary.

CohenCyril:
@hivert thanks for contributing, I asked a few more script changes. Although I did them without Coq, I feel pretty confident. If one of my suggestion fails, 1. sorry for the inconvenience, 2. Tell me and I will take the time to download and try my changes.

mathcomp/ssreflect/seq.v
hivert: @CohenCyril True... Fixed the proof

mathcomp/ssreflect/seq.v
hivert: @CohenCyril 

The abbreviation `exact:` for `by apply:` doesn't work if there is something following like in `by apply: take_oversize; rewrite size_iota.` Does it ?

mathcomp/ssreflect/seq.v
hivert: Just `move => /IHs ->` doesn't work...

hivert:
@CohenCyril I closed the change you requested that I manage to apply I commented the two remaining one. Since they are just minor comment I think I've taken care of all your requests. Please resolve the remaining conversations if you agree. I'm now  updating the changelog as asked by @amahboubi and resubmitting my PR. Thanks for the comments and the tips.

CohenCyril:
@hivert thanks for your update. I missed your update of the changelog, I am afraid you have to rebase and solve the conflict in the changelog now, sorry.

hivert:
This is a first time I resolve a conflict with the github interface. I hope this is Ok. 

By the way, do you need me to squash all the change in a unique commit ?

CohenCyril:
@hivert 
> This is a first time I resolve a conflict with the github interface. I hope this is Ok.

As with any git subtle manipulation done through the github interface, the result is not ok. (indeed it created a merge commit inside the branch). Can you please rebase this branch on top of master instead?

Thanks for your patience.

CohenCyril:
Also, yes, please squash.

mathcomp/ssreflect/seq.v
```diff
@@ -742,6 +746,40 @@ move=> lt_i_n0 s; case lt_n0_s: (n0 < size s).
...
+Qed.
+
+Lemma take_drop i j s : take i (drop j s) = drop j (take (i + j) s).
+Proof. by rewrite addnC; elim: s i j => // x s IHs [|i] [|j] /=. Qed.
+
+Lemma take_addn i j s : take (i + j) s = take i s ++ take j (drop i s).
```
CohenCyril: oh and I just realised `take_addn` should rather be named `takeD`

hivert: This is not consistent with `rot_addn`. It should probably be renamed `rotD`, shouldn't it ?

CohenCyril: Mmmh, ok my mistake then... can you revert to the previous state? The change of suffix should be done in one go to answer #359 in order to maintain consistency of the whole library...

mathcomp/ssreflect/seq.v
```diff
@@ -277,6 +277,10 @@ Lemma cat_cons x s1 s2 : (x :: s1) ++ s2 = x :: s1 ++ s2. Proof. by []. Qed.
 Lemma cat_nseq n x s : nseq n x ++ s = ncons n x s.
 Proof. by elim: n => //= n ->. Qed.
 
+Lemma nseqD n1 n2 x :
```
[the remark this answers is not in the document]
hivert: Done

ybertot:
Thanks

Issue #315, opened 2019-03-29
`make` may invoke the same targets more than twice


pi8027:
When one specifies multiple targets (e.g., `ssreflect/all_ssreflect.vo` and `fingroup/all_fingroup.vo`), the common sub-targets of them are invoked more than twice:

```
$ make -v
GNU Make 4.1
Built for x86_64-pc-linux-gnu
...
$ make -j 8 ssreflect/all_ssreflect.vo fingroup/all_fingroup.vo
/usr/local/coq8.9.0/bin/coq_makefile  -f Make -o Makefile.coq
make -f Makefile.coq --no-print-directory ssreflect/all_ssreflect.vo
make -f Makefile.coq --no-print-directory fingroup/all_fingroup.vo
COQDEP VFILES
COQDEP VFILES
COQC ssreflect/ssreflect.v
COQC ssreflect/ssreflect.v
COQC ssreflect/ssrnotations.v
COQC ssreflect/ssrbool.v
COQC ssreflect/ssrbool.v
...
```

This issue also breaks native compilation:

```
$ make COQFLAGS='-native-compiler' -j 8 ssreflect/all_ssreflect.vo fingroup/all_fingroup.vo
/usr/local/coq8.9.0/bin/coq_makefile  -f Make -o Makefile.coq
make -f Makefile.coq --no-print-directory ssreflect/all_ssreflect.vo
make -f Makefile.coq --no-print-directory fingroup/all_fingroup.vo
COQDEP VFILES
COQDEP VFILES
COQC ssreflect/ssreflect.v
COQC ssreflect/ssrbool.v
COQC ssreflect/ssrnotations.v
COQC ssreflect/ssrbool.v
COQC ssreflect/ssreflect.v
COQC ssreflect/ssrnotations.v
File "./ssreflect/.coq-native/Nmathcomp_ssreflect_ssreflect.native", line 1:
Error: I/O error: ./ssreflect/.coq-native/Nmathcomp_ssreflect_ssreflect.native: No such file or directory
File "./ssreflect/ssreflect.v", line 6, characters 0-34:
Warning: Native compiler exited with status 2
[native-compiler-failed,native-compiler]
File "./ssreflect/ssreflect.v", line 6, characters 0-34:
Warning: Removed file ./ssreflect/ssreflect.vo
Error: Could not compile the library to native code.

Makefile.coq:662: recipe for target 'ssreflect/ssreflect.vo' failed
```

gares:
[...]

PR #384, opened 2019-09-26
Fix and improve the test suite and Makefile

pi8027:

This PR consists of the following changes:
- improving an error message produced by the `check_join` tactic,
- fixing the build of the test suite: `make test-suite` (it was broken in #340),
- CI support for `make test-suite`, and
- adding a new rule `only` to build a subset of MathComp (Usually I don't need to build `solvable` and `character`, see #315).

Closes: #315

##### Things done/to do

~Should we add CI support for the test suite? (Currently I don't see how to do it.) cc: @erikmd~ Done.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -8,6 +8,7 @@
 !*.opam
 !plugin
 !mathcomp
+!etc/utils/hierarchy.ml
 
 **/*.d
 **/*.vo
--- a/CHANGELOG_UNRELEASED.md
+++ b/CHANGELOG_UNRELEASED.md
@@ -32,3 +32,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Generalized the `allpairs_catr` lemma to the case where the types of `s`,
   `t1`, and `t2` are non-`eqType`s in `[seq E | i <- s, j <- t1 ++ t2]`.
 
+### Infrastructure
+
+- `Makefile` now supports the `test-suite` and `only` targets. Currently,
+  `make test-suite` will verify the implementation of mathematical structures
+  and their inheritances of MathComp automatically, by using the `hierarchy.ml`
+  utility. One can use the `only` target to build the sub-libraries of MathComp
+  specified by the `TGTS` variable, e.g.,
+  `make only TGTS="ssreflect/all_ssreflect.vo fingroup/all_fingroup.vo"`.
+
--- a/Dockerfile.make
+++ b/Dockerfile.make
@@ -26,4 +26,5 @@ RUN ["/bin/bash", "--login", "-c", "set -x \
   && cd mathcomp \
   && make Makefile.coq \
   && make -f Makefile.coq -j ${NJOBS} all \
-  && make -f Makefile.coq install"]
+  && make -f Makefile.coq install \
+  && make test-suite"]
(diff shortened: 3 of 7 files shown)


erikmd:
Hi @pi8027 
[sorry for late reply]

> Should we add CI support for the test suite? (Currently I don't see how to do it.)

Yes, sounds good:

you just need to: add one (or several) job(s) that trigger the proper `make` command, or extend an existing job (such as **`make-coq-latest:`**, i.e. MathComp with the latest stable release of Coq) to do this.

If you choose that last option, you'd just need to add the following in the end of file `Dockerfile.make` (not tested though, I assume `make test-suite` is the relevant line to add here):

```bash
  …
  && cd mathcomp \
  && make Makefile.coq \
  && make -f Makefile.coq -j ${NJOBS} all \
  && make -f Makefile.coq install \
  && make test-suite"]
```

and the GitLab CI job `make-coq-latest:` will essentially do this:

`docker build -f Dockerfile.make --pull --build-arg=coq_image="coqorg/coq:latest" --build-arg=compiler="edge" -t "mathcomp-dev:make_coq-latest" .`

pi8027:
Hi @erikmd, thank you for the answer. I changed `Dockerfile.make` as you said. Let see how it runs.

pi8027:
```
+ make test-suite
/home/coq/.opam/4.07.1+flambda/bin/coq_makefile  -f Make -o Makefile.coq
make -f Makefile.coq --no-print-directory
make[2]: Nothing to be done for 'real-all'.
COQBIN=/home/coq/.opam/4.07.1+flambda/bin/ ocaml ../etc/utils/hierarchy.ml -verify -R . mathcomp -lib all.all > test_suite/hierarchy_test.v
Cannot find file ../etc/utils/hierarchy.ml.
make: *** [Makefile.common:60: test_suite/hierarchy_test.v] Error 2
```
Hmm... How can I call my utility in `etc` from `make` in this setting? I guess I need to declare a new variable in `Makefile.common` for the path of `hierarchy.ml` then use it to specify that path from `Dockerfile.make`.

erikmd:
Hi, very probably the issue is that the path ../etc/... was ignored by the .dockerignore file (which is mandatory to have so that the so-called Docker build context contains exactly the source code files... but not the .git folder and so on)

erikmd:
The CI is now green, but there is a [warning](https://gitlab.com/math-comp/math-comp/-/jobs/309186276):

*/home/coq/.opam/4.07.1+flambda/bin/coq_makefile  -f Make -o Makefile.coqmake -f Makefile.coq --no-print-directorymake[2]: Nothing to be done for 'real-all'.COQBIN=/home/coq/.opam/4.07.1+flambda/bin/ ocaml ../etc/utils/hierarchy.ml -verify -R . mathcomp -lib all.all > test_suite/hierarchy_test.vcd test_suite; /home/coq/.opam/4.07.1+flambda/bin/coq_makefile  -f Make -o Makefile.coq **Warning: .. (used in -R or -Q) is not a subdirectory of the current directory***

Do you think these lines should be refactored to avoid `..`? or is it a benign issue.
https://github.com/math-comp/math-comp/blob/a4594c56cce42d4f3a110e0958cbecb851076d10/mathcomp/test_suite/Make#L3-L4

pi8027:
@erikmd I moved `test_suite/Make` and `test_suite/Makefile.coq` to `Make.test-suite` and `Makefile.test-suite` respectively to avoid that warning. If I place more than two `Makefile`s generated by the `coq_makefile` utility in the same directory, the name of their dependency file `.coqdep.d` will conflict. But this name can be given from outside like this:
https://github.com/math-comp/math-comp/pull/384/files#diff-9c0cdc05f486cab3d6cace8379825a57R34

Issue #225, opened 2018-09-04
The comments mention a type void that does not exist


xavierm02:
The comments mention a type `void` that does not seem to exist: https://github.com/math-comp/math-comp/search?q=void&unscoped_q=void

amahboubi:
Hi @xavierm02, well spotted. This header is pretty old... May be almost as old as @ggonthier's Four Colour theorem proof?  I guess `void` refers to `False` here. What is your use case?

xavierm02:
I use a type as a flag to (dis)activate a constructor in a inductive:

```
Require Import ssreflect.

Inductive void : Type := .

Definition bool_to_Type b := if b then unit else void.

Section MyType.

  Context {has_B : bool}.

  Inductive my_type :=
    | A : my_type
    | B : bool_to_Type has_B -> my_type
  .
End MyType.

Lemma test1 : forall x : @my_type false, x = A.
  by case.
Qed.

Lemma test2 : exists x y : @my_type true, x <> y.
  exists A.
  exists (@B true tt).
  move => H.
  by inversion H.
Qed.
```

CohenCyril:
Hi, I believe you can do the same by taking `void` to be `False`.

Here is your code with this approach.
To suggest an alternative, I refactored your code a little bit and I also replaced your explicit coercion  `bool_to_Type` by the coercion `is_true` from `bool` to `Prop` and it should be equivalent, since `unit` is equivalent to `true = true`.

```
Require Import ssreflect ssrfun ssrbool.

Inductive my_type (has_B : bool) :=
| A : my_type has_B
| B : has_B -> my_type has_B.
Arguments A {has_B}.
Arguments B {has_B}.

Lemma test1 : forall x : my_type false, x = A.
Proof. by case. Qed.

Lemma test2 : exists x y : my_type true, x <> y.
Proof. by exists A, (B isT) => []. Qed.

```

arthuraa:
FYI I have put up a [repository](https://github.com/arthuraa/coq-void) with instances of basic MathComp classes for `Empty_set`. I might add that to the OPAM archive later, but it would be really useful if those were declared in ssreflect.

amahboubi:
Thanks @arthuraa ! I am adding this to the agenda of the next dev meeting (Oct 16th), so that we provide a feedback by then at the latest.

anton-trunov:
[...]

PR #393, opened 2019-10-19
Instances for empty type.

arthuraa:

Several users have expressed their wish for an empty type in mathcomp (e.g. https://github.com/math-comp/math-comp/issues/225). This PR adds a `void` notation for the `Empty_set` type of the standard library, along with instances of basic classes.

##### Things done/to do

The documentation already existed in `eqtype`. I have added a note in `fintype`, but left `choice` alone since the instance list there was not exhaustive.

The notation is defined in `ssrfun`. Since most of this file lives in the Coq source now, I did not add a corresponding entry there. Would it make sense to do so?


amahboubi:
Hi @arthuraa, thanks for your contribution! Regarding your question:
> The notation is defined in ssrfun. Since most of this file lives in the Coq source now, I did not add a corresponding entry there. Would it make sense to do so?

I think you should add a header to mathcomp's ssrfun, since you introduced the notation here. This is what happens, e.g., in file [ssreflect.v](https://github.com/math-comp/math-comp/blob/5c4328261edace427e7cad73fdca288f6d73b87d/mathcomp/ssreflect/ssreflect.v#L9). I will merge as soon as you add this documentation.

arthuraa:
@amahboubi Sounds good! How does this look?

CohenCyril:
Oh, but if you add something to the local `ssrfun` you should also add it to Coq's... did you? If not, could you do it?

arthuraa:
@CohenCyril Just did: https://github.com/coq/coq/pull/10932.

amahboubi:
@CohenCyril, the PR has been merged in Coq. Anything else you would like @arthuraa to do before we merge?

PR #395, opened 2019-10-24
Progress #270

pi8027:
##### Things done/to do


amahboubi:
@pi8027 Could you explain this PR?

pi8027:
@amahboubi Yes. This PR partially addresses TODOs listed in https://github.com/math-comp/math-comp/pull/270#issuecomment-545343995.

> TODO list on order.v:
> 
> * header:  remove misleading sentences, list the notations and definitions introduced by the library
> * header: improve the documentation of what you get by importing selected of sub-modules

Done by @affeldt-aist. The `Order.Def` is removed and the `Order.Syntax` is exported from `ssreflect.order` so I think we don't need to explain them.

> * header: document better the tagging mechanism allowing to work with different order structures on a same type. This could be in the header and/or in later in the file

Not done yet. This will be addressed by explaining "instances for division on nat".

> * remove the no-op last Import
> * BTW do we need Order.Def?

Done. `Def` modules (`Order.Def`, `Order.POrderDef`, ...) are removed so that definitions should be accessed by using their qualified names (`Order.le`, `Order.lt`, ...).

> * What about importing Order.Syntax in all of the subsequent Order.LTheory, Order.TTheory, etc. Note: a Syntax module is only useful to solve potential grammar conflicts. Which does not seem to be the case here.

Now `Order.Syntax` is exported by default.

> * ensure all notations do have a scope

Done.

> * declare instances for division on nat inside this file, and use this as a documentation/template

That `porderType` instance is provided but not explained yet. I have a problem with proving `left_distributive gcdn lcmn` so there is no `latticeType` instance for this now.

CohenCyril:
@amahboubi 
> @pi8027 Could you explain this PR?

This is a PR on top of https://github.com/math-comp/math-comp/tree/experiment/order to amend it. If @pi8027 had commit rights to mathcomp he would commit directly there. Hence IMO, the only explanation this commit should have is: "Contribution to PR #270: @CohenCyril have look and add it to #270 unless you see something obviously wrong."

mathcomp/ssreflect/order.v
```diff
@@ -5556,4 +5531,39 @@ Module DefaultProdLexiOrder := Order.DefaultProdLexiOrder.
 Module DefaultSeqLexiOrder := Order.DefaultSeqLexiOrder.
 Module DefaultTupleLexiOrder := Order.DefaultTupleLexiOrder.
 
-Import Order.Syntax.
+Module Example_NatDvdOrder.
```
CohenCyril: Can you leave this out? I have a more complete alternative for this already.

pi8027: Sure.

amahboubi:
@pi8027 Many thanks for the detailed explanation, I really appreciate.
@CohenCyril 
> Hence IMO, the only explanation this commit should have is: ...
Whatever level of explanation, it's important for others to have a line of comment describing its role/status. This one had none.

CohenCyril:
> Whatever level of explanation, it's important for others to have a line of comment describing its role/status. This one had none.

Yes, I agreed.

PR #399, opened 2019-10-25
More arithmetic theorems

### LINKED ISSUE #398: leq_subr misnamed?
https://github.com/math-comp/math-comp/blob/cd81418979c9783f9dae65d2aea98742919420e5/mathcomp/ssreflect/ssrnat.v#L520

It looks to me that this lemma should be called `leq_subl` (because the subtraction is on the left). Am I right or am I missing something?

CohenCyril:

Missing arithmetic theorems.
- Generalizing `ltn_subr` and naming it `ltn_subrL`
- Adding `ltn_subrR` and `leq_subrR`
- Changing conclusion of `ltn_predl` to `0 < n` instead of `n != 0` and renaming it `ltn_predL`
- Renaming `ltn_predr` to `ltn_predRL`
- augmenting `CONTRIBUTING.md` with conventions about suffixes `L`, `R`, and also `F`

Also, I did a renaming to fix #398 

##### Things done/to do

### DIFF
--- a/mathcomp/ssreflect/ssrnat.v
+++ b/mathcomp/ssreflect/ssrnat.v
@@ -326,7 +326,7 @@ Hint Resolve leqnSn : core.
 Lemma leq_pred n : n.-1 <= n.           Proof. by case: n => /=. Qed.
 Lemma leqSpred n : n <= n.-1.+1.        Proof. by case: n => /=. Qed.
 
-Lemma ltn_predl n : (n.-1 < n) = (n != 0).
+Lemma ltn_predl n : (n.-1 < n) = (0 < n).
 Proof. by case: n => [//|n]; rewrite ltnSn. Qed.
 
 Lemma ltn_predr m n : (m < n.-1) = (m.+1 < n).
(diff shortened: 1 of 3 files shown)


mathcomp/ssreflect/ssrnat.v
```diff
@@ -520,6 +520,15 @@ Proof. by rewrite -subn_eq0 -subnDA. Qed.
 Lemma leq_subr m n : n - m <= n.
 Proof. by rewrite leq_subLR leq_addl. Qed.
 
+Lemma ltn_subl m n : n < n - m = false.
```
anton-trunov: I think `ltn_subl` should have `F` suffix because of `false`. Maybe something like `ltn_subrF`?

CohenCyril: The `F` suffix is there when there could be an ambiguity... so indeed if we reuse the name `ltn_subr`, yes we should, but I am not convinced...

CohenCyril:
@ybertot this PR is now ready for review

CohenCyril:
@ybertot fixed.
I would like to get @ggonthier's opinion on the renaming though...

ggonthier:
@CohenCyril I have to disagree with the naming convention you use. `l` and `r` are used to refer to the sides of a binary operators; for a relation `L` and `R` are used, though often omitted when the side is unambiguous. For inequalities we also use `ge` of `gt` to indicate the operation is on the small (left-hand) side of a comparison, so a more reasonable renaming would be `leq_subr -> geq_subr`. But given how disruptive renaming with allocation of such a widely used lemma is, I'm fairly reluctant - the new lemma will be nowhere nearly as useful, so it can have a somewhat less natural name.

ggonthier:
A technical point: `<`, `<=`, and `=` are all at level 70, which is in principle non-associative. Though camlp5 does not implement this (equating it to left associative), you should not take advantage of this, and parenthesise the left-hand side of equations of the form `(m < m - n) = false`.

CohenCyril:
@ybertot @ggonthier I took @ggonthier into account and documented https://github.com/math-comp/math-comp/pull/399#issuecomment-554429326 in the CONTRIBUTING.md

ybertot:
@ggonthier Can you confirm you approve the new naming scheme?

ggonthier:
@ybertot @CohenCyril The new naming looks good to me.

ybertot:
Thanks

PR #405, opened 2019-10-31
Fix notation modifiers and scopes (part of #270)

pi8027:

- Notations for explicit application should have `fun_scope`.
- There were so many missing `only parsing` modifiers.

##### Things done/to do

### DIFF
--- a/mathcomp/ssreflect/order.v
+++ b/mathcomp/ssreflect/order.v
@@ -963,34 +963,34 @@ Arguments gt {_ _}.
 
 Module Import POSyntax.
 
-Notation "<=%O" := le : order_scope.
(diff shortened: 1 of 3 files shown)


mathcomp/ssreflect/order.v
```diff
@@ -972,25 +972,25 @@ Notation ">=<%O" := comparable : order_scope.
 Notation "><%O" := (fun x y => ~~ (comparable x y)) : order_scope.
```
CohenCyril: Notations ending with `%O` in order.v and with `%R` in ssrnum.v should be in scope fun_scope as well.

pi8027: Yes. Fixed.

PR #426, opened 2019-11-14
Injectivity lemmas in fintype

pi8027:

Missing theorems.

##### Things done/to do

### DIFF
--- a/mathcomp/ssreflect/fintype.v
+++ b/mathcomp/ssreflect/fintype.v
@@ -1831,6 +1831,14 @@ move=> Ax0; exists (enum_rank_in Ax0) => [i _|]; last exact: enum_rankK_in.
 exact: enum_valK_in.
 Qed.
 
+Lemma eq_enum_rank_in (x0 y0 : T) A (Ax0 : x0 \in A) (Ay0 : y0 \in A) :
+  {in A, enum_rank_in Ax0 =1 enum_rank_in Ay0}.
+Proof. by move=> x xA; apply: enum_val_inj; rewrite !enum_rankK_in. Qed.
+
+Lemma enum_rank_in_inj (x0 y0 : T) A (Ax0 : x0 \in A) (Ay0 : y0 \in A) :
+  {in A &, forall x y, enum_rank_in Ax0 x = enum_rank_in Ay0 y -> x = y}.
+Proof. by move=> x y xA yA /(congr1 enum_val); rewrite !enum_rankK_in. Qed.
+
 Lemma enum_rank_bij : bijective enum_rank.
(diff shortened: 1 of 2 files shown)


mathcomp/ssreflect/fintype.v
```diff
@@ -1816,6 +1816,10 @@ move=> x; apply: ord_inj; rewrite insubdK; last first.
 by rewrite index_uniq ?enum_uniq // -cardE.
 Qed.
 
+Lemma enum_rank_in_inj x y A (Ax : x \in A) (Ay : y \in A) :
```
CohenCyril: I believe this is the combination of can_inj_in and enum_rankK_in so no need for a dedicated theorem...

pi8027: `can_in_inj (enum_rankK_in _)` requires that `Ax` and `Ay` are the same. This `enum_rank_in_inj` is slightly general.

CohenCyril: Well if you really want to be as general as possible you should write instead:
```coq
Lemma enum_rank_in_inj (x0 y0 : T) A (Ax0 : x0 \in A) (Ay0 : y0 \in A) :
  {in A &, forall x y, enum_rank_in Ax0 x = enum_rank_in Ay0 y -> x = y}.
Proof. by move=> x y xA yA /(congr1 enum_val); rewrite !enum_rankK_in. Qed.
```

CohenCyril: (Otherwise you restrict witnesses to be equal to the arguments, and it is unlikely to happen often enough...)

CohenCyril: But then I guess it is better to provide 
```coq
Lemma eq_enum_rank_in (x0 y0 : T) A (Ax0 : x0 \in A) (Ay0 : y0 \in A) :
  {in A, enum_rank_in Ax0 =1 enum_rank_in Ay0}.
Proof. by move=> x xA; apply: enum_val_inj; rewrite !enum_rankK_in. Qed.
```
instead...

CohenCyril: (to give more fine-grained control over the use of different witnesses)

pi8027: That's so true. Since `eq_enum_rank_in` doesn't match with my use case well, I'd like to put both of those.

pi8027:
(Since my use case of this PR also requires #270 and #388, I'm totally fine to target 1.11.0 here.)

amahboubi:
@ybertot should we merge this one now or after releasing?

ybertot:
As I see, this PR is already approved, so it can be integrated (same treatment as #432).

PR #431, opened 2019-11-17
Remove hint declarations using non-global definitions.

ppedrot:

See coq/coq#7791.

##### Things done/to do

### DIFF
--- a/mathcomp/field/algnum.v
+++ b/mathcomp/field/algnum.v
@@ -53,7 +53,8 @@ Local Notation pZtoQ := (map_poly ZtoQ).
 Local Notation pZtoC := (map_poly ZtoC).
 Local Notation pQtoC := (map_poly ratr).
 
-Local Hint Resolve (intr_inj : injective ZtoC) : core.
+Local Definition intr_inj_ZtoC := (intr_inj : injective ZtoC).
+Local Hint Resolve intr_inj_ZtoC : core.
 Local Notation QtoCm := [rmorphism of QtoC].
 
 (* Number fields and rational spans. *)
--- a/mathcomp/algebra/mxpoly.v
+++ b/mathcomp/algebra/mxpoly.v
@@ -901,7 +901,10 @@ apply: integral_horner_root mon_p pu0 intRp _.
 by apply/integral_poly => i; rewrite coefX; apply: integral_nat.
 Qed.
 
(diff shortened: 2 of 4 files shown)


mathcomp/algebra/ssrnum.v
```diff
@@ -3769,7 +3770,8 @@ Section RealDomainOperations.
 
 Variable R : realDomainType.
 Implicit Types x y z t : R.
-Hint Resolve (@num_real R) : core.
+Let num_real_R := @num_real R.
+Hint Resolve num_real_R : core.
```
gares: In the rest of the PR you consistently use `Let` followed by `Local Hint` which seems fine. In this file, last two hunks, you use `Let` followed by `Hint`. Is it a problem?

mathcomp/field/algnum.v
```diff
@@ -53,7 +53,8 @@ Local Notation pZtoQ := (map_poly ZtoQ).
 Local Notation pZtoC := (map_poly ZtoC).
 Local Notation pQtoC := (map_poly ratr).
 
-Local Hint Resolve (intr_inj : injective ZtoC) : core.
+Local Definition intr_inj_ZtoC := (intr_inj : injective ZtoC).
+Local Hint Resolve intr_inj_ZtoC : core.
```
gares: For example here the hint is `Local` but `intr_inj_ZtoC` stays (even if its name is always qualified).

mathcomp/algebra/ssrnum.v
ppedrot: Doesn't matter, in a section hints are implicitly local.

CohenCyril:
I am a bit nervous about this ... `Let` declarations pollute the goal display, and in this case for theorems that are `Hint Resolve`, hence meant to be transparent to the user. Is there a new "invisible `Let`" feature that I am not aware of?

ppedrot:
@CohenCyril Alternatively we can define them inside the section if you prefer. Nonetheless, it seems that let-definitions are used pervasively in the mathcomp library.

mathcomp/algebra/ssrnum.v
gares: ok

affeldt-aist:
It seems to the release managers that this PR could make it to the upcoming release. Can you complete it?

affeldt-aist:
We are thinking about merging this PR by tomorrow if nobody disagrees.

PR #433, opened 2019-11-18
[test suite] infrastructure to test how some statements are printed

gares:

Detect Coq changes when printing goals

##### Things done/to do

- [x] ~added corresponding entries in `CHANGELOG_UNRELEASED.md`~
- [x] call `make test-suite` in Coq's CI https://github.com/coq/coq/pull/13014 (merged)


gares:
The reference file is generated with Coq 8.10, indeed some coercions are displayed, CFR: https://github.com/coq/coq/issues/11033

gares:
For the reviewers:
- please suggest other lemmas to be printed and scopes to be opened

gares:
CI should test if coq/coq#11090 (that I used as the reference output) agrees with Coq 8.9.
Coq 8.10 prints coercions, so there it will fail.

mathcomp/Makefile.coq.local
```diff
@@ -0,0 +1,17 @@
...
+	  if [ $$? -ne 0 ]; \
+	    then echo "Output of file $$f differs!";\
+		diff -u $$f.out $$f.out.new;\
+		exit 1;\
+	  fi;\
+	done
+
+$(OUTPUT_ARTIFACTS): %.v.out.new: %.v all/all.vo
+	$(COQC) $(COQFLAGS) $(COQLIBS) $< > $<.out.new
+
```
pi8027: This should be moved to `Makefile.test-suite.local`. Currently, the use of `coq_makefile` is split into two parts: the library and test-suite, because generating `test_suite/hierarchy_test.v` requires `all/all.vo`.

pi8027:
Also `Make.test-suite` have to be changed. Currently, the test-suite facility is invoked only from `make-coq-latest` and not invoked from `coq-8.9` and `coq-8.10` in CI. (see `Dockerfile.make` and `.gitlab-ci.yml`.)

mathcomp/Makefile.coq.local
ybertot: @gares do you confirm that this PR should really be in release 1.11?

ybertot:
@gares, do you confirm that this PR should be in release 1.11

gares:
We don't really care since it is not for the final user, it is for us and for the coq devs.

But I completely forgot about this PR, thanks for the heads up

CohenCyril:
@gares I think this PR is of extreme importance to prevent Coq regressions going under the radar.
@maximedenes can I assign you to torture @gares?

maximedenes:
> can I assign you to "torture" @gares?

Sure, sounds like something I could do ;)

ejgallego:
Note that if #316 is merged, dune does provide some built-in support for this

gares:
I've revived this. It would be nice to have a second round of review on CI & Docker since I'm really not an expert there.

CohenCyril:
@erikmd would you mind taking a look?

erikmd:
@CohenCyril yes, I plan to have a look today (otherwise ping me!)

gares:
It requires some fixes. In particular you can provide a different output reference depending on the version of Coq (plus a default one). I must have messed things up, so the reference for 8.12 is OK, the default one is wrong (my intent was to have it for < 8.12) and the 8.13 reference is wrong as well. Anyway, fixing the reference files should be orthogonal to the infrastructure.

.gitlab-ci.yml
```diff
@@ -70,6 +70,7 @@ make-coq-latest:
   script:
     - docker build --pull --build-arg=coq_image="coqorg/${CI_JOB_NAME//-/:}" --build-arg=compiler="${OPAM_SWITCH}" -t "${IMAGE}" .
     - docker push "${IMAGE}"
+    - docker run --mount "type=bind,source=$(pwd),target=/home/coq/mathcomp" -w /home/coq/mathcomp -t "${IMAGE}" /bin/bash --login -c "set -x; cd mathcomp; make test-suite"
```
erikmd, line 73: This command works but the place where it is run is suboptimal. More precisely:
It runs `make test-suite` (existing tests + notation test) for each version of Coq, which is very fine, but these tests should ideally be separated from the build job to "separate concerns", so that:

* the other stages don't wait too much to be started (otherwise it would take [42'](https://gitlab.com/math-comp/math-comp/-/jobs/729111262) while it used to take [26'](https://gitlab.com/math-comp/math-comp/-/jobs/730243138)),
* and a failure in this new notation test won't prevent other tests to be run.

However, the change I suggest (cf. my main review post) should not prevent running these tests in the nightly build, so that we can be aware of any unexpected notation change from coq.dev.

mathcomp/Makefile.test-suite.coq.local
```diff
@@ -0,0 +1,22 @@
...
+	  else REFERENCE=$$f.out; fi;\
+	  diff -u $$REFERENCE $$f.out.new > /dev/null;\
+	  if [ $$? -ne 0 ]; \
+	    then echo "Output of file $$f differs!";\
+		diff -u $$REFERENCE $$f.out.new;\
+		exit 1;\
+	  fi;\
+	done
+
+$(OUTPUT_ARTIFACTS): %.v.out.new: %.v all/all.vo
```
erikmd, line 21: it seems [from the log](https://gitlab.com/math-comp/math-comp/-/jobs/729111262) that mathcomp is needlessly recompiled twice.

I guess you should just remove the `all/all.vo` prerequisite, so that the `test_suite/output.v` compilation will rely on the existing build of mathcomp (installed by the CI job using opam)

```suggestion
$(OUTPUT_ARTIFACTS): %.v.out.new: %.v
```

erikmd (commented):
Hi @gares, the feature brought by this PR is very nice!

However as briefly mentioned in my inline comments, the `docker run … ${IMAGE}` command could be replaced with some dedicated jobs, in order to run the tests independently of the build/deploy job. An advantage would also be that you chould choose which versions of Coq to test with `make test-suite`.

Below is a suggested snippet that you may want to try (untested for now). Finally, there are a few comments I left in the Makefile code.

```yaml
################
##### test stage
################

# run "make test-suite" (required variable: COQ_VERSION)
.test:
  stage: test
  image: "${CI_REGISTRY_IMAGE}:${CI_PIPELINE_IID}_${CI_COMMIT_REF_SLUG}_coq-${COQ_VERSION}"
  before_script:
    - cat /proc/{cpu,mem}info || true
    # don't printenv to avoid cluttering the log
    - opam config list
    - opam repo list
    - opam list
    - coqc --version
    - echo "${COQ_VERSION}"
    - git rev-parse --verify HEAD
    - git describe --all --long --abbrev=40 --always --dirty
    - pwd
    - cd mathcomp
    - make test-suite
  except:
    - tags
    - merge_requests

# run "make test-suite" only for push pipelines (not for scheduled pipelines)
.test-once:
  extends: .test
  except:
    - tags
    - merge_requests
    - schedules

test-coq-dev:
  extends: .test
  variables:
    COQ_VERSION: "dev"

test-coq-8.13:
  # to be replaced with .test-once (to disable this nightly build) when 8.13.0 is released
  extends: .test
  variables:
    COQ_VERSION: "8.13"

test-coq-8.12:
  extends: .test-once
  variables:
    COQ_VERSION: "8.12"

# set CONTRIB_URL, script, COQ_VERSION, CONTRIB_VERSION when using
.ci:
# etc.
```

erikmd:
BTW, a related question:
are you interested in keeping the generated files `test_suite/output.v.out.new` as build artifacts? (e.g., with an expiration of 1 month)

gares:
the diff in the log is enough I guess, but why not.

erikmd:
@gares

> the diff in the log is enough I guess, but why not.

OK so maybe it is unnecessary (otherwise you might want to add an `artifacts` spec after these `script` lines below)

```yaml
  script:
    - cd mathcomp
    - make test-suite 
  artifacts:
    name: "output-test_$CI_JOB_NAME"
    when: always
    paths:
      - mathcomp/test_suite/*.v.out.new
    expire_in: 1 month
```

Anyway, don't forget to remove the line
```
- docker run --mount "type=bind,source=$(pwd),target=/home/coq/mathcomp" -w /home/coq/mathcomp -t "${IMAGE}" /bin/bash --login -c "set -x; cd mathcomp; make test-suite"
```
(which is subsumed by the `test-*` jobs)

erikmd:
It seems there are two remaining issues:

1. `make test-suite` rebuilds the whole mathcomp (instead of relying on the installed `.vo`s): https://gitlab.com/math-comp/math-comp/-/jobs/732817696#L188
2. for some reason, the `diff` is performed with `test_suite/output.v.out` instead of `test_suite/output.v.out.8.9` (?): https://gitlab.com/math-comp/math-comp/-/jobs/732817696#L327

mathcomp/Makefile.test-suite.coq.local
```diff
@@ -0,0 +1,20 @@
...
+
+OUTPUT_ARTIFACTS=$(OUTPUT_TESTS:%.v=%.v.out.new)
+
+COQ_VERSION=$(shell $(COQC) -print-version | cut -d ' ' -f 1 | cut -d '.' -f 1-2)
+
+# Given a file f we compare its compilation output f.out.new with
+# f.out.COQ_VERSION (or f.out if f.out.COQ_VERSION does not exist)
+post-all:: $(OUTPUT_ARTIFACTS)
+	@for f in $(OUTPUT_TESTS); do\
+	  if [ -e $$f.out.$(COQ_VERSION) ]; then REFERENCE=$$f.out.$(COQ_VERSION);\
```
erikmd, line 11: @gares 
I've just found the origin of issue 2. mentioned in https://github.com/math-comp/math-comp/pull/433#issuecomment-691033191 (variable name clash)

```suggestion
COQ_VERSION_MINOR=$(shell $(COQC) -print-version | cut -d ' ' -f 1 | cut -d '.' -f 1-2)

# Given a file f we compare its compilation output f.out.new with
# f.out.COQ_VERSION_MINOR (or f.out if f.out.COQ_VERSION_MINOR does not exist)
post-all:: $(OUTPUT_ARTIFACTS)
	@for f in $(OUTPUT_TESTS); do\
	  if [ -e $$f.out.$(COQ_VERSION_MINOR) ]; then REFERENCE=$$f.out.$(COQ_VERSION_MINOR);\
```

gares:
For 1. I miss there the dependency is. Do you know? I'm fine with the env variable thing, but I could not find where to use it.

gares:
I think I found it. Let's see how CI goes

erikmd:
Hi @gares,

the pipeline failed (so I cancelled to avoid running the remaining jobs needlessly)
because IIUC, the `all.vo` file is not installed by opam install.

cf. https://gitlab.com/math-comp/math-comp/-/jobs/734043221
```
$ make test-suite TEST_SKIP_BUILD=1
COQBIN=/home/coq/.opam/4.07.1+flambda/bin/ ocaml ../etc/utils/hierarchy.ml -verify -R . mathcomp -lib all.all > test_suite/test_hierarchy_all.v
/home/coq/.opam/4.07.1+flambda/bin/coq_makefile  -f Make.test-suite -o Makefile.test-suite.coq
make -f Makefile.test-suite.coq VDFILE=".coqdeps.test-suite" --no-print-directory
COQDEP VFILES
make[2]: *** No rule to make target 'all/all.vo', needed by 'test_suite/test_hierarchy_all.vo'.  Stop.
make[1]: *** [Makefile.test-suite.coq:339: all] Error 2
make: *** [Makefile.common:96: this-test-suite] Error 2
ERROR: Job failed: exit code 1
```

Hopefully just reverting this line https://github.com/math-comp/math-comp/commit/ff61778018f5eff362a50e48e5b3fd30dec962d2
(and keeping your other change dealing with `TEST_SKIP_BUILD=1` would be OK, to compile `all.vo` and *just* `all.vo`?)

pi8027:
I think it is acceptable to replace this `-lib all.all` with `-lib all_ssreflect -lib all_algebra ... -lib all_character` to avoid the issue with `all.vo`.
https://github.com/math-comp/math-comp/blob/c2e7bad5c95a11f42f9f6d282ee2e5d84e27cbed/mathcomp/Makefile.common#L60

gares:
I guess I need to remove the -R as well

pi8027:
> I guess I need to remove the -R as well

@gares Removing `-R . mathcomp` breaks the `test-suite` target (without installation). Should we rather pass `COQLIBS` variable of `Makefile.test-suite.coq` to `hierarchy.ml` and use it to locate libraries?

gares:
I've the impression that the Make.test-suite is not designed work in both local and system wide modes.

What one could do is to have the file contain no -R and, depending on the mode (a variable), invoke `coq_makefile -f Make.test-suite -R . mathcomp` or `coq_makefile -f Make.test-suite -R test_suite mathcomp.test_suite`.

Is what you propose simpler/cleaner?

gares:
This looks finally ready for squash+merge!

erikmd:
:+1:
and the night after the squash+merge, we'll just need to check that the new job `test-coq-dev` is properly triggered in the scheduled [pipeline](https://gitlab.com/math-comp/math-comp/-/pipelines).

CohenCyril:
Before merge I have few questions : how many tests should we perform? Would it be possible to test intermediate statements? (i.e. after application of a few tactics).

gares:
> Before merge I have few questions : how many tests should we perform? Would it be possible to test intermediate statements? (i.e. after application of a few tactics).

How may I don't know, the only limit is about maintenance, if the statement changes or Coq changes (while staying correct) then the tests may need an update.

Yes, just start a proof and write `Show.` when you feel like the goal is interesting.

gares:
As a start it would help to have some tests covering rings, polynomials, lattices and matrices for example, since the statement I picked covers many group stuff, some set and seq stuff as well, but not much more. But not in this PR ;-)

gares:
ping @maximedenes

pi8027 (approved):
At least, `Make.test-suite` and `Makefile.common` seem OK.

erikmd (approved):
The gitlab-ci part of the PR is ok for me as well.

CohenCyril:
OK! I'm taking the lead and merging.

PR #434, opened 2019-11-20
New generalised induction idiom

### LINKED ISSUE #422: Strong induction for nat
Just a reminder that @ggonthier expects to include a new idiom for strong induction in the library, to work for the upcoming release.

ggonthier:
Replace the legacy generalised induction idiom with a more robust one
that does not rely on the `{-2}` numerical occurrence selector, using
either new helper lemmas `ubnP`, `ltnSE`, `ubnPleq`, `ubnPgeq`, `ubnPeq`, or a specific `nat`
induction principle `ltn_ind`.
   This change anticipates an expected evolution of `fintype` to integrate `finmap`. It is likely that the new definition of the `#|A|` notation will hide multiple occurrences of `A`, which will break the `{-2}` induction idiom. Client libraries should update before the 1.11 release.

Fixes #422

### DIFF
--- a/CHANGELOG_UNRELEASED.md
+++ b/CHANGELOG_UNRELEASED.md
@@ -13,6 +13,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Added a `void` notation for the `Empty_set` type of the standard library, the
   canonical injection `of_void` and its cancellation lemma `of_voidK`, and
   `eq`, `choice`, `count` and `fin` instances.
+  
+- Added `ltn_ind` general induction principle for `nat` variables, helper lemmas `ubnP`, `ltnSE`, ubnPleq, ubnPgeq and ubnPeq, in support of a generalized induction idiom for `nat` measures that does not rely on the `{-2}` numerical occurrence selector, and purged this idiom from the `mathcomp` library (see below).
 
 - Added fixpoint and cofixpoint constructions to `finset`: `fixset`,
   `cofixset` and `fix_order`, with a few theorems about them
@@ -84,6 +86,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Changed
 
+- Replaced the legacy generalised induction idiom with a more robust one
+that does not rely on the `{-2}` numerical occurrence selector, using
+new `ssrnat` helper lemmas `ltn_ind`, `ubnP`, `ubnPleq`,  ...., (see above). The new idiom is documented in `ssrnat`.
+   This change anticipates an expected evolution of `fintype` to integrate `finmap`. It is likely that the new definition of the `#|A|` notation will hide multiple occurrences of `A`, which will break the `{-2}` induction idiom. Client libraries should update before the 1.11 release.
+   
 - `eqVneq` lemma is changed from `{x = y} + {x != y}` to
   `eq_xor_neq x y (y == x) (x == y)`, on which a case analysis performs
   simultaneous replacement of expressions of the form `x == y` and `y == x`
--- a/mathcomp/fingroup/perm.v
+++ b/mathcomp/fingroup/perm.v
@@ -465,18 +465,18 @@ Arguments dpair {eT}.
 Lemma prod_tpermP s :
   {ts : seq (T * T) | s = \prod_(t <- ts) tperm t.1 t.2 & all dpair ts}.
 Proof.
-elim: {s}_.+1 {-2}s (ltnSn #|[pred x | s x != x]|) => // n IHn s.
-rewrite ltnS => le_s_n; case: (pickP (fun x => s x != x)) => [x s_x | s_id].
-  have [|ts def_s ne_ts] := IHn (tperm x (s^-1 x) * s).
-    rewrite (cardD1 x) !inE s_x in le_s_n; apply: leq_ltn_trans le_s_n.
-    apply: subset_leq_card; apply/subsetP=> y.
-    rewrite !inE permM permE /= -(canF_eq (permK _)).
-    have [-> | ne_yx] := altP (y =P x); first by rewrite permKV eqxx.
-    by case: (s y =P x) => // -> _; rewrite eq_sym.
+have [n] := ubnP #|[pred x | s x != x]|; elim: n s => // n IHn s /ltnSE-le_s_n.
+case: (pickP (fun x => s x != x)) => [x s_x | s_id]; last first.
+  exists nil; rewrite // big_nil; apply/permP=> x.
+  by apply/eqP/idPn; rewrite perm1 s_id.
+have [|ts def_s ne_ts] := IHn (tperm x (s^-1 x) * s); last first.
   exists ((x, s^-1 x) :: ts); last by rewrite /= -(canF_eq (permK _)) s_x.
   by rewrite big_cons -def_s mulgA tperm2 mul1g.
-exists nil; rewrite // big_nil; apply/permP=> x.
-by apply/eqP/idPn; rewrite perm1 s_id.
+rewrite (cardD1 x) !inE s_x in le_s_n; apply: leq_ltn_trans le_s_n.
+apply: subset_leq_card; apply/subsetP=> y.
+rewrite !inE permM permE /= -(canF_eq (permK _)).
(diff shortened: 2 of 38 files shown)


mathcomp/fingroup/perm.v
```diff
@@ -465,18 +465,18 @@ Arguments dpair {eT}.
...
-    rewrite (cardD1 x) !inE s_x in le_s_n; apply: leq_ltn_trans le_s_n.
-    apply: subset_leq_card; apply/subsetP=> y.
-    rewrite !inE permM permE /= -(canF_eq (permK _)).
-    have [-> | ne_yx] := altP (y =P x); first by rewrite permKV eqxx.
-    by case: (s y =P x) => // -> _; rewrite eq_sym.
+have [n] := ubnP #|[pred x | s x != x]|; elim: n s => // n IHn s /ltnSE-le_s_n.
+case: (pickP (fun x => s x != x)) => [x s_x | s_id]; last first.
+  exists nil; rewrite // big_nil; apply/permP=> x.
+  by apply/eqP/idPn; rewrite perm1 s_id.
+have [|ts def_s ne_ts] := IHn (tperm x (s^-1 x) * s); last first.
```
amahboubi: This idiom seems to indicate that allowing `suff <intro-pat> := ...` would be useful.

CHANGELOG_UNRELEASED.md
```diff
@@ -13,6 +13,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Added a `void` notation for the `Empty_set` type of the standard library, the
   canonical injection `of_void` and its cancellation lemma `of_voidK`, and
   `eq`, `choice`, `count` and `fin` instances.
+  
+- Added `ltn_ind` general induction principle for `nat` variables, helper lemmas `ubnP` and `ltnSE` in support of a generalized induction idiom for `nat` mesures that does not rely on the `{-2}` numerical occurrence selector, and purged this idiom from the `mathcomp` library. The `{-2}` idiom is likely to fail in upcoming releases for cardinal (i.e., `#|A|`) measures as the new `#|A|` notation may hide multiple occurrences of `A`.
```
ybertot: mesures --> measures

mathcomp/solvable/abelian.v
```diff
@@ -1656,10 +1658,10 @@ Theorem abelian_structure G :
     abelian G ->
   {b | \big[dprod/1]_(x <- b) <[x]> = G & map order b = abelian_type G}.
 Proof.
-rewrite /abelian_type genGidG.
-elim: {G}#|G| {-2 5}G (leqnn #|G|) => /= [|n IHn] G leGn cGG.
-  by rewrite leqNgt cardG_gt0 in leGn.
-rewrite {1}cGG /=; case: ifP => [ntG|/eqP->]; last first.
+rewrite /abelian_type genGidG; set n := #|G|.
+have: #|G| <= n by []; clearbody n.
```
amahboubi: Why is this `clearbody` needed? Also, I though that the idiom would by `by clearbody n`, with `by []` used only before a dot.

ggonthier: The `clearbody n` is executed _after_ the `have`, whose proof relies on `n` being convertible to `#|G|`. It is needed because the following line needs to abstract `n` _after_ `G`; without it Coq would complain that "`G` occurs in `n`". Alternatively I could have rewritten that line
```
elim: {G}n (G) => // n G ...
```
but I felt the `clearbody` was more self-contained.
I also considered adding a helper lemma in `ssrnat` to support this idiom:
```
Variant ubn_leq_spec m : nat -> Type := UbnLeq n of m <= n : ubn_leq_spec m n.
Lemma ubnPleq m : ubn_leq_spec m m. Proof. by []. Qed.
```
alongside a similar one for `==`, but in the end decided against this because there were relatively few instances. I can restore them by popular demand.

mathcomp/fingroup/perm.v
ggonthier: It would be a somewhat awkward tactic, because `have ... := ...` does not usually generate two subgoals. It only does so here because the statement of `IHn s` has the form `premise -> inductive`. This it is very much a special case; if there were more premises more subgoals would be generated.

CHANGELOG_UNRELEASED.md
ggonthier: Well spotted; should I also include the more direct call for clients to update as well (which I added to the PR description), perhaps splitting the paragraph between the `Added` and `Changed` sections?

mathcomp/solvable/abelian.v
CohenCyril: I prefer a helper lemma.
However the `have [] := ubnPleq x` you suggest abstracts all occurences of `x` in the goal while `have [] := ubnP x` does not, so the apparent similarity in naming could be misleading...

ggonthier: There's nothing to abstract to in `ubnP M`, because the bound produced is `M.+1`. To be precise, `have [n] := ubnP M` could only replace `M` with `n.-1`, which is fairly unnatural behaviour likely to yield unusable induction hypotheses.

CohenCyril: OK then I am all in for your `ubnPleq`.

CHANGELOG_UNRELEASED.md
ybertot: I agree adding the direct call for clients to update is a good use of the CHANGELOG file.

mathcomp/solvable/abelian.v
CohenCyril: Alternatively if you want to factor the cases for `<=` and `==`, we could have
```coq
Variant ub_spec T x (r : rel T) : T -> Type := UbRefl y of r x y : ub_spec x r y.
Lemma ubP T (r : rel T) x : r x x -> ub_spec x r x. Proof. by []. Qed.

Lemma test P (m : nat) : P m.
have [n] := ubP (leqnn m).
Restart.
have [n] := ubP (eqxx m).
Abort.
```

ggonthier: I believe it would be counter-productive to have to state the proof of the (in)equality. The name `ubP` wouldn't be appropriate, because it is actually a helper lemma for selectively abstracting the argument of a bool predicate, which doesn't cover that many use cases. I'd rather use the tactic language and its context selectors in such cases. To give an example, `upP` couldn't be used to generalise `G : {group gT}` as an `H \subset G`, because `G` actually occurs as `(mem (gval G))` in the type `G \subset G` of `subxx G G`.

CohenCyril: Oh! right...

mathcomp/fingroup/perm.v
amahboubi: Oh sure, I see now that my suggestion is pointless. I guess I would have created a separate, preliminary,  `have` for the proof of the premise in this case.I'm resolving the conversation anyway.

mathcomp/solvable/abelian.v
amahboubi: @CohenCyril is this conversation resolved?

CHANGELOG_UNRELEASED.md
amahboubi: @ggonthier I propose to merge as soon as you implement @ybertot 's suggestion.

mathcomp/solvable/abelian.v
ggonthier: I've added the new helper lemmas, including one for the `geq` case that is useful in a couple of instances, for remembering that the inductive index remains below its final value.

mathcomp/fingroup/perm.v
ggonthier: That last remark is actually a good stylistic improvement, which I've implemented.

CHANGELOG_UNRELEASED.md
ggonthier: Done (and squashed), along with the inclusion of the additional helpers that arose from the conversation with @CohenCyril.

gares:
I arrive a bit late, but my suggestion can be applied a posteriori.

I believe the *diff* of this PR is very relevant for the user willing to port his files. It may be nice to have a pointer to this PR or this commit in the changelog, eg "This commit ports math-comp to the new idiom: <link>".

PR #441, opened 2019-11-26
Big enum

ggonthier:
Added lemmas `big_enum_cond`, `big_enum` and `big_enumP` to handle more
explicitly big ops iterating over explicit enumerations in a `finType`.
The previous practice was to rely on the convertibility between
`enum A` and `filter A (index_enum T)`, sometimes explicitly via the
`filter_index_enum` equality, more often than not implicitly.
Both are likely to fail after the integration of `finmap`, as the
`choiceType` theory can’t guarantee that the order in selected
enumerations is consistent.
For this reason `big_enum` and the related (but currently unused)
`big_image` lemmas are restricted to the abelian case. The `big_enumP`
lemma can be used to handle enumerations in the non-abelian case, as
explained in the `bigop.v` internal documentation.
The Changelog entry enjoins clients to stop relying on either
`filter_index_enum` and convertibility (though this PR still provides
both), and warns about the restriction of the `big_image` lemma set to
the abelian case, as it it a possible source of incompatibility.

### DIFF
--- a/CHANGELOG_UNRELEASED.md
+++ b/CHANGELOG_UNRELEASED.md
@@ -36,9 +36,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   `eq_enum_rank_in`, `enum_rank_in_inj`, `lshift_inj`, and
   `rshift_inj`.
 
-- Bigop theorems: `big_rmcond`, `bigD1_seq`,
+- Bigop theorems: `index_enum_uniq`, `big_rmcond`, `bigD1_seq`,
   `big_enum_val_cond`, `big_enum_rank_cond`,
-  `big_enum_val`, `big_enum_rank`, `big_set`.
+  `big_enum_val`, `big_enum_rank`, `big_set`,
+  `big_enumP`, `big_enum_cond`, `big_enum`
 
 - Arithmetic theorems in ssrnat and div:
   - some trivial results in ssrnat: `ltn_predL`, `ltn_predRL`,
@@ -101,7 +102,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Replaced the legacy generalised induction idiom with a more robust one
 that does not rely on the `{-2}` numerical occurrence selector, using
 new `ssrnat` helper lemmas `ltn_ind`, `ubnP`, `ubnPleq`,  ...., (see above). The new idiom is documented in `ssrnat`.
-   This change anticipates an expected evolution of `fintype` to integrate `finmap`. It is likely that the new definition of the `#|A|` notation will hide multiple occurrences of `A`, which will break the `{-2}` induction idiom. Client libraries should update before the 1.11 release.
+   This change anticipates an expected evolution of `fintype` to integrate `finmap`. It is likely that the new definition of the `#|A|` notation will hide multiple occurrences of `A`, which will break the `{-2}` induction idiom. Client libraries should update before the 1.11 release (see [PR #434](https://github.com/math-comp/math-comp/pull/434) for examples).
+   
+ - Replaced the use of the accidental convertibility between `enum A` and 
+   `filter A (index_enum T)` with more explicit lemmas `big_enumP`, `big_enum`, `big_enum_cond`, `big_image` added to the `bigop` library, and deprecated the `filter_index_enum` lemma that states the corresponding equality. Both convertibility and equality may no longer hold in future `mathcomp` releases when sets over `finType`s are generalised to finite sets over `choiceType`s, so client libraries should stop relying on this identity. File `bigop.v` has some boilerplate to help with the port; also see [PR #441](https://github.com/math-comp/math-comp/pull/441) for examples.
+   
+ - Restricted `big_image`, `big_image_cond`, `big_image_id` and `big_image_cond_id`
+ to `bigop`s over _abelian_ monoids, anticipating the change in the definition of `enum`. This may introduce some incompatibilities - non-abelian instances should be dealt with a combination of `big_map` and `big_enumP`.
    
 - `eqVneq` lemma is changed from `{x = y} + {x != y}` to
   `eq_xor_neq x y (y == x) (x == y)`, on which a case analysis performs
--- a/mathcomp/ssreflect/bigop.v
+++ b/mathcomp/ssreflect/bigop.v
@@ -545,20 +545,35 @@ Canonical bigop_unlock := Unlockable BigOp.bigopE.
 
 Definition index_iota m n := iota m (n - m).
 
-Definition index_enum (T : finType) := Finite.enum T.
-
 Lemma mem_index_iota m n i : i \in index_iota m n = (m <= i < n).
 Proof.
 rewrite mem_iota; case le_m_i: (m <= i) => //=.
 by rewrite -leq_subLR subSn // -subn_gt0 -subnDA subnKC // subn_gt0.
 Qed.
 
+(* Legacy mathcomp scripts have been relying on the fact that enum A and      *)
+(* filter A (index_enum T) are convertible. This is likely to change in the   *)
+(* next mathcomp release when enum, pick, subset and card are generalised to  *)
+(* predicates with finite support in a choiceType - in fact the two will only *)
+(* be equal up to permutation in this new theory.                             *)
+(*  It is therefore advisable to stop relying on this, and use the new        *)
+(* facilities provided in this library: lemmas big_enumP, big_enum, big_image *)
+(* and such. Users wishing to test compliance should change the Defined in    *)
+(* index_enum_key to Qed, and comment out the filter_index_enum compatibility *)
+(* definition below (or Import Deprecation.Reject).                           *)
+Fact index_enum_key : unit. Proof. split. Defined. (* Qed. *)
+Definition index_enum (T : finType) :=
+  locked_with index_enum_key (Finite.enum T).
+
+Lemma deprecated_filter_index_enum T P : filter P (index_enum T) = enum P.
+Proof. by rewrite [index_enum T]unlock. Qed.
+
 Lemma mem_index_enum T i : i \in index_enum T.
-Proof. by rewrite -[index_enum T]enumT mem_enum. Qed.
+Proof. by rewrite [index_enum T]unlock -enumT mem_enum. Qed.
 Hint Resolve mem_index_enum : core.
 
-Lemma filter_index_enum T P : filter P (index_enum T) = enum P.
-Proof. by []. Qed.
+Lemma index_enum_uniq T : uniq (index_enum T).
+Proof. by rewrite [index_enum T]unlock -enumT enum_uniq. Qed.
 
 Notation "\big [ op / idx ]_ ( i <- r | P ) F" :=
   (bigop idx r (fun i => BigBody i op P%B F)) : big_scope.
@@ -872,10 +887,6 @@ rewrite -big_filter -(big_filter r2) filter_cat.
 by rewrite has_count -size_filter; case: filter.
 Qed.
 
-Lemma big_const_seq r (P : pred I) x :
-  \big[op/idx]_(i <- r | P i) x = iter (count P r) (op x) idx.
-Proof. by rewrite unlock; elim: r => //= i r ->; case: (P i). Qed.
-
 End SeqExtension.
 
 Lemma big_map_id J (h : J -> R) r (P : pred R) :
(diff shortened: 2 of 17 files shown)


CHANGELOG_UNRELEASED.md
```diff
@@ -101,7 +102,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
...
-   This change anticipates an expected evolution of `fintype` to integrate `finmap`. It is likely that the new definition of the `#|A|` notation will hide multiple occurrences of `A`, which will break the `{-2}` induction idiom. Client libraries should update before the 1.11 release.
+   This change anticipates an expected evolution of `fintype` to integrate `finmap`. It is likely that the new definition of the `#|A|` notation will hide multiple occurrences of `A`, which will break the `{-2}` induction idiom. Client libraries should update before the 1.11 release (see PR #434 for examples).
+   
+ - Replaced the use of the accidental convertibility between `enum A` and 
+   `filter A (index_enum T)` with more explicit lemmas `big_enumP`, big_enum`, `big_enum_cond`, `big_image` added to the `bigop` library. This convertibilty, and indeed equality may no longer hold in future `mathcomp` releases when sets over `finType`s are generalised to finite sets over `choiceType`s, so client libraraies should stop relying on this convertibility, as well as the `filter_index_enum` lemma. File `bigop.v` has some boilerplate to help with the port; also see PR #441 for examples.
```
CohenCyril: missing opening `` ` `` before `big_enum`

CHANGELOG_UNRELEASED.md
```diff
@@ -101,7 +102,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
...
+   
+ - Restricted `big_image`, big_image_cond`, `big_image_id` and `big_image_cond_id`
```
CohenCyril: missing opening `` ` `` before `big_image_cond`

CHANGELOG_UNRELEASED.md
```diff
@@ -101,7 +102,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Replaced the legacy generalised induction idiom with a more robust one
 that does not rely on the `{-2}` numerical occurrence selector, using
 new `ssrnat` helper lemmas `ltn_ind`, `ubnP`, `ubnPleq`,  ...., (see above). The new idiom is documented in `ssrnat`.
-   This change anticipates an expected evolution of `fintype` to integrate `finmap`. It is likely that the new definition of the `#|A|` notation will hide multiple occurrences of `A`, which will break the `{-2}` induction idiom. Client libraries should update before the 1.11 release.
+   This change anticipates an expected evolution of `fintype` to integrate `finmap`. It is likely that the new definition of the `#|A|` notation will hide multiple occurrences of `A`, which will break the `{-2}` induction idiom. Client libraries should update before the 1.11 release (see PR #434 for examples).
```
CohenCyril: Make PR an hyper link
```suggestion
   This change anticipates an expected evolution of `fintype` to integrate `finmap`. It is likely that the new definition of the `#|A|` notation will hide multiple occurrences of `A`, which will break the `{-2}` induction idiom. Client libraries should update before the 1.11 release (see PR #434 for examples).
```
```suggestion
   This change anticipates an expected evolution of `fintype` to integrate `finmap`. It is likely that the new definition of the `#|A|` notation will hide multiple occurrences of `A`, which will break the `{-2}` induction idiom. Client libraries should update before the 1.11 release (see PR [#434](https://github.com/math-comp/math-comp/pull/434) for examples).
```

CHANGELOG_UNRELEASED.md
```diff
@@ -101,7 +102,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
...
+   
+ - Replaced the use of the accidental convertibility between `enum A` and 
+   `filter A (index_enum T)` with more explicit lemmas `big_enumP`, big_enum`, `big_enum_cond`, `big_image` added to the `bigop` library. This convertibilty, and indeed equality may no longer hold in future `mathcomp` releases when sets over `finType`s are generalised to finite sets over `choiceType`s, so client libraraies should stop relying on this convertibility, as well as the `filter_index_enum` lemma. File `bigop.v` has some boilerplate to help with the port; also see PR #441 for examples.
+   
+ - Restricted `big_image`, big_image_cond`, `big_image_id` and `big_image_cond_id`
+ to `bigop`s over _abelian_ monoids, anticipating the change in the definition of `enum`. This may introduce some incompatibilities - non-abelian instances should be dealt with a combination of `big_map` and `big_enumP`.
```
CohenCyril: Is there an example of this in the library right now, if so can you indicate the theorem + file?

mathcomp/ssreflect/bigop.v
```diff
@@ -545,20 +545,34 @@ Canonical bigop_unlock := Unlockable BigOp.bigopE.
...
+(* be equal up to permutation in this new theory.                             *)
+(*  It is therefore advisable to stop relying on this, and use the new        *)
+(* facilities provided in this library: lemmas big_enumP, big_enum, big_image *)
+(* and such. Users wishing to test compliance should change the Defined in    *)
+(* index_enum_key to Qed, and comment out Lemma filter_index_enum.            *)
+Fact index_enum_key : unit. Proof. split. Defined. (* Qed. *)
+Definition index_enum (T : finType) :=
+  locked_with index_enum_key (Finite.enum T).
+
+Lemma filter_index_enum T P : filter P (index_enum T) = enum P.
```
CohenCyril: Should we deprecate this lemma?

mathcomp/ssreflect/bigop.v
```diff
@@ -1070,45 +1080,68 @@ rewrite big_ltn // big_add1 /= big_mkord; congr op.
...
+  BigEnumSpec e of
+    forall R idx op (F : I -> R),
+      \big[op/idx]_(i <- e) F i = \big[op/idx]_(i | P i) F i
+  & uniq e /\ (forall i, i \in e = P i)
+  & (let cP := [pred i | P i] in perm_eq e (enum cP) /\ size e = #|cP|)
+  : big_enum_spec P e.
+
+(*   This lemma can be used to introduce an enumeration into a non-abelian    *)
+(* bigop, in one of three ways:                                               *)
+(*   have [e big_e [Ue mem_e] [e_enum size_e]] := big_enumP P.                *)
```
CohenCyril: Maybe reference one canonical example of use of this pattern in the existing code?

CHANGELOG_UNRELEASED.md
ggonthier: Right; I was expecting markdown to turn the reference into a hyperlink as it does in the comments (viz. #434), but apparently that doesn't work.

CohenCyril: > Right; I was expecting markdown to turn the reference into a hyperlink as it does in the comments (viz. #434), but apparently that doesn't work.

it works only on github :wink: but the CHANGELOG is destined to be read independently

CHANGELOG_UNRELEASED.md
ggonthier: There isn't, and indeed none of the `big_image` lemmas were used in the library at any point, and almost all the instances where `big_map` + `big_filter` or `enumT` were used as a proxy for `big_image` _were_ in the abelian case, so the PR actually introduces `big_image` there. The only exception is the `split_diagA` technical lemma in `mxpoly`, but it's non-standard because I actually _define_ `diagA` to be the `map` over the enumeration that is meant to be abstracted by `big_enumP`, which means I don't use `big_enumP` to abstract over it right after I use `big_map`, as I normally should. That twist is due to the fact I can't use a lemma like `big_enumP` to create constants at the vernacular level.

CHANGELOG_UNRELEASED.md
ggonthier: Since we've switched to markdown the CHANGELOG is going to be hard to read outside GitHub anyway.., but it actually doesn't even work when viewing the LOG file on GitHub.

mathcomp/ssreflect/bigop.v
ggonthier: Yes, indeed this did cross my mind, but I thought the bespoke `deprecate` we use (until we can rely on 8.10) would print the wrong indication. You've just reminded me my own contraption can actually be made to do the right thing.

ggonthier:
I've implemented the documentation and deprecation improvements, and rebased.

CHANGELOG_UNRELEASED.md
```diff
@@ -101,7 +102,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
...
+   This change anticipates an expected evolution of `fintype` to integrate `finmap`. It is likely that the new definition of the `#|A|` notation will hide multiple occurrences of `A`, which will break the `{-2}` induction idiom. Client libraries should update before the 1.11 release (see PR #434(https://github.com/math-comp/math-comp/pull/434) for examples).
+   
+ - Replaced the use of the accidental convertibility between `enum A` and 
+   `filter A (index_enum T)` with more explicit lemmas `big_enumP`, `big_enum`, `big_enum_cond`, `big_image` added to the `bigop` library, and deprecated the `filter_index_enum` lemma that states the corresponding equality. Both convertibility and equality may no longer hold in future `mathcomp` releases when sets over `finType`s are generalised to finite sets over `choiceType`s, so client libraries should stop relying on this identity. File `bigop.v` has some boilerplate to help with the port; also see PR #441(https://github.com/math-comp/math-comp/pull/441) for examples.
```
CohenCyril: The proper syntax for hyperlinks in markdown (to be printed nicely on github) is `[text](url)`.
You could simply click to commit my suggestion below:
```suggestion
   `filter A (index_enum T)` with more explicit lemmas `big_enumP`, `big_enum`, `big_enum_cond`, `big_image` added to the `bigop` library, and deprecated the `filter_index_enum` lemma that states the corresponding equality. Both convertibility and equality may no longer hold in future `mathcomp` releases when sets over `finType`s are generalised to finite sets over `choiceType`s, so client libraries should stop relying on this identity. File `bigop.v` has some boilerplate to help with the port; also see [PR #441](https://github.com/math-comp/math-comp/pull/441) for examples.
```

CHANGELOG_UNRELEASED.md
```diff
@@ -101,7 +102,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Replaced the legacy generalised induction idiom with a more robust one
 that does not rely on the `{-2}` numerical occurrence selector, using
 new `ssrnat` helper lemmas `ltn_ind`, `ubnP`, `ubnPleq`,  ...., (see above). The new idiom is documented in `ssrnat`.
-   This change anticipates an expected evolution of `fintype` to integrate `finmap`. It is likely that the new definition of the `#|A|` notation will hide multiple occurrences of `A`, which will break the `{-2}` induction idiom. Client libraries should update before the 1.11 release.
+   This change anticipates an expected evolution of `fintype` to integrate `finmap`. It is likely that the new definition of the `#|A|` notation will hide multiple occurrences of `A`, which will break the `{-2}` induction idiom. Client libraries should update before the 1.11 release (see PR #434(https://github.com/math-comp/math-comp/pull/434) for examples).
```
CohenCyril:
```suggestion
   This change anticipates an expected evolution of `fintype` to integrate `finmap`. It is likely that the new definition of the `#|A|` notation will hide multiple occurrences of `A`, which will break the `{-2}` induction idiom. Client libraries should update before the 1.11 release (see PR [#434](https://github.com/math-comp/math-comp/pull/434) for examples).
```

CHANGELOG_UNRELEASED.md
ggonthier: Now fixed; I was trying to roll all changes together so I just snarfed the text you suggested, but missed the []. I took the opportunity to put them both around the [PR #nnn] expression.

mathcomp/ssreflect/bigop.v
CohenCyril: Sorry about that I got confused about the description that comes below.

mathcomp/ssreflect/bigop.v
```diff
@@ -1070,45 +1081,72 @@ rewrite big_ltn // big_add1 /= big_mkord; congr op.
...
+(* gives a permutation e of enum P alongside a equation big_e for converting  *)
+(* between bigops iterating on (i <- e) and ones on (i | P i). Usually not    *)
+(* all properties of e are needed, but see below the proof of big_distr_big   *)
```
CohenCyril: I believe the lemma you wanted to cite was `big_distr_big_dep`

ggonthier: Right again (& fixed...)

PR #443, opened 2019-11-28
Refactoring and linting proofs especially in polydiv.v

pi8027:

This PR refactors several proofs especially in `polydiv.v` but does not change any definition and statements of theorems.
- Replace `altP eqP` and `altP (_ =P _)` with `eqVneq`:
  The improved `eqVneq` lemma (#351) is redesigned as a comparison predicate and introduces a hypothesis in the form of `x != y` in the second case. Thus, `case: (altP eqP)`, `case: (altP (x =P _))` and `case: (altP (x =P y))` idioms can be replaced with `case: eqVneq`, `case: (eqVneq x)` and `case: (eqVneq x y)` respectively. This replacement slightly simplifies and reduces proof scripts.
- Use `have [] :=` rather than `case:` if it is better.
- `by apply:` -> `exact:`.
- `apply/lem1; apply/lem2` or `apply: lem1; apply: lem2` -> `apply/lem1/lem2`.
- `move/lem1; move/lem2` -> `move/lem1/lem2`.
- Remove `GRing.` prefix if applicable.
- `negbTE` -> `negPf`, `eq_refl` -> `eqxx`, and `sym_equal` -> `esym`.

##### Things done/to do

### DIFF
--- a/mathcomp/algebra/finalg.v
+++ b/mathcomp/algebra/finalg.v
@@ -161,7 +161,7 @@ Lemma zmod1gE : 1%g = 0 :> U.            Proof. by []. Qed.
 Lemma zmodVgE x : x^-1%g = - x.          Proof. by []. Qed.
 Lemma zmodMgE x y : (x * y)%g = x + y.   Proof. by []. Qed.
 Lemma zmodXgE n x : (x ^+ n)%g = x *+ n. Proof. by []. Qed.
-Lemma zmod_mulgC x y : commute x y.      Proof. exact: GRing.addrC. Qed.
+Lemma zmod_mulgC x y : commute x y.      Proof. exact: addrC. Qed.
 Lemma zmod_abelian (A : {set U}) : abelian A.
 Proof. by apply/centsP=> x _ y _; apply: zmod_mulgC. Qed.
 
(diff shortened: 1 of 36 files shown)


pi8027:
I have discovered many different kinds of redundancies mainly in `polydiv` so reduced them. I marked this PR as a draft just to do not merge before #270 so that we do not introduce new conflicts in #270.

CohenCyril:
@pi8027 thanks, I think this is overall a great improvement. However I have a small doubt on the style of `have stuff : ... := ...`, and I will ask others about it...

@ggonthier @gares @amahboubi  is there an (unwritten?) invariant of the ssreflect tactic language stating that the intro pattern occurring right after the `have` keyword must/should only consume elements of the stack introduced by the statement occuring after `:`. Or is it ok to also introduce (or case or use a forward view on, etc) hypothesis that were there before the `have`, as in https://github.com/math-comp/math-comp/pull/443/files#diff-dd7d7be19aa85c8361134ccf162dca93R1177.

gares:
Apparently the invariant is not enforced. It makes sense to me, but I have no idea if the lack of enforcement was widely exploiter or if there are good use cases. Correct me if I'm wrong, but in this PR it is a bit borderline.

What I think is enforced is that you do one case split, no more, but then nothing prevents you to introduce more than what comes from the case in each branch...

pi8027:
Since I need to rebase some patches on top of this, I would like to revert that kind of use of `have` so that we can merge this PR without more discussion on that point.

pi8027:
> Since I probably need to rebase some patches on top of this, I would like to revert that kind of use of `have` so that we can merge this PR without more discussion on that point.

Done. @CohenCyril

PR #445, opened 2019-12-02
take advantage of opam variables and their default values

ybertot:

When proposing new packages to opam-coq-archive for release 1.10.0, maintainers of this archive suggested that we remove 'name: ' and 'version: ' lines from the opam files, because these are variables that can receive a default value, fixed by the name of the file or the name of directory above.  Similarly, the fact that several `coq-mathcomp-*` share the same version can be carried around using the `version` variable.  These modifications lead to changes in the `etc/utils/packager` script.

The code touched by this pull-request is relevant to developers who want to work with the development version of math-comp and opam using the opam files found in the git root directory and to future release managers.

##### Things done/to do

### DIFF
--- a/coq-mathcomp-algebra.opam
+++ b/coq-mathcomp-algebra.opam
@@ -1,5 +1,4 @@
 opam-version: "2.0"
-name: "coq-mathcomp-algebra"
 version: "dev"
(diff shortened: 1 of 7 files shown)


gares:
I don't think we want a changelog entry for this fix, so I'm merging as is.

Issue #297, opened 2019-03-12
19 inconvertible ambiguous paths in ssralg


pi8027:
I haven't checked how serious is this issue. But they don't seem to be used for packed classes. I found this in https://github.com/coq/coq/pull/9743.

```
Ambiguous paths: [GRing.subring_closedM; GRing.smulr_closedN] : GRing.subring_closed >-> GRing.oppr_closed
Ambiguous paths:
[GRing.subring_closed_semi; GRing.semiring_closedM] : GRing.subring_closed >-> GRing.mulr_closed
[GRing.subring_closed_semi; GRing.semiring_closedD] : GRing.subring_closed >-> GRing.addr_closed
Ambiguous paths: [GRing.sdivr_closed_div; GRing.divr_closedM] : GRing.sdivr_closed >-> GRing.mulr_closed
Ambiguous paths:
[GRing.subalg_closedBM; GRing.subring_closedB; GRing.zmod_closedN] : GRing.subalg_closed >-> GRing.oppr_closed
[GRing.subalg_closedBM; GRing.subring_closedB; GRing.zmod_closedD] : GRing.subalg_closed >-> GRing.addr_closed
[GRing.subalg_closedBM; GRing.subring_closedB] : GRing.subalg_closed >-> GRing.zmod_closed
Ambiguous paths:
[GRing.divring_closed_div; GRing.sdivr_closedM; GRing.smulr_closedM] : GRing.divring_closed >-> GRing.mulr_closed
[GRing.divring_closed_div; GRing.sdivr_closedM] : GRing.divring_closed >-> GRing.smulr_closed
[GRing.divring_closed_div; GRing.sdivr_closedM; GRing.smulr_closedN] : GRing.divring_closed >-> GRing.oppr_closed
Ambiguous paths:
[GRing.divalg_closedBdiv; GRing.divring_closedBM] : GRing.divalg_closed >-> GRing.subring_closed
[GRing.divalg_closedBdiv; GRing.divring_closedBM; GRing.subring_closed_semi] : GRing.divalg_closed >-> GRing.semiring_closed
[GRing.divalg_closedBdiv; GRing.divring_closedBM; GRing.subring_closedM; GRing.smulr_closedM] : GRing.divalg_closed >-> GRing.mulr_closed
[GRing.divalg_closedBdiv; GRing.divring_closedBM; GRing.subring_closedM] : GRing.divalg_closed >-> GRing.smulr_closed
[...]

PR #447, opened 2019-12-04
treating deprecation warnings that appear with coq-8.10

ybertot:

Remove deprecation warnings that became apparent at the time of the 1.10.0 release.
##### Things done/to do

### DIFF
--- a/mathcomp/_CoqProject
+++ b/mathcomp/_CoqProject
@@ -5,3 +5,5 @@
 -arg -w -arg -redundant-canonical-projection
 -arg -w -arg -notation-overridden
 -arg -w -arg -duplicate-clear
+-arg -w -arg -ambiguous-paths
+-arg -w -arg -undeclared-scope
--- a/mathcomp/algebra/mxalgebra.v
+++ b/mathcomp/algebra/mxalgebra.v
@@ -1999,7 +1999,7 @@ Arguments capmx_idPr {F n m1 m2 A B}.
 Arguments capmx_idPl {F n m1 m2 A B}.
 Arguments bigcapmx_inf [F I] i0 [P m n A_ B].
 Arguments sub_bigcapmxP {F I P m n A B_}.
-Arguments mxrank_injP {F m n} p [A f].
+Arguments mxrank_injP {F m n} p {A f}.
 Arguments mxdirectP {F n S}.
 Arguments mxdirect_addsP {F m1 m2 n A B}.
 Arguments mxdirect_sumsP {F I P n A_}.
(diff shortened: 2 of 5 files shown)


mathcomp/_CoqProject
```diff
@@ -5,3 +5,5 @@
 -arg -w -arg -redundant-canonical-projection
 -arg -w -arg -notation-overridden
 -arg -w -arg -duplicate-clear
+-arg -w -arg -ambiguous-paths
+-arg -w -arg -undeclared-scope
```
CohenCyril: Is this silencing "undeclared-scope" warnings? If so it is not completely clear to me that we want to do this rather than fixing it...

mathcomp/algebra/mxalgebra.v
```diff
@@ -1999,7 +1999,7 @@ Arguments capmx_idPr {F n m1 m2 A B}.
 Arguments capmx_idPl {F n m1 m2 A B}.
 Arguments bigcapmx_inf [F I] i0 [P m n A_ B].
 Arguments sub_bigcapmxP {F I P m n A B_}.
-Arguments mxrank_injP {F m n} p [A f].
+Arguments mxrank_injP {F m n} p {A f}.
```
CohenCyril: Why do you declare `{A f}` as maximal implicits?
I think these were really mean to be "on demand" so that a naked `mxrank_injP` is still applied implicitly `F m n` (for view compositions if I am correct) but not applied to `A f` unless we give `p`...
Same question for all the following changes.

mathcomp/_CoqProject
```diff
@@ -5,3 +5,5 @@
 -arg -w -arg -redundant-canonical-projection
 -arg -w -arg -notation-overridden
 -arg -w -arg -duplicate-clear
+-arg -w -arg -ambiguous-paths
```
pi8027: IMO, we should be aware of ambiguous paths with regard to packed classes. Also, I'm trying to reduce the number of ambiguous paths reported so that we can remove this from `mathcomp/Make` (see #297, coq/coq#11258, and coq/coq#11381).

pi8027:
By the way, what about turning treated deprecation warnings to errors by adding `-arg -w -arg +...` so that we do not introduce the use of deprecated features, e.g., #438?

mathcomp/_CoqProject
CohenCyril: Oh, after re-reading myself, it seems obvious that the reason is backward compatibility. My bad...

mathcomp/algebra/mxalgebra.v
CohenCyril: @ybertot I still do not understand the motivation behind this change though

anton-trunov: I think that the change is needed, because, unless I'm mistaken, non-maximal implicits are forced only if the term is applied to at least one more argument after them. So the trailing implicits only make sense if they are maximal.

ybertot: Not making this argument maximal provokes another warning:

`Warning: Argument number 4 is a trailing implicit so must be maximal`

pi8027: Those changes have already been merged in #452.

CohenCyril:
@ybertot sorry this PR went completely under my radar for a while... @affeldt-aist why is this for 1.12.0 rather than 1.11.0 ?

PR #454, opened 2020-01-10
Renaming converse to dual in order.v

CohenCyril:
The name `converse` and notation `^c` were already taken in
https://github.com/math-comp/math-comp/blob/master/mathcomp/algebra/ssralg.v#L989-L990
I am renaming to `dual` following @hivert recommendation https://github.com/math-comp/finmap/issues/23.

##### Things done/to do

- ~added corresponding entries in `CHANGELOG_UNRELEASED.md`~

### DIFF
--- a/mathcomp/ssreflect/order.v
+++ b/mathcomp/ssreflect/order.v
@@ -2147,10 +2147,10 @@ Export FinTotal.Exports.
 
 Definition dual T : Type := T.
 Definition dual_display : unit -> unit. Proof. exact. Qed.
-Local Notation "T ^d" := (dual T) (at level 2, format "T ^d") : type_scope.
 
 Module Import DualSyntax.
 
+Notation "T ^d" := (dual T) (at level 2, format "T ^d") : type_scope.
 Notation "<=^d%O" := (@le (dual_display _) _) : fun_scope.
 Notation ">=^d%O" := (@ge (dual_display _) _)  : fun_scope.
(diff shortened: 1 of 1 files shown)


hivert:
Excellent ! While you are at it, I'd like to be able to write something like:
```
Variable (u : unit) (Ord : orderType u).
Let OrdC := [orderType of Ord^c].
```
However the notation `Ord^c` is not defined.

CohenCyril:
`^c` was renamed to `^d`

hivert:
> `^c` was renamed to `^d`

It doesn't work for me with the updated version.

```
From mathcomp Require Import boolp.

Require Import order.


Set Implicit Arguments.
Unset Strict Implicit.

Import Order.Syntax.
Import Order.TTheory.
Open Scope order_scope.

Section Dual.

Variable (u : unit) (Ord : orderType u).

Let OrdC := [orderType of Ord^d].
```
Gives:
```
Error: The reference d was not found in the current environment.
```
Maybe I didn't export the right module or the right scope. See my precise example above...

hivert:
It seems that it is only defined as a local notation:
```
Local Notation "T ^d" := (dual T) (at level 2, format "T ^d") : type_scope.
```

CohenCyril:
> It seems that it is only defined as a local notation:

oh, right, should be fixed now

hivert:
Excellent ! These missing canonicals seem to fix all my issues ! Thanks for the extremely quick fix.

PR #457, opened 2020-01-28
Find

CohenCyril:

I am adding shortcut lemmas in `seq.v`.
- lemmas `hasNfind` and `memNindex` which state `find` and `index` take their maximum value `size s` in the case where the predicate is not valid or there is non membership
- lemma `findP` which combines `hasNfind`, `nth_find` and `before_find` and makes a case analysis.
- lemmas `foldr_rcons`, `foldl_rcons` and `scanl_rcons`, and also `nth_cons_scanl` to encompass the case `take 0`.

##### Things done/to do

### DIFF
--- a/CHANGELOG_UNRELEASED.md
+++ b/CHANGELOG_UNRELEASED.md
@@ -20,6 +20,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Ported `order.v` from the finmap library, which provides structures of ordered
   sets (`porderType`, `distrLatticeType`, `orderType`, etc.) and its theory.
 
+- Added lemmas `hasNfind`, `memNindex` and `findP` in `seq`
+
 ### Changed
 
 - Reorganized the algebraic hierarchy and the theory of `ssrnum.v`.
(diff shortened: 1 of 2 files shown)


thery:
Just one littlle remark, I am not a big fan of using ```N``` as a separator (```hasNfind``` and ```memNindex```) I think I find ```hasN_find``` and ```memN_index``` like in ```ifN_eq```

CohenCyril:
> Just one littlle remark, I am not a big fan of using `N` as a separator (`hasNfind` and `memNindex`) I think I find `hasN_find` and `memN_index` like in `ifN_eq` more legible

I guess that is a matter of taste, and mine is I like it better in the current way. Furthermore, it seems that the most wide spread convention is to use it as a separator as in `gtnNdvd`, `contraNeq`, `ltnNge`,  and at least 9 other occurrences of this kind... (cf `grep -e "[a-z0-9]N[a-z0-9]" **/*.v | grep "Lemma"` filtering out the numerous examples where  `N` is used as for arithmetic negation) while the example you cite has exactly two occurences `ifN_eq` and `ifN_eqC`.

thery:
as there are so few occurrences we can still go either ways :smile:
Maybe we can merge this pull request and leave the N as an issue for later

ybertot:
Thanks @CohenCyril and @thery

PR #462, opened 2020-03-01
Using primitive records with packed classes in order.v

pi8027:

This PR demonstrates that it is possible to use primitive records with packed classes. With some more work, it should be possible to make `L^d^d` and `L` definitionally equal where `L` is any `latticeType` and `^d` means dual.

##### Things done/to do

### DIFF
--- a/mathcomp/ssreflect/order.v
+++ b/mathcomp/ssreflect/order.v
@@ -385,6 +385,7 @@ From mathcomp Require Import path fintype tuple bigop finset div prime.
 Set Implicit Arguments.
 Unset Strict Implicit.
 Unset Printing Implicit Defensive.
+Local Set Primitive Projections.
 
 Delimit Scope order_scope with O.
 Local Open Scope order_scope.
@@ -962,10 +963,10 @@ Section ClassDef.
 Record mixin_of (T : eqType) := Mixin {
   le : rel T;
   lt : rel T;
-  _  : forall x y, lt x y = (y != x) && (le x y);
-  _  : reflexive     le;
-  _  : antisymmetric le;
-  _  : transitive    le
+  lt_def   : forall x y, lt x y = (y != x) && (le x y);
+  lexx     : reflexive     le;
+  le_anti  : antisymmetric le;
+  le_trans : transitive    le
 }.
 
 Record class_of T := Class {
@@ -975,17 +976,17 @@ Record class_of T := Class {
 
 Local Coercion base : class_of >-> Choice.class_of.
 
-Structure type (disp : unit) := Pack { sort; _ : class_of sort }.
+Structure type (disp : unit) :=
+  Pack { sort; #[canonical(false)] class : class_of sort }.
 
 Local Coercion sort : type >-> Sortclass.
 
 Variables (T : Type) (disp : unit) (cT : type disp).
 
-Definition class := let: Pack _ c as cT' := cT return class_of cT' in c.
-Definition clone c of phant_id class c := @Pack disp T c.
-Definition clone_with disp' c of phant_id class c := @Pack disp' T c.
+Definition clone c of phant_id (class cT) c := @Pack disp T c.
+Definition clone_with disp' c of phant_id (class cT) c := @Pack disp' T c.
 Let xT := let: Pack T _ := cT in T.
-Notation xclass := (class : class_of xT).
+Notation xclass := (class cT : class_of xT).
 
 Definition pack :=
   fun bT b & phant_id (Choice.class bT) b =>
@@ -1142,13 +1143,13 @@ Section ClassDef.
 Record mixin_of d (T : porderType d) := Mixin {
   meet : T -> T -> T;
   join : T -> T -> T;
-  _ : commutative meet;
-  _ : commutative join;
-  _ : associative meet;
-  _ : associative join;
-  _ : forall y x, meet x (join x y) = x;
-  _ : forall y x, join x (meet x y) = x;
-  _ : forall x y, (x <= y) = (meet x y == x);
+  meetC   : commutative meet;
+  joinC   : commutative join;
+  meetA   : associative meet;
+  joinA   : associative join;
+  joinKI  : forall y x, meet x (join x y) = x;
+  meetKU  : forall y x, join x (meet x y) = x;
+  leEmeet : forall x y, (x <= y) = (meet x y == x);
 }.
 
 Record class_of (T : Type) := Class {
@@ -1159,17 +1160,17 @@ Record class_of (T : Type) := Class {
 
 Local Coercion base : class_of >-> POrder.class_of.
 
-Structure type (disp : unit) := Pack { sort; _ : class_of sort }.
+Structure type (disp : unit) :=
+  Pack { sort; #[canonical(false)] class : class_of sort }.
 
 Local Coercion sort : type >-> Sortclass.
 
 Variables (T : Type) (disp : unit) (cT : type disp).
 
-Definition class := let: Pack _ c as cT' := cT return class_of cT' in c.
-Definition clone c of phant_id class c := @Pack disp T c.
-Definition clone_with disp' c of phant_id class c := @Pack disp' T c.
+Definition clone c of phant_id (class cT) c := @Pack disp T c.
+Definition clone_with disp' c of phant_id (class cT) c := @Pack disp' T c.
 Let xT := let: Pack T _ := cT in T.
-Notation xclass := (class : class_of xT).
+Notation xclass := (class cT : class_of xT).
 
 Definition pack disp0 (T0 : porderType disp0) (m0 : mixin_of T0) :=
   fun bT b & phant_id (@POrder.class disp bT) b =>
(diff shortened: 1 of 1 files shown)


ggonthier:
This PR is meant to demonstrate that `Canonical Structures` can work with primitive records.
The demonstration is only partial because the `#[canonical(false)]` pragmas are syntax errors in all released versions of Coq - this could be easily remedied as these are non-essential optimisation of the `Canonical Projections` table. Also, this PR does not implement the intended application of the switch to primitive records - convertibility between order structures and their bidual.
  Looking deeper into the matter, it seems that this works because the current version of Coq only partially implements eta-conversion for primitive records; it may not be prudent to sort relying on this.
  Specifically, Coq implements eta-convertibility `p ~ (fst p, snd p)`, eta-expansion during unification (reduce `p ~ (a, b)` to `fst p ~ a`, `snd p ~ b`) but not pattern unification for records
(solve `fst ?p ~ a` with `?p := (a, ?s)`). It's the latter that is at odds with `Canonical Structure` inference, since it is tantamount to _always_ resolving structures, as one would by declaring a default instance (as with, e.g., `Canonical Pack.`). This can result in unnatural behaviour and poor diagnostics for erroneous inputs, which are not tested by recompiling (error-free) libraries. It would also break more elaborate uses of `Canonical Structures` that rely on the failure of structure inference to force the expansion of tags or insert coercions (e.g., for `predType`).
  Perhaps we can get some assurance that Coq will never implement unification patterns for records, but this seems unlikely. For one thing, this makes unification sensitive to eta-expansion: `?p ~ (0, 1)` succeeds, but `(fst ?p, snd ?p) ~ (0, 1)` reduces to `fst ?p ~ 0`, `snd ?p ~ 1`, which fail...
  There are also differences in reduction strategy between classical and primitive records, namely that explicit `match` on records are iota-reduces eagerly, whereas primitive projections are not. The `xclass` idiom in packed classes relies on this to mitigate the exponential complexity of term comparison (the details are in the extensive log entry of the commit that introduces that idiom, but that's in the `CoqFinitGroup` svn repository, which has disappeared from `forge` - hopefully @gares still has a copy). The good news is that that particular idiom may no longer be needed if `class_of` objects are primitive records, because the type parameter of `class_of`, which is the cause of the divergent complexity, would be implicit.
  Finally, the intended application does not seem all that compelling. In any concrete use of the order interface, the order structures will be explicit records that will be convertible to their bidual in all cases. This, the change would only improve facilitate writing library code for abstract orders. Further, there is a workaround even for this case, namely using primitive records for classes and some mixins, and parametrising the development that needs bidual convertibility on a manifest `Pack` of `Type` and `class_of` parameters, e.g.,
  ``Context typeT classT (T := @Pack typeT classT).``
  To summarize, perhaps the best course of action would be to switch to primitive records for `class_of` records only, _if_ this removes the need for `xclass`, but keep classical records for `Structures`, and use the idiom above if eta convertibility is needed.

pi8027:
@ggonthier Thank you for the feedback. I finished what I planned here (`L^d^d = L`) in #464 by following your suggestions. I think that your comments are useful alongside my diff to understand the inference mechanism, in particular, the `xclass` idiom. So I close this PR and keep it for a record.

CohenCyril:
BTW, here is the designated commit log message:

> Author: gonthier <gonthier@305884b8-be16-0410-aa53-a3be2363cef1>  2012-03-19 22:21:11
>
>     This revision was meant to be a local improvement to falgbra <-> galois, but
>     it uncovered a major inefficiency of the processing of packed class declarations
>     by the Coq kernel, which could only be mitigated by adjusting the packed class
>     idiom.
>       The problem arises from the interference of several questionable design
>     choices in the Coq kernel, in particular the blind maximal sharing of
>     "reductions" (which turn out to be full expansions here). The diagnosis was
>     prompted by the anomalous behaviour of line 302 of galois.v, which took over 3
>     minutes of user time to execute. This line attempts to use an eqP view on an
>     assumption of the form root p x. This is convertible to p.[x] == 0, but the
>     unfolding heuristic of Coq causes it to unfold the definition of (_ == _) in
>     the eqP statement instead, only turning to Hy once it has reached a head normal
>     form for the (_ == _), and thus having to unfold root p x to head normal form as
>     well. Though unnecessary, by itself this should not be a problem, even though we
>     are 9 levels deep in the hierarchy (eqType >-> fieldExtType). Things break down
>     because the kernel cannot manage to compare the structurally identical terms
>     that result from this expansion. These have the form
>        let: Equality.Mixin op _ :=
>          let: Choice.Class bc _ :=
>            ....
>            let: FieldExt.Class feb _ _ _ :=
>              let: FieldExt.Pack _ fec _ := L return FieldExt.class_of L in fec
>            in feb
>            ...
>          in bc
>        in op.
>     so the problem is not immediately obvious... until one recalls that
>     non-dependent matches also carry explicit return type annotations. Thus the
>     term is in fact
>        let: Equality.Mixin op _  :=
>          let: Choice.Class bc _
>            ....
>            let: FieldExt.Class feb _ _ _ :=
>              let: FieldExt.Pack _ fec _ := L return FieldExt.class_of L in fec
>            return Falgebra.class_of F0 (Phant F0) (FieldExt.sort L)
>            in feb
>            ...
>          return Equality.class_of (Equality.sort Feq)) :=
>          in bc
>          return rel (Equality.sort Feq)
>        in op.
>     where Feq is the eqType structure that was passed explicitly to eq_op.
>     Feq is the result of several type inference refinements, so is is not
>     obtained directly as FieldExt.eqType F0 L, but rather as a composite
>        Zmodule.eqType (Ring.zmodType
>          (Lalgebra.ringType (Falgebra.vect_lalgType (FieldExt.FalgType L))))
>     that evaluates ot the same normal form as FieldExt.eqType F0 L.
>       If the comparison was based on this (slightly complicated) version of Feq
>     there would be no problem. However, to arrive at the chain of lets above the
>     Coq kernel has had to expand the definition of Feq -- and it then proceeds to
>     share this expansion maximally across the entire expression. Feq is thus
>     now a fairly large graph, where each of  Zmodule.eqType, Ring.zmodType, etc
>     have been expanded, and the "base" coercions have been expanded: the class
>     components are subterms of the top "let:" chain, e.g., the class component
>     of Feq is the let: Choice.class ... subterm. As the kernel "optimizes" the
>     comparison of Equality.sort (Equality.Pack <monster record>)
>                    =  Equality.sort (Equality.Pack <monster record>)
>     by not evaluating the projection and comparing all the <monster record>
>     rather that just the sort field, does so by expanding the (shared) term graph
>     into an (unshared) tree, and these type constraints occur everywhere inside the
>     let: chain (including in the invisible lambda binders of the predicates and
>     branch subterms), the diverging behavior is unsurprising.
>       Note that it is the kernel that diverges here: evarconv unification has
>     absolutely no problem here, because it uses a different heuristic which does
>     not allow sharing to interfere with comparison; indeed, it backtracks over
>     the reduction of the argument of a blocked match, effectively allowing the
>     expansion to "align" on the topmost match, without having to expose the others.
>       Our partial workaround to this conundrum is to replace, as much as possible,
>     unexpanded projections like Equality.sort with exposed matches. This causes the
>     kernel to evaluate the projection when it propagates an expanded value for the
>     record, rather than going into the op-described tailspin. The workaround is
>     only partial because some explicit constants must remain to let unification find
>     canonical instances, e.g., the Choice.sort field of Choice.eqType, though
>     fortunately (in this case) evaluation makes most of these vanish. Alas, the
>     largest instance (the Equality.sort _)cannot be made to simplify in this way:
>     it is needed for Canonical Structure unification of, say (x \in pred1 y) with
>     x and y in a finType, with ?x == ?y. The problem here is that evarconv unifies
>     left to right in the Coq v5-7 term order, where the predicate came before the
>     match (the called "case") statement, so it has to infer the eqType for ?x and ?y
>     on the basis of the let: return type.
>       In practice, this means that the packed class idiom gets two extra lines
>      Let xT := let: Pack T _ _ := cT in T.
>      Notation xclass := (class : class_of xT).
>     and that the various canonical "inheritance" instances should now use
>     an explicit cT value for the sort, xclass for the class, and xT for the tag,
>     e.g.,
>       Definition zmodType := @Zmodule.Pack cT xclass xT.
>     The fix seems reasonably effective in spite of these shrotcomings: time for
>     line 302 in galois drops from 167s to 1s user time, which is more acceptable
>     (but still 2 to 3 orders of magnitude too slow!).
>     
>     Other changes in this revision:
>       - "Lagrange" spelling corrected
>       - the Falgebra interface now contains the UnitRing interface (as an Falgebra
>         always has a unitRingType structure -- there is a lemma proving this).
>       - based fieldExtType on Falgebra rather than Field, as this allows all mixins
>         to be in Prop.
>       - added construction of simple field extension from an irreducible polynomial,
>         used it to remove the separability condition in the proof that splitting
>         fields are normal.
>     
>     
>     
>     git-svn-id: svn+ssh://scm.gforge.inria.fr/svn/coqfinitgroup/trunk@3751 305884b8-be16-0410-aa53-a3be2363cef1

CohenCyril:
https://gist.github.com/CohenCyril/c2c7d87c9677ea757a5b67d56a985fb8#file-this-revision-was-meant-to-be-a-local-improvement-to.patch

PR #406, opened 2019-11-01
Commutative Algebras

hivert:
##### Motivation for this change

Currently, there is no interface for commutative algebras and unitary commutative algebras. I'd like to have them to state the initial properties of multivariate polynomials. Also for univariate polynomial, the initial properties is stated in a very weak form (ie: through a ring morphism) which is not always practical. I created `honer_alg a p` which evaluate the polynom `p` at `a : A` for `A : algType`. It is an algebra morphism an slightly easier to use than `horner_moprh`. 

##### Things done/to do

- [X] added corresponding entries in `CHANGELOG_UNRELEASED.md`
- [X] added corresponding documentation in the headers

##### Automatic note to reviewers

Read [this Checklist](https://github.com/math-comp/math-comp/wiki/Checklist-for-following,-reviewing-and-playing-with-a-PR#checklist-for-reviewing-a-pr) and make sure there is a milestone.

PR #463, opened 2020-03-01
Fix hierarchy.ml to compute the transitive closure of a hierarchy

pi8027:

There was a function `is_transitive` in the `hierarchy.ml` utility that checks whether a given inheritance relation is transitive or not, but it has never been used. This patch fixes it by rather computing the transitive closure of the hierarchy. (See https://github.com/math-comp/math-comp/pull/406#pullrequestreview-310810808)

##### Things done/to do


pi8027:
@gares Hello. Could you assign yourself to this PR?

PR #471, opened 2020-04-05
Make `all2` better wrt the guard condition

### LINKED ISSUE #469: all2 vs. the guard condition
Here is an example that is rejected by Coq 8.11 (tested with mathcomp 1.10.0)

~~~
From mathcomp
Require Import all_ssreflect.

Inductive tree := Node { children : seq tree }.

Fail Fixpoint eqb (x y: tree) {struct x} : bool :=
  all2 eqb x.(children) y.(children).
(*
Recursive definition of eqb is ill-formed.
In environment
eqb : tree -> tree -> bool
x : tree
y : tree
Recursive call to eqb has not enough arguments.
Recursive definition is: "fun x y : tree => all2 eqb (children x) (children y)".
*)
~~~

Regarding the guard condition, the initial implementation of the `all2` function (in #224) was better. I don’t know why it was rewritten in #324

CohenCyril:

fixes #469

##### Things done/to do

### DIFF
--- a/mathcomp/ssreflect/seq.v
+++ b/mathcomp/ssreflect/seq.v
@@ -2634,7 +2634,7 @@ Prenex Implicits mask map pmap foldr foldl scanl pairmap.
 
 Section Zip.
 
-Variables S T : Type.
+Variables (S T : Type) (r : S -> T -> bool).
 
 Fixpoint zip (s : seq S) (t : seq T) {struct t} :=
   match s, t with
@@ -2645,10 +2645,10 @@ Fixpoint zip (s : seq S) (t : seq T) {struct t} :=
 Definition unzip1 := map (@fst S T).
 Definition unzip2 := map (@snd S T).
 
-Fixpoint all2 (r : S -> T -> bool) s t :=
+Fixpoint all2 s t :=
   match s, t with
   | [::], [::] => true
-  | x :: s, y :: t => r x y && all2 r s t
+  | x :: s, y :: t => r x y && all2 s t
   | _, _ => false
   end.
 
@@ -2696,8 +2696,8 @@ move: s t; apply: seq_ind2 => //= x y s t eq_sz IHs.
 by rewrite !rev_cons IHs zip_rcons ?size_rev.
 Qed.
 
-Lemma all2E r s t :
-  all2 r s t = (size s == size t) && all [pred xy | r xy.1 xy.2] (zip s t).
+Lemma all2E s t :
+  all2 s t = (size s == size t) && all [pred xy | r xy.1 xy.2] (zip s t).
 Proof. by elim: s t => [|x s IHs] [|y t] //=; rewrite IHs andbCA. Qed.
 
 End Zip.
--- a/CHANGELOG_UNRELEASED.md
+++ b/CHANGELOG_UNRELEASED.md
@@ -143,6 +143,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
     `ssrnat.mc_1_10` module. One may compile proofs compatible with the version
     1.10 in newer versions by using this module.
 
+- The definition of `all2` was slightly altered for a better interaction with
+  the guard condition (#469)
+
 ### Renamed
 
 - `real_lerP` -> `real_leP`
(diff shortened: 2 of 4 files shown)


mathcomp/ssreflect/seq.v
```diff
@@ -2645,10 +2645,10 @@ Fixpoint zip (s : seq S) (t : seq T) {struct t} :=
 Definition unzip1 := map (@fst S T).
 Definition unzip2 := map (@snd S T).
 
-Fixpoint all2 (r : S -> T -> bool) s t :=
+Definition all2 (r : S -> T -> bool) := fix loop s t :=
```
pi8027, line 2648: I suggest moving `all2` before `all2E` and use `Variable` to factor out `r`. This allows us to use `Fixpoint` rather than `Definition` and then slightly improves the readability.

pi8027:
It would be nice to have a comment about the guard condition not to revert this change later, but this may be more general and placed somewhere else. I think what we should say here, in general, is that we should try to factor out function arguments of higher-order, polymorphic, recursive functions. For example, can we reproduce the same issue with the `map` function?

pi8027:
> For example, can we reproduce the same issue with the `map` function?

I self-answer this question. It is possible:
```coq

Set Implicit Arguments.
Unset Strict Implicit.

Inductive tree (A : Type) := singleton of A | branch of list (tree A).

Module Bad.

Fixpoint map (A B : Type) (f : A -> B) (xs : list A) : list B :=
  if xs is x :: xs then f x :: map f xs else [::].

Fail Fixpoint tmap (A B : Type) (f : A -> B) (t : tree A) : tree B :=
  match t with
    | singleton x => singleton (f x)
    | branch ts => branch (map (tmap f) ts)
  end.

End Bad.

Module Good.

Definition map (A B : Type) (f : A -> B) :=
  fix map' (xs : list A) := if xs is x :: xs then f x :: map' xs else [::].

Definition tmap (A B : Type) (f : A -> B) := fix tmap' (t : tree A) : tree B :=
  match t with
    | singleton x => singleton (f x)
    | branch ts => branch (map tmap' ts)
  end.

End Good.
```

To conclude, if this is not a commonly known technique to write terminating functions, it would worth documenting it somewhere.

mathcomp/ssreflect/seq.v
affeldt-aist: I would agree.

affeldt-aist:
> To conclude, if this is not a commonly known technique to write terminating functions, it would worth
> documenting it somewhere.

I would say https://github.com/math-comp/math-comp/blob/master/CONTRIBUTING.md.

mathcomp/ssreflect/seq.v
affeldt-aist: @CohenCyril Which solution do you think is best? (1) The change proposed by @pi8027 (with a comment to say that it is crucial that the `Fixpoint` be inside a `Section`) or (2) add a comment just before the `Definition` using `fix`?

pi8027:
@affeldt-aist Sure. Also, we can do that in a separate PR.

mathcomp/ssreflect/seq.v
CohenCyril: The third alternative is to move nothing but add a a section variable, as in the new commit I just pushed.

CohenCyril:
> > For example, can we reproduce the same issue with the `map` function?
> 
> I self-answer this question. It is possible:
> 
> ```coq
> Module Good.
> 
> Definition map (A B : Type) (f : A -> B) :=
>   fix map' (xs : list A) := if xs is x :: xs then f x :: map' xs else [::].
> 
> Definition tmap (A B : Type) (f : A -> B) := fix tmap' (t : tree A) : tree B :=
>   match t with
>     | singleton x => singleton (f x)
>     | branch ts => branch (map tmap' ts)
>   end.
> 
> End Good.
> ```
> 
> To conclude, if this is not a commonly known technique to write terminating functions, it would worth documenting it somewhere.

I think such code should be added to the test-suite.

pi8027:
@CohenCyril 
> I think such code should be added to the test-suite, in order to detect possible reverts...

So true. Could you add a test case for `all2`? I will add more test cases later.

CohenCyril:
> Looks good to me, modulo a test case. Since we can add it later, it should not block the release.

I added the test case.

CohenCyril:
@affeldt-aist we reached consensus, you can merge

PR #474, opened 2020-04-07
Documentation typos

llelf:


affeldt-aist:
For me, this PR can be merged as it is.

mathcomp/ssreflect/ssrnat.v
```diff
@@ -90,7 +90,7 @@ Require Export Ring.
 (*   V (infix) -- disjunction, as in                                          *)
 (*      leq_eqVlt : (m <= n) = (m == n) || (m < n).                           *)
 (*   X - exponentiation, as in lognX : logn p (m ^ n) = logn p m * n in       *)
-(*         file prime.v (the suffix is not used in ths file).                 *)
+(*         file prime.v (the suffix is not used in the file).                 *)
```
ybertot, line 93:
```suggestion
(*         file prime.v (the suffix is not used in this file).                 *)
```

mathcomp/ssreflect/tuple.v
```diff
@@ -18,7 +18,7 @@ Unset Printing Implicit Defensive.
 (* [tuple x1; ..; xn] == the explicit n.-tuple <x1; ..; xn>.                  *)
 (*  [tuple E | i < n] == the n.-tuple with general term E (i : 'I_n is bound  *)
 (*                       in E).                                               *)
-(*        tcast Emn t == the m-tuple t cast as an n-tuple using Emn : m = n.  *)
+(*        tcast Emn t == the m.-tuple t cast as an n-tuple using Emn : m = n. *)
```
ybertot, line 21: I disagree with this modification.  I don't see why adding a period here will help.  Can you elaborate the need for this change?

mathcomp/ssreflect/eqtype.v
```diff
@@ -47,7 +47,7 @@ From mathcomp Require Import ssreflect ssrfun ssrbool.
 (*  predU1 a P, predD1 P a == applicative versions of the above.              *)
 (*              frel f == the relation associated with f : T -> T.            *)
 (*                     := [rel x y | f x == y].                               *)
-(*       invariant k f == elements of T whose k-class is f-invariant.         *)
+(*       invariant f k == elements of T whose k-class is f-invariant.         *)
```
ybertot, line 50: While you are at it, you should correct line 440 of the same file:
(* The invariant of an function ... *)  --> invariant of a function

mathcomp/ssreflect/tuple.v
llelf: For consistency. `foo.-tuple` (as a notation for `tuple_of`) is used everywhere in this file's header.

llelf: it should be even `the m.-tuple t cast as an n.-tuple` then.
@ybertot apply or remove?

ybertot: @llelf You convinced me, and you are right about the second correction on the same line: --> "n.tuple"

pi8027: So I propose to fix L.16: `(size s)-tuple` -> `(size s).-tuple`.

llelf: fixed both

ybertot:
Thanks a lot @llelf, it was great working with you on this pull request.

PR #429, opened 2019-11-15
Extend comparison predicates for nat with minn and maxn and reorder arguments of those in order.v

pi8027:
##### Motivation for this change

This PR partially addresses #404. It extends the comparison predicates `leqP`, `ltnP`, and `ltngtP` in `ssrnat.v` to allow case analysis on `minn` and `maxn`. It also changes the ordering of arguments of `lcomparableP`, `(lcomparable_)leP`, `(lcomparable_)ltP`, and `(lcomparable_)ltgtP` in `order.v`.

##### Things done/to do

- [x] merge dependency: #435
- [x] added a compatibility module
- [x] added corresponding entries in `CHANGELOG_UNRELEASED.md`
- ~[ ] added corresponding documentation in the headers~

##### Automatic note to reviewers

Read [this Checklist](https://github.com/math-comp/math-comp/wiki/Checklist-for-following,-reviewing-and-playing-with-a-PR#checklist-for-reviewing-a-pr) and make sure there is a milestone.

PR #495, opened 2020-05-04
Reword a CHANGELOG entry introduced in #429

pi8027:
that explains an incompatibility between development versions, as I suggested in https://github.com/math-comp/math-comp/pull/429#discussion_r392858464.

##### Things done/to do


gares:
As per WG this need an update (instead of removing the entry, write how to migrate)

pi8027:
Done.

PR #501, opened 2020-05-08
Intro pattern extensions for rewrite

CohenCyril:

 Intro pattern ltac views (rewrite, dup, swap, apply) in order not to break intro patterns for silly rewrites.
  - calling rewrite from an intro pattern, use with parsimony
    `=> /[1! rules]`  does `rewrite 1!rules`
    `=> /[! rules]`  does `rewrite !rules`

This is a part of #372, simplified, rewritten and rebased.

##### Things done/to do

- [ ] open corresponding PR in Coq to integrate changes from `ssreflect.v`

### DIFF
--- a/mathcomp/algebra/intdiv.v
+++ b/mathcomp/algebra/intdiv.v
@@ -1006,7 +1006,7 @@ without loss{IHa} /forallP/(_ (_, _))/= a_dvM: / [forall k, a %| M k.1 k.2]%Z.
     by exists i; rewrite mxE.
   exists R^T; last exists L^T; rewrite ?unitmx_tr //; exists d => //.
   rewrite -[M]trmxK dM !trmx_mul mulmxA; congr (_ *m _ *m _).
-  by apply/matrixP=> i1 j1; rewrite !mxE; case: eqVneq => // ->.
+  by apply/matrixP=> i1 j1 /[!mxE]; case: eqVneq => // ->.
 without loss{nz_a a_dvM} a1: M a Da / a = 1.
   pose M1 := map_mx (divz^~ a) M; case/(_ M1 1)=> // [k|L uL [R uR [d dvD dM]]].
     by rewrite !mxE Da divzz nz_a.
@@ -1098,7 +1098,7 @@ case Zv: (map_mx denq (vv *m pinvmx T) == const_mx 1).
     rewrite {1}(coord_vbasis s_v); apply: eq_bigr => j _; congr (_ *: _).
     have ->: map_mx intr a = vv *m pinvmx T *m map_mx intr (dsubmx Gud).
       rewrite map_mxM /=; congr (_ *m _); apply/rowP=> i; rewrite 2!mxE numqE.
-      by have /eqP/rowP/(_ i) := Zv; rewrite !mxE => ->; rewrite mulr1.
+      by have /eqP/rowP/(_ i)/[!mxE]-> := Zv; rewrite mulr1.
     by rewrite -(mulmxA _ _ S) mulmxKpV ?mxE // -eqST submx_full.
   rewrite (coord_vbasis (s_Zs _)); apply: eq_bigr => j _; congr (_ *: _).
   rewrite linear_sum mxE; apply: eq_bigr => i _.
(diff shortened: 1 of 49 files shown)


gares:
doc is incomplete, the RHS is always the same

chdoc:
I'm a bit skeptical as it comes to the chosen syntax. Is `/[ ... ]` currently a valid intro pattern? If it is not, why not use:

- `=> /[rules]` : perform `rewrite rules` (where `rules` may include `!`, `?`, and  `-` as usual?)

Also, what's the expected behavior if there is a lemma or local assumption called `apply`, `dup`, or `swap`. Can one still use them as views?

CohenCyril:
@gares
> doc is incomplete, the RHS is always the same

I do not understand. What do you want to add to the doc?


@chdoc for now I will send you [here](https://github.com/math-comp/math-comp/pull/372#issuecomment-513786319), and on monday we could talk.

gares:
In the top post you say "foo does bar" but bar is always the same. It is clear what you mean, I was just warning to not copy that doc as is

CohenCyril:
> In the top post you say "foo does bar" but bar is always the same. It is clear what you mean, I was just warning to not copy that doc as is

Ah sh** I updated the commit but forget the PR OP...

soraros:
Maybe /apply etc. should be capitalized since they have predefined magical meaning?

gares:
I think such a change should be discussed at a meeting. I'm supportive but the notation design can only benefit from multiple scrutiny. For example the `/[1rule]` thing with no space and no `!` is kind of odd.

I'd like to see a few tests added to the test suite (and later backported to Coq), for example to test the rewrite notation with inlined multi-rules.

CohenCyril:
> For example the `/[1rule]` thing with no space and no `!` is kind of odd.

I agree we could have `/[1!rule]` instead, it would indeed look nicer!

mathcomp/algebra/intdiv.v
```diff
@@ -1060,7 +1060,7 @@ case Zv: (map_mx denq (vv *m pinvmx T) == const_mx 1).
     rewrite {1}(coord_vbasis s_v); apply: eq_bigr => j _; congr (_ *: _).
     have ->: map_mx intr a = vv *m pinvmx T *m map_mx intr (dsubmx Gud).
       rewrite map_mxM /=; congr (_ *m _); apply/rowP=> i; rewrite 2!mxE numqE.
-      by have /eqP/rowP/(_ i) := Zv; rewrite !mxE => ->; rewrite mulr1.
+      by have /eqP/rowP/(_ i)/[!mxE]-> := Zv; rewrite mulr1.
```
gares, line 1063: maybe a space after /rowP ?

CohenCyril: why?

gares: because `/v1/v2/../vn` has a meaning,  all views are applied and typechecked together, while `/(_ i)` an `/[!mxE]` are separate steps. I'd write
```coq
by have /eqP/rowP /(_ i) /[!mxE] -> := Zv; rewrite mulr1.
```
and eat the spaces only if they help being < 80

gares: Anyway, not really important...

chdoc:
After a cursory look, I'm not entirely convinced that replacing every occurrence of `; rewrite inE` with `=> /[1!inE]` is really that big an improvement, in particular when replacing `by apply: foo; rewrite inE` with `by apply: foo => /[1!inE]` (found some instances in `action.v`, but there are probably many others). IMHO, the `/[!inE]` pattern should mainly be used if it becomes part of another into pattern (or when necessary to stay within the 80 characters limit). 

@CohenCyril , what's your opinion or rationale for this?

CohenCyril:
@chdoc finally updated :)

chdoc (approved):
LGTM! :tada:

gares:
OK, I'm merging this one.
I wonder if the notation could be improved further to allow for `[!inE, other, ..]` since the `()` serve really little purpose (syntactically). But that would be for another PR.

gares (approved):

PR #514, opened 2020-05-21
Lemma addition to ssrnum

affeldt-aist:
#### Motivation for this change

A few lemmas about norm that we found useful in the context of the mathcomp-analysis project (among them: https://github.com/math-comp/analysis/blob/f8d361a93db5906a73327d4fcf1fe9070e1960b8/theories/normedtype.v#L90-L103) and which we think are better suited to mathcomp.

##### Things done/to do

### DIFF
--- a/CHANGELOG_UNRELEASED.md
+++ b/CHANGELOG_UNRELEASED.md
@@ -11,9 +11,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Added
 
 - in `ssrnum.v`, new lemmas:
-  + `{real_,}ltr_normlW`, `{real_,}ltrNnormlW`, `{real_,}ler_normlW`, `{real_,}lerNnormlW`
-  + `{real_,}ltr_distl_addr`, `{real_,}ler_distl_addr`, `{real_,}ltr_distl_addrC`, `{real_,}ler_distl_addrC`,
-    `{real_,}ltr_distl_subl`, `{real_,}ler_distl_subl`, `{real_,}ltr_distl_sublC`, `{real_,}ler_distl_sublC`
+  + `(real_)ltr_normlW`, `(real_)ltrNnormlW`, `(real_)ler_normlW`, `(real_)lerNnormlW`
+  + `(real_)ltr_distl_addr`, `(real_)ler_distl_addr`, `(real_)ltr_distl_addrC`, `(real_)ler_distl_addrC`,
+    `(real_)ltr_distl_subl`, `(real_)ler_distl_subl`, `(real_)ltr_distl_sublC`, `(real_)ler_distl_sublC`
 
 ### Changed
 
--- a/CHANGELOG_UNRELEASED.md
+++ b/CHANGELOG_UNRELEASED.md
@@ -11,10 +11,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Added
 
 - in `ssrnum.v`, new lemmas:
-  + `gt_norm_eqF`
-  + `{real_,}gtr_norm`, `{real_,}gtrNnorm`, `{real_,}ger_norm`, `{real_,}gerNnorm`
-  + `{real_,}ltr_dist_addl`, `{real_,}ler_dist_addl`, `{real_,}ltr_distr_addl`, `{real_,}ler_distr_addl`,
-    `{real_,}ltr_dist_subl`, `{real_,}ler_dist_subl`, `{real_,}ltr_distr_subr`, `{real_,}ler_distr_subr`
+  + `{real_,}ltr_normlW`, `{real_,}ltrNnormlW`, `{real_,}ler_normlW`, `{real_,}lerNnormlW`
+  + `{real_,}ltr_distl_addr`, `{real_,}ler_distl_addr`, `{real_,}ltr_distl_addrC`, `{real_,}ler_distl_addrC`,
+    `{real_,}ltr_distl_subl`, `{real_,}ler_distl_subl`, `{real_,}ltr_distl_sublC`, `{real_,}ler_distl_sublC`
 
 ### Changed
 
--- a/CHANGELOG_UNRELEASED.md
+++ b/CHANGELOG_UNRELEASED.md
@@ -10,9 +10,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Added
 
-- lemmas `ltr_dist_addl`, `ler_dist_addl`, `ltr_distr_addl`,
-  `ler_distr_addl`, `ltr_dist_subl`, `ler_dist_subl`,
-  `ltr_distr_subr`, `ler_distr_subr`, `norm_lt_eqF` in `ssrnum`
+- in `ssrnum.v`, new lemmas:
+  + `gt_norm_eqF`
+  + `{real_,}gtr_norm`, `{real_,}gtrNnorm`, `{real_,}ger_norm`, `{real_,}gerNnorm`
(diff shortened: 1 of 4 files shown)


mathcomp/algebra/ssralg.v
```diff
@@ -764,6 +764,9 @@ Proof. by rewrite opprD addrA addrK. Qed.
 Lemma subrKA z x y : (x - z) + (z + y) = x + y.
 Proof. by rewrite addrA addrNK. Qed.
 
+Lemma subr_trans z x y : (x - z) + (z - y) = x - y.
```
CohenCyril, line 767: This lemma was already generalized and added as `subrKA` (just above), math-comp analysis should use it instead now.

mathcomp/algebra/ssrnum.v
```diff
@@ -3754,6 +3754,12 @@ Proof. by rewrite lter_norml !lter_sub_addl. Qed.
 
 Definition lter_distl := (ler_distl, ltr_distl).
 
+Lemma ltr_distW x y e : `|x - y| < e -> y - e < x.
```
CohenCyril, line 3757: I believe the naming I picked in mathcomp analysis is wrong, and we should provide a more complete interface the following lemmas:
- `` `|x - y| < e -> y - e < x``
- `` `|x - y| < e -> x < y + e``
- `` `|x - y| < e -> y < x + e``
- same with `<=`

and the naming convention should encompass all these variants

CohenCyril: maybe something like `(gtr|ltr)_dist[lr]`? WDYT?

affeldt-aist: I couldn't make sense of it right away so I looked for an alternative (also for the sake of the conversation). What about?

- `ltr_dist_addl` : on the model of `ltr_subl_addl` because it looks the same
- `ltr_dist_subl` : `sub` instead of `add` (does not change the order between `x` and `y`)
- `ltr_distr_{addl,subl}` : because it is just a matter of reversing the dist

mathcomp/algebra/ssrnum.v
```diff
@@ -3754,6 +3754,36 @@ Proof. by rewrite lter_norml !lter_sub_addl. Qed.
...
+Lemma ler_dist_subl x y e : `|x - y| <= e -> x - e <= y.
+Proof. by move/ler_dist_addl; rewrite -ler_subl_addr. Qed.
+
+Lemma ltr_distr_subr x y e : `|x - y| < e -> y - e < x.
+Proof. by rewrite distrC => /ltr_dist_subl. Qed.
+
+Lemma ler_distr_subr x y e : `|x - y| <= e -> y - e <= x.
+Proof. by rewrite distrC => /ler_dist_subl. Qed.
+
+Lemma nlt_eqF (x y : R) : `|x| < y -> (x == - y = false) * (x == y = false).
```
CohenCyril, line 3781: The prefix `n` for `norm` does not exist yet, and I am not in favor if it. I suggest this is renamed `norm_lt_eqF`.

CohenCyril: @affeldt-aist I edited my comment by the mail you might have received was not. I prefer `gt_norm_eqF`

mathcomp/algebra/ssrnum.v
```diff
@@ -3754,6 +3754,36 @@ Proof. by rewrite lter_norml !lter_sub_addl. Qed.
...
+Lemma ler_dist_subl x y e : `|x - y| <= e -> x - e <= y.
+Proof. by move/ler_dist_addl; rewrite -ler_subl_addr. Qed.
+
+Lemma ltr_distr_subr x y e : `|x - y| < e -> y - e < x.
+Proof. by rewrite distrC => /ltr_dist_subl. Qed.
+
+Lemma ler_distr_subr x y e : `|x - y| <= e -> y - e <= x.
+Proof. by rewrite distrC => /ler_dist_subl. Qed.
+
+Lemma norm_lt_eqF (x y : R) : `|x| < y -> (x == - y = false) * (x == y = false).
```
CohenCyril, line 3781:
```suggestion
Lemma gt_norm_eqF (x y : R) : `|x| < y -> (x == - y = false) * (x == y = false).
```

mathcomp/algebra/ssrnum.v
```diff
@@ -3754,6 +3754,36 @@ Proof. by rewrite lter_norml !lter_sub_addl. Qed.
 
 Definition lter_distl := (ler_distl, ltr_distl).
 
+Lemma ltr_dist_addl x y e : `|x - y| < e -> x < y + e.
```
CohenCyril, line 3757: For consistency, we should also add
```coq
Lemma gtr_norm : `|x| < y -> x < y.
Lemma gtrNnorm : `|x| < y -> - y < x.
Lemma ger_norm : `|x| <= y -> x <= y.
Lemma gerNnorm : `|x| <= y -> - y <= x.
```
and all the `real_` variants...

mathcomp/algebra/ssrnum.v
```diff
@@ -3754,6 +3798,36 @@ Proof. by rewrite lter_norml !lter_sub_addl. Qed.
...
+Lemma ler_dist_subl x y e : `|x - y| <= e -> x - e <= y.
+Proof. exact: real_ler_dist_subl. Qed.
+
+Lemma ltr_distr_subr x y e : `|x - y| < e -> y - e < x.
+Proof. exact: real_ltr_distr_subr. Qed.
+
+Lemma ler_distr_subr x y e : `|x - y| <= e -> y - e <= x.
+Proof. exact: real_ler_distr_subr. Qed.
+
+Lemma gt_norm_eqF (x y : R) : `|x| < y -> (x == - y = false) * (x == y = false).
```
CohenCyril, line 3825: Actually I believe this one should be obtained by letting the user compose a `gtr_norm` lemma with `gt_eqF`, otherwise there are two many variants and multi-rules to take into account.

mathcomp/algebra/ssrnum.v
CohenCyril: I do not like the lack of symmetry between `dist` and `distr`, (the former should then be `distl`), but actually all of them must be `distl` because `dist` is on the left side of `<`, swapping arguments of `dist` should be replaced with the use of suffix `C`.

BTW I noticed a discrepancy introduced by this PR: 
- the current convention is that: `norm` on the right of `<` is `ltr_normr`
- the current convention is that: `norm` on the left of `<` is `ltr_norml`, while this PR introduces the convention `gt_norm`. I'd rather keep the old convention for `1.11` and do a global pass for `1.12`. In which case we cannot keep `gtr_norm` and we should go for e.g. `ltr_normlW`.

CHANGELOG_UNRELEASED.md
```diff
@@ -10,6 +10,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Added
 
+- in `ssrnum.v`, new lemmas:
+  + `{real_,}ltr_normlW`, `{real_,}ltrNnormlW`, `{real_,}ler_normlW`, `{real_,}lerNnormlW`
+  + `{real_,}ltr_distl_addr`, `{real_,}ler_distl_addr`, `{real_,}ltr_distl_addrC`, `{real_,}ler_distl_addrC`,
+    `{real_,}ltr_distl_subl`, `{real_,}ler_distl_subl`, `{real_,}ltr_distl_sublC`, `{real_,}ler_distl_sublC`
```
pi8027, line 16: Using `(real_)` rather than `{real_,}` would be consistent with the released changelog. See, e.g.: https://github.com/math-comp/math-comp/blob/master/CHANGELOG.md#renamed

affeldt-aist:
@CohenCyril Is the last change ok?

CohenCyril:
@affeldt-aist `C` should qualify `dist` nor `add` or `sub`

ybertot:
Thanks to authors and reviewers!

PR #516, opened 2020-05-30
Generalizing max and min to porderType

### LINKED ISSUE #515: max and min are too specific
From 1.10 to 1.11+beta1 we lost in generality in ``ssrnum.v``: indeed ``max`` and ``min`` where defined in a ``numDomainType``, but they are now defined in terms of  ``meet`` and ``join``, which requires at least lattice, which a ``numDomainType`` is not...

One possible fix is to reintroduce ``Order.max`` (and ``Order.min``) as ``if x <= y then y else x`` (and similar for min), and develop their theory in a partial order, when elements are all comparable between themselves. Then link them with meet and join when appropriate.

Another fix is to reintroduce ``Num.max`` and ``Num.min`` but this would be confusing if we keep ``Order.min`` and ``Order.max``.

In any case these definitions must match exactly the one that existed in 1.10.0 for backward compatiblity reasons and the equations given in the factories for total order (which should be updated if they do not match).

CohenCyril:

From 1.10 to 1.11+beta1, max and min had lost their generality, introducing a regression.
This PR redefines max and min with a general theory in order.v, and thus fixes #515.
It also fixes several potential bugs in their definition.

- max and min are not defined in terms or meet and join anymore
- extremum_inP is a generalization of extremum suitable for a partial order
- arg_min and arg_max are usable in a porderType
- equivalences between min and meet, and max and join are proven for orderType
- leP, ltP, ltgtP, and `[l]?comparable_.*P` lemmas have been generalized to take this into account
- total_display was completely removed
- `Order.max` and `Order.min`  are now convertible to `maxn` and `minn`
- Inserting `(fun _ _ => erefl)` in `LeOrderMixin` and `LtOrderMixin`
  gives `meet`/`join` which are convertible to `min`/`max`
- `Order.max` and `Order.min` are not convertible to former `Num.max` and `Num.min`
- min and max can now be used in a partial order (sometimes under preconditions)
- min and max can now be used in a `numDomainType` (sometimes under preconditions)
- the compatibility module has been fixed accordingly
- deprecating `fintype.arg_(min|max)P`
- removing dangling comments connecting min max and meet join
- better compatibility module
- removing broken notations with `\min` and `\max` (no neutral available)
- fixing `incompare` and `incomparel` in order.v
- adding missing elimination lemmas (`(comparable_)?(max|min)E[lg][et]`)
- adding missing `(comparable|real)_arg(min|max)P`

##### Things done/to do

- [x] added corresponding entries in `CHANGELOG_UNRELEASED.md` (explain the renaming)
- [x] added corresponding documentation in the headers (must add `min` and `max` in the header of `order.v`)

### DIFF
--- a/mathcomp/algebra/ssrnum.v
+++ b/mathcomp/algebra/ssrnum.v
@@ -389,10 +389,10 @@ Notation "@ 'lerif' R" := (@Order.leif ring_display R)
 Notation comparabler := (@Order.comparable ring_display _) (only parsing).
 Notation "@ 'comparabler' R" := (@Order.comparable ring_display R)
   (at level 10, R at level 8, only parsing) : fun_scope.
-Notation maxr := (@Order.join ring_display _).
+Notation maxr := (@Order.max ring_display _).
 Notation "@ 'maxr' R" := (@Order.join ring_display R)
     (at level 10, R at level 8, only parsing) : fun_scope.
-Notation minr := (@Order.meet ring_display _).
+Notation minr := (@Order.min ring_display _).
 Notation "@ 'minr' R" := (@Order.meet ring_display R)
   (at level 10, R at level 8, only parsing) : fun_scope.
 
@@ -1550,6 +1550,8 @@ Definition subr_cp0 := (subr_lte0, subr_gte0).
 
 (* Comparability in a numDomain *)
 
+Lemma comparable0r x : (0 >=< x)%R = (x \is Num.real). Proof. by []. Qed.
+
 Lemma comparabler0 x : (x >=< 0)%R = (x \is Num.real).
 Proof. by rewrite comparable_sym. Qed.
 
@@ -1676,11 +1678,7 @@ Lemma ler_leVge x y : x <= 0 -> y <= 0 -> (x <= y) || (y <= x).
 Proof. by rewrite -!oppr_ge0 => /(ger_leVge _) h /h; rewrite !ler_opp2. Qed.
 
 Lemma real_leVge x y : x \is real -> y \is real -> (x <= y) || (y <= x).
-Proof.
-rewrite !realE; have [x_ge0 _|x_nge0 /= x_le0] := boolP (_ <= _); last first.
-  by have [/(le_trans x_le0)->|_ /(ler_leVge x_le0) //] := boolP (0 <= _).
-by have [/(ger_leVge x_ge0)|_ /le_trans->] := boolP (0 <= _); rewrite ?orbT.
-Qed.
+Proof. by rewrite -comparabler0 -comparable0r => /comparabler_trans P/P. Qed.
 
 Lemma real_comparable x y : x \is real -> y \is real -> x >=< y.
 Proof. exact: real_leVge. Qed.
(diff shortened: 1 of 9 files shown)


CohenCyril:
@affeldt-aist see the TODO

mathcomp/ssreflect/order.v
```diff
@@ -1032,39 +1035,50 @@ Definition leif (x y : T) C : Prop := ((x <= y) * ((x == y) = C))%type.
...
-Variant lt_xor_ge (x y : T) : bool -> bool -> Set :=
-  | LtNotGe of x < y  : lt_xor_ge x y false true
-  | GeNotLt of y <= x : lt_xor_ge x y true false.
+Variant lt_xor_ge (x y : T) :
+  T -> T -> T -> T -> bool -> bool -> Set :=
+  | LtNotGe of x < y  : lt_xor_ge x y x x y y false true
+  | GeNotLt of y <= x : lt_xor_ge x y y y x x true false.
+
+Definition min x y := if x < y then x else y.
+Definition max x y := if x < y then y else x.
```
pi8027, line 1049: If we wish to make it compatible as much as possible with `minr` and `maxr` of version 1.10.0, shouldn't we use `<=` rather than `<`? https://github.com/math-comp/math-comp/blob/mathcomp-1.10.0/mathcomp/algebra/ssrnum.v#L256-L257

CohenCyril: `minn` and `maxn` are defined using `<` and `<=`, I'd rather loose backward compatibility than convertibility with nat definitions.

pi8027: OK. That's fine (as I said in a MC call).

mathcomp/algebra/ssrnum.v
```diff
@@ -332,10 +331,6 @@ Variables (T : Type) (cT : type).
 Let xT := let: Pack T _ := cT in T.
 Notation xclass := (class cT : class_of xT).
 
-(* Definition normedZmodType : normedZmodType cT := *)
-(*   @NormedZmodule.Pack *)
-(*      cT (Phant cT) cT *)
-(*      (NormedZmodule.Class (NumDomain.normed_mixin xclass)). *)
```
pi8027, line 338: I think this was left as a comment on purpose. See L.293 (and L.351 which you may also want to remove).

CohenCyril: :+1: I withdrew this deletion from the PR, since it was an accidental removal unrelated to the changes in this PR, maybe we can open an issue to document or remove this comment.

pi8027:
I hope I will give a detailed review next week if it is not too late for 1.11.0.

affeldt-aist:
@CohenCyril @pi8027 Are we confident that the bugfix is ok? (We saw that it was tested with real-closed at least). Since we are late for the release, we would like to merge. Is it ok to defer further reviews to later?

CohenCyril:
@affeldt-aist the bugfix is not okay until I finish porting math-comp/analysis, which is almost done.

CohenCyril:
@ybertot @affeldt-aist tested on analysis, this ready.

mathcomp/ssreflect/order.v
```diff
@@ -186,14 +188,14 @@ From mathcomp Require Import path fintype tuple bigop finset div prime.
...
+(* dvd_display (both explained above), ring_display (from algebra/ssrnum      *)
+(* to change the scope of the usual notations to ring_scope). We also provide *)
+(* lexi_display and prod_display for lexicographic and product order          *)
+(* respectively.                                                              *)
 (* The default display is tt and users can define their own as explained      *)
 (* above.                                                                     *)
 (*                                                                            *)
-(* For orderType we provide the following operations (in total_display)       *)
+(* For porderType we provide the following operations                         *)
 (*   [arg minr_(i < i0 | P) M] == a value i : T minimizing M : R, subject to  *)
```
CohenCyril, line 199:
```suggestion
(*   [arg min_(i < i0 | P) M] == a value i : T minimizing M : R, subject to  *)
```

mathcomp/ssreflect/order.v
```diff
@@ -196,7 +196,7 @@ From mathcomp Require Import path fintype tuple bigop finset div prime.
 (* above.                                                                     *)
 (*                                                                            *)
 (* For porderType we provide the following operations                         *)
-(*   [arg minr_(i < i0 | P) M] == a value i : T minimizing M : R, subject to  *)
+(*   [arg min_(i < i0 | P) M] == a value i : T minimizing M : R, subject to   *)
 (*                      the condition P (i may appear in P and M), and        *)
 (*                      provided P holds for i0.                              *)
 (*   [arg maxr_(i > i0 | P) M] == a value i maximizing M subject to P and     *)
```
CohenCyril, line 202:
```suggestion
(*   [arg max_(i > i0 | P) M] == a value i maximizing M subject to P and      *)
```

PR #536, opened 2020-06-15
Fix some hierarchy.ml related issues

pi8027:

- Fix some issues reported by @strub:
  - pass `Unix.environment ()` to `coqtop` to preserve the parent process environment,
  - check the exit status of `coqtop` and report an error if it is wrong.
- Fix some `Makefile` issues:
  - rename `hierarchy_test.v` to `test_hierarchy_all.v` because other test cases are named `test_*.v` and we may also have `test_hierarchy_ssreflect.v`, `test_hierarchy_fingroup.v`, etc. (but I actually had some issues to implement this feature in the `Makefile`),
  - since the `test_suite` directory exists from the beginning, we do not need to do `mkdir -p test_suite`,
  - ~the `build` prerequisite of `test_suite/test_hierarchy_all.v` should be order only (right?),~
  - fix `make test_suite/*.vo` by adding a rule.

##### Things done/to do

- ~added corresponding entries in `CHANGELOG_UNRELEASED.md`~

### DIFF
--- a/mathcomp/Makefile.common
+++ b/mathcomp/Makefile.common
@@ -56,11 +56,10 @@ Makefile.coq: pre-makefile $(COQPROJECT) Makefile
 
 # Test suite ---------------------------------------------------------
 
-test_suite/hierarchy_test.v: build
-	mkdir -p test_suite
-	COQBIN=$(COQBIN) ocaml ../etc/utils/hierarchy.ml -verify -R . mathcomp -lib all.all > test_suite/hierarchy_test.v
+test_suite/test_hierarchy_all.v: build
+	COQBIN=$(COQBIN) ocaml ../etc/utils/hierarchy.ml -verify -R . mathcomp -lib all.all > test_suite/test_hierarchy_all.v
 
-Makefile.test-suite.coq: test_suite/hierarchy_test.v
+Makefile.test-suite.coq: test_suite/test_hierarchy_all.v
 	$(COQMAKEFILE) $(COQMAKEFILEOPTIONS) -f Make.test-suite -o Makefile.test-suite.coq
 
 # Global config, build, clean and distclean --------------------------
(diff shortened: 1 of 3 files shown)


CohenCyril:
@pi8027 thanks for addressing these issues!

> * the `build` prerequisite of `test_suite/test_hierarchy_all.v` should be order only (right?),

I did not understand this sentence, what do you mean by *order on*?

mathcomp/Makefile.common
```diff
@@ -56,11 +56,10 @@ Makefile.coq: pre-makefile $(COQPROJECT) Makefile
 
 # Test suite ---------------------------------------------------------
 
-test_suite/hierarchy_test.v: build
-	mkdir -p test_suite
-	COQBIN=$(COQBIN) ocaml ../etc/utils/hierarchy.ml -verify -R . mathcomp -lib all.all > test_suite/hierarchy_test.v
+test_suite/test_hierarchy_all.v: | build
```
pi8027, line 59: @CohenCyril I meant "order-only prerequisites" in [this documentation](https://www.gnu.org/software/make/manual/html_node/Prerequisite-Types.html). If we remove this pipe symbol, repeating `make test-suite` forces us to rerun the test suite.

pi8027: After re-reading the documentation, I think this is not the right solution. Sorry for the mess.

CohenCyril:
@pi8027 is this PR ready for reviews?

pi8027:
@CohenCyril Yes.

CohenCyril:
@maximedenes could you have a look or would you prefer me to ask another review?

maximedenes:
> could you have a look or would you prefer me to ask another review?

@CohenCyril Sorry, this went completely below my radar.

PR #548, opened 2020-07-16
Use predicators even extensively

pi8027:

This PR attempts to restate several lemmas using predicators such as `cancel`, `(left|right)_(id|zero|inverse|distributive)`, `idempotent`, `associative`, and `commutative`. Some of them are probably too aggressive and I would like to have comments on this point.

##### Things done/to do

  - If the ordering of arguments are changed, it has to be mentioned explicitly.

### DIFF
--- a/mathcomp/algebra/intdiv.v
+++ b/mathcomp/algebra/intdiv.v
@@ -89,10 +89,8 @@ Proof.
 by rewrite {3}[d]intEsign !mulr_sign; case: ifP => -> //; rewrite divzN opprK.
 Qed.
 
-Lemma div0z d : (0 %/ d)%Z = 0.
-Proof.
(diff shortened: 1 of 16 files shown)


CohenCyril:
@pi8027 I'm a bit hesitant about that one, it should be discussed at the MC meeting and go hand in hand with an improvement of Coq search tool to index the delta-expansion of certain statments...

pi8027:
@CohenCyril I agree. This PR also breaks odd-order by changing argument scopes of some lemmas, but putting all the scope information back by `Arguments` command seems too much to do.

pi8027:
For the record, there was a discussion about this PR in [the last MC call](https://github.com/math-comp/math-comp/wiki/Minutes-September-23-2020).

PR #553, opened 2020-08-11
fix non-reversible-notation warnings

chdoc:

fix `non-reversible-notation` warnings in `ssrAC.v` caused by Ltac in notations. 

##### Things done/to do


chdoc:
@CohenCyril I just realized that this PR subsumes #527 . I needed to add some more instanced of `(only parsing)` though, to get rid of all warnings.

CohenCyril:
I do not think it subsumes #527 which also added a scope to the notations...

chdoc:
> I do not think it subsumes #527 which also added a scope to the notations...

Indeed, I overlooked the scope annotations. Still, there are more than just the three notations modified in #527 that cause warnings. Feel free to adapt as you see fit.

CohenCyril:
@chdoc I will merge this one, and we can adapt #527 later (there was an issue with the rescoping)

PR #207, opened 2018-07-27
adding forms and spectral theorems

CohenCyril:
Content:
- Adding bilinear, sesquilinear, hermitian (symmetric, skew symmetric and hermitian) and hermitian positive definite forms.
- A proof of the spectral theorem for normal matrices, and some theory of stability wrt a sub(mx)space.
- Several auxiliary lemmas of general purpose.

TODO:
- [x] document,
- [ ] replace ad hoc `cfdot` in classfun (@LaurenceRideau),
- [x] backport several theorems to matrix, vector or mxalgebra,
- [ ] bridge vector and mxalgebra to port results.
- [x] update ChangeLog
- [ ] clear dependencies

TODO update after @LaurenceRideau tried on the odd-order repo (https://github.com/math-comp/odd-order/pull/2)
- [ ] Infer the canonical form on a vector space in notations such as `'[_ ,_]`, `orthogonal`, `isometry`...
- [ ] Make the right `linear_for` on the right and the left for bilinear forms (so that `'[a *: x, b *: y] = a * b^* * '[x ,y]` instead of `'[a *: x, b *: y] = a *: b^* *: '[x ,y]`)
- [x] Have lemmas `linearZl`, `linearZr`, `linearZlr`, `linearZrl` (right now only `linearZl_LR` is only available)

PR #558, opened 2020-09-01
Adding allrel predicate

CohenCyril:

This predicate is useful in stating that a sequence of matrices all commute.
This is a part of #207.

##### Things done/to do

### DIFF
--- a/CHANGELOG_UNRELEASED.md
+++ b/CHANGELOG_UNRELEASED.md
@@ -42,6 +42,14 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   coefficient.
 
 - in `poly.v`, new lemma `commr_horner`.
+- in `seq.v`, new lemma `mkseqP` to abstract a sequence `s` with
+  `mkseq f n`, where `f` and `n` are fresh variables.
+
+- in `seq.v`, new high-order predicate `allrel r s` which
+  asserts that a relation `r` holds on all pairs of elements of `s`, and
+  + lemmas `allrel_map`, `allrelP` and `allrel0`.
+  + lemmas `allrel1`, `allrel2` and `allrel_cons`,
+    under assumptions of reflexivity and symmetry of `r`.
 
 ### Changed
 
--- a/mathcomp/ssreflect/seq.v
+++ b/mathcomp/ssreflect/seq.v
@@ -132,6 +132,8 @@ From mathcomp Require Import ssreflect ssrfun ssrbool eqtype ssrnat.
 (*                    i.e. self expanding definition for                      *)
 (*                      [seq f x y | x <- s, y <- t]                          *)
 (*               := [:: f x_1 y_1; ...; f x_1 y_m; f x_2 y_1; ...; f x_n y_m] *)
+(*     allrel r s := all id [seq r x y | x <- xs, y <- xs]                    *)
+(*                == the proposition r x y holds for all possible x, y in xs  *)
 (*        map f s == the sequence [:: f x_1, ..., f x_n].                     *)
 (*      pmap pf s == the sequence [:: y_i1, ..., y_ik] where i1 < ... < ik,   *)
 (*                   pf x_i = Some y_i, and pf x_j = None iff j is not in     *)
@@ -2474,6 +2476,12 @@ Proof.
 by apply: (@eq_from_nth _ x0); rewrite size_mkseq // => i Hi; rewrite nth_mkseq.
 Qed.
 
+Variant mkseq_spec s : seq T -> Type :=
+| MapIota n f : s = mkseq f n -> mkseq_spec s (mkseq f n).
+
+Lemma mkseqP s : mkseq_spec s s.
+Proof. by rewrite -[s]mkseq_nth; constructor. Qed.
+
 End MakeSeq.
 
 Section MakeEqSeq.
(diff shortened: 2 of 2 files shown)


CohenCyril:
@gares, @pi8027, @strub can one of you assign themselves?

gares:
It looks ok to me, but I'm not in love with the `are_` part of the name. Is it similar to something already existing?

mathcomp/ssreflect/seq.v
```diff
@@ -3138,6 +3140,52 @@ Arguments allpairsP {S T R f s t z}.
 Arguments perm_nilP {T s}.
 Arguments perm_consP {T x s t}.
 
+Section AreAllPairs.
+
+Definition are_allpairs {T : Type} (r : rel T) xs :=
```
chdoc, line 3145: I agree with @gares in not being happy with this name.

mathcomp/ssreflect/seq.v
```diff
@@ -3138,6 +3140,52 @@ Arguments allpairsP {S T R f s t z}.
...
+Proof. by rewrite /are_allpairs/= !rxx [r y x]rsym !(andbT, andbb). Qed.
+
+Lemma are_allpairs_cons x xs :
+  are_allpairs r (x :: xs) = all (r x) xs && are_allpairs r xs.
+Proof.
+pose f := nth x xs; have -> : xs = [seq f i | i <- iota 0 (size xs)].
+  apply: (@eq_from_nth _ x); first by rewrite size_map size_iota.
+  by move=> i i_lt; rewrite (nth_map 0) ?size_iota// nth_iota.
+have -> : x = f (size xs).+1 by rewrite /f nth_default.
+move: (size xs) (iota 0 _) => k s; rewrite -map_cons !are_allpairs_map all_map.
```
chdoc, line 3176: The more natural generalization would probably be `_.+1` rather than `size xs`

mathcomp/ssreflect/seq.v
```diff
@@ -3138,6 +3140,52 @@ Arguments allpairsP {S T R f s t z}.
...
+rewrite /are_allpairs/= ?rxx/= all_cat all_map; have [xxs/=|//] := boolP (all _ s).
```
chdoc, line 3177: This line has 83 characters, I suggest splitting it at the semicolon :grin:

chdoc (changes requested):

mathcomp/ssreflect/seq.v
gares: @CohenCyril probably took my vscode configuration...

chdoc:
Also, is it useful to have `T : nonPropType` rather than `T : eqType` while requiring `r : rel T` to be a boolean relation? I mean, the proof of `are_allpairs_cons` is nifty, but I would have expected a simpler proof based in `are_allpairsP`.

mathcomp/ssreflect/seq.v
CohenCyril: > Also, is it useful to have `T : nonPropType` rather than `T : eqType` while requiring `r : rel T` to be a boolean relation? I mean, the proof of `are_allpairs_cons` is nifty, but I would have expected a simpler proof based in `are_allpairsP`.

It may happen that a relation is decidable even on some non discrete types (e.g. a pair of an ordered type with a no discrete type), and I would like to preserve the full generality of the theorem on such occurences.

CohenCyril:
> Also, is it useful to have `T : nonPropType` rather than `T : eqType` while requiring `r : rel T` to be a boolean relation? I mean, the proof of `are_allpairs_cons` is nifty, but I would have expected a simpler proof based in `are_allpairsP`.

It may happen that a relation is decidable even on some non discrete types (e.g. a pair of an ordered type with a no discrete type), and I would like to preserve the full generality of the theorem on such occurences.

CohenCyril:
> It looks ok to me, but I'm not in love with the `are_` part of the name. Is it similar to something already existing?

I went back and forth a lot in the naming (the working name was the ugly `all11` (hence the 80 character overflow caused by sed)...
I am open to suggestions on the naming but I was pretty happy with this one being meaningful ("are all pairs satisfying r on s") and exposing the implementation using `allpairs` in the name itself..

CohenCyril:
Actually I would have gone for `is` except `are` is grammatically more accurate (since it is about a relation, not a predicate).

chdoc:
Well, the predicate states that "all pairs of `s` are _in_ the relation `r`" (using the usual mathematical speak of relations as sets/predicates). So how about `allpairs_in r s`? For this one even the partial application `allpairs_in r` seems reasonable to me. Unfortunately `in` is used differently elsewhere.

chdoc:
In any event, `are_allpairs r s` is a proposition not a question. So even `allpairs_are r s` ("all pairs _are_ r(-related)") seems preferable to me.

gares:
> So even `allpairs_are r s` ("all pairs _are_ r(-related)") seems preferable to me.

+1

To me the thing is messy because `allpairs` does not really build pairs but steals the name. If we had a `pairs` thing that builds the sequence of all pairs, we could use allpairs for your new concept, eg `allpairs r xs := all (fixup r) (pairs xs xs)`.

chdoc:
> To me the thing is messy because `allpairs` does not really build pairs but steals the name.

While the string "allpairs" appears in a considerable number of of lemmas, the definition `allpairs` seems to be virtually unused, because `allpairs` is (almost) always inlined. The only lemma about `allpairs` in `ssreflect` and `algebra` appears to be `allpairs_tupleP`. :thinking:

CohenCyril:
> > To me the thing is messy because `allpairs` does not really build pairs but steals the name.
> 
> While the string "allpairs" appears in a considerable number of of lemmas, the definition `allpairs` seems to be virtually unused, because `allpairs` is (almost) always inlined. The only lemma about `allpairs` in `ssreflect` and `algebra` appears to be `allpairs_tupleP`.

`allpairs` actually immediately unfolds to the idiom `[seq f x y | x <- s, y <- t x]` which gives its name to the lemmas, and it is a rather historic convention (the self-expanding definition is newer as far as I remember). Moreover it indeed does not form pairs but applies a function symbol `f` to them (and this function symbol can be dependent), sometimes it's a `pair`, sometimes a `sigT` and sometimes something else, so it makes it more flexible to let `f` variable.

I'm all in favor of `allpairs_are`anyway!

gares:
I saw that, allpairs maps a function overs all the pairs, which could be named, to me, map_allpairs. Anyway, I'm OK with the renaming proposed by @chdoc

CohenCyril:
I fixed the names and introduced the `mkseqP` idiom.

mathcomp/ssreflect/seq.v
```diff
@@ -2476,6 +2478,13 @@ Qed.
 
 End MakeSeq.
 
+Variant mkseq_spec T (s : seq T) : seq T -> Type :=
+| NoMapIota : s = [::] -> mkseq_spec s [::]
+| MapIota n f : s = mkseq f n -> mkseq_spec s (mkseq f n).
+
+Lemma mkseqP T (s : seq T) : mkseq_spec s s.
+Proof. by case: s => [|x s]; rewrite -1?[x :: s](mkseq_nth x); constructor. Qed.
```
chdoc, line 2486: Do you think that generating the `s = [::]` case, which reappears a as a subcase of the `s = mkseq f n` case preferable to requiring some `x0 : T`? If so, maybe replace `mkseq f n` with `mkseq f n.+1`, do avoid the case duplication?

CohenCyril: Actually, no I'm not sure... and having `n` looks dangerous `(iota 0 n.+1)` might expand and the point of `mkseqP` was to abstract away these... Maybe I will switch to requiring `x0` here...

CohenCyril: @chdoc looks better now I think

CohenCyril:
I have a new name suggestion: how about `allrel`?

chdoc:
> I have a new name suggestion: how about `allrel`?

I approve, in particular because it avoids the `_`, leading to nicer lemma names.

CHANGELOG_UNRELEASED.md
```diff
@@ -30,6 +30,15 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - in `bigop.v` new lemma `sig_big_dep`, analogous to `pair_big_dep`
   but with an additional dependency in the index types `I` and `J`.
 
+- in `seq.v`, new lemma `mkseqP` to abstract a sequence `s` with
+  `mkseq f n`, where `f` and `n` are fresh variables.
+
+- in `seq.v`, new high-order predicate `allrel r s` which
+  asserts a relation `r` holds on all pairs of elements of `s`, and
```
chdoc, line 37: which asserts _that_ / asserting _that_

chdoc (approved):
LGTM (apart from the grammar error in the changelog :smirk: )

CohenCyril:
@gare all good?

PR #573, opened 2020-09-02
Polynomial evaluation and minimal/char poly of a diagonal/triangular matrices

CohenCyril:

This PR states some essential properties about triangular and diagonal matrices. In particular, it exhibits the characteristic polynomial of triangular matrices, the minimal polynomial of diagonal matrices, and explain how to evaluate a polynomial in a diagonal matrix.

Part of #207.

##### Things done/to do

### DIFF
--- a/mathcomp/algebra/matrix.v
+++ b/mathcomp/algebra/matrix.v
@@ -2603,6 +2603,16 @@ Lemma det_lblock n1 n2 Aul (Adl : 'M[R]_(n2, n1)) Adr :
   \det (block_mx Aul 0 Adl Adr) = \det Aul * \det Adr.
 Proof. by rewrite -det_tr tr_block_mx trmx0 det_ublock !det_tr. Qed.
 
+Lemma det_trig n (A : 'M[R]_n) : is_trig_mx A -> \det A = \prod_(i < n) A i i.
+Proof.
+elim: n => [|n IHn] in A * => /is_trig_mxP Atrig; rewrite ?big_ord0 ?det_mx00//.
+rewrite -[n.+1]/(1 + n)%N in A Atrig *; rewrite big_ord_recl -[A in LHS]submxK.
+have -> : ursubmx A = 0 by apply/matrixP => i j; rewrite !mxE Atrig//= ord1.
+rewrite det_lblock IHn; last by apply/is_trig_mxP => *; rewrite !mxE Atrig.
+rewrite [ulsubmx A]mx11_scalar det_scalar1 !mxE !lshift0; congr (_ * _).
+by apply: eq_bigr => i; rewrite !mxE !rshift1.
+Qed.
+
 End ComMatrix.
 
 Arguments lin_mul_row {R m n} u.
--- a/mathcomp/algebra/matrix.v
+++ b/mathcomp/algebra/matrix.v
@@ -1421,7 +1421,7 @@ Lemma trigmx_ind (P : forall m n, 'M_(m, n) -> Type) :
     P m n A -> P (1 + m)%N (1 + n)%N (block_mx x 0 c A)) ->
   forall m n A, is_trig_mx A -> P m n A.
 Proof.
-move=> P0l P0r PS m n A; elim: A => [{}m|{}n|{}m {}n xx r c] A PA;
+move=> P0l P0r PS m n A; elim: A => {m n} [m|n|m n xx r c] A PA;
   do ?by rewrite (flatmx0, thinmx0); by [apply: P0l|apply: P0r].
 by rewrite is_trig_block_mx => // /and3P[/eqP-> _ Atrig]; apply: PS (PA _).
 Qed.
@@ -1430,7 +1430,7 @@ Lemma trigsqmx_ind (P : forall n, 'M[V]_n -> Type) : (P 0%N 0) ->
   (forall n x c A, is_trig_mx A -> P n A -> P (1 + n)%N (block_mx x 0 c A)) ->
   forall n A, is_trig_mx A -> P n A.
 Proof.
-move=> P0 PS n A; elim/sqmx_ind: A => [|{}n x r c] A PA.
+move=> P0 PS n A; elim/sqmx_ind: A => {n} [|n x r c] A PA.
   by rewrite thinmx0; apply: P0.
 by rewrite is_trig_block_mx => // /and3P[/eqP-> _ Atrig]; apply: PS (PA _).
 Qed.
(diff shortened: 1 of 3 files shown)


mathcomp/algebra/matrix.v
```diff
@@ -2560,6 +2560,16 @@ Lemma det_lblock n1 n2 Aul (Adl : 'M[R]_(n2, n1)) Adr :
   \det (block_mx Aul 0 Adl Adr) = \det Aul * \det Adr.
 Proof. by rewrite -det_tr tr_block_mx trmx0 det_ublock !det_tr. Qed.
 
+Lemma det_trig n (A : 'M[R]_n) : is_trig_mx A -> \det A = \prod_(i < n) A i i.
+Proof.
+elim: n => [|n IHn] in A * => /is_trig_mxP Atrig; rewrite ?big_ord0 ?det_mx00//.
+rewrite -[n.+1]/(1 + n)%N in A Atrig *; rewrite big_ord_recl -[A in LHS]submxK.
+have -> : ursubmx A = 0 by apply/matrixP => i j; rewrite !mxE Atrig//= ord1.
```
thery, line 2567: this seems a typical proof by induction on triangular matrices.
Maybe ursubmx A = 0 and is_triang_mx (lrsubmx A)
could be usefull theorems to have in the library.
Same holds for diagonal matrices.

Also I have noticed that there is no doc for is_triang_mx and is_diag_mx in the header of matrix

mathcomp/algebra/matrix.v
```diff
@@ -2560,6 +2560,16 @@ Lemma det_lblock n1 n2 Aul (Adl : 'M[R]_(n2, n1)) Adr :
   \det (block_mx Aul 0 Adl Adr) = \det Aul * \det Adr.
 Proof. by rewrite -det_tr tr_block_mx trmx0 det_ublock !det_tr. Qed.
 
+Lemma det_trig n (A : 'M[R]_n) : is_trig_mx A -> \det A = \prod_(i < n) A i i.
+Proof.
+elim: n => [|n IHn] in A * => /is_trig_mxP Atrig; rewrite ?big_ord0 ?det_mx00//.
```
thery, line 2565: in the file matrix, the style seems to have spaces before a //

thery (changes requested):

mathcomp/algebra/matrix.v
CohenCyril: > Maybe ursubmx A = 0 and is_triang_mx (lrsubmx A)

Do you mean we should make it an idiom?

> Maybe ursubmx A = 0 and is_triang_mx (lrsubmx A)

I could extract them indeed.

> Same holds for diagonal matrices.

Since `is_diag_mx A -> is_trig_mx A` holds, it may not be necessary... do you think we should add all variants for diagonal matrices.

> Also I have noticed that there is no doc for is_triang_mx and is_diag_mx in the header of matrix

I thought I did it in #571... my bad...
￼

mathcomp/algebra/matrix.v
CohenCyril: I think the new style conventions depend on whether `//` is related to one `?id` or several... In this case, I believe you are right and we should put a spacE.

thery: too complicated for my simple brain. You mean ?thm // and ?thm// will not have the same semantic?

mathcomp/algebra/matrix.v
thery: Everyone that will do a proof by induction on the size of a triangular or a diagonal matrix will have to reprove this kind  of theorems. If you turn them into lemma, I think your proof of `det_trig` will become elementary.

CohenCyril: I tried to split & add many subresults. Can you tell me what you think?

mathcomp/algebra/matrix.v
CohenCyril: I mean that, even if the formal semantic is exactly the same, by convention: if `thm` alone spawned sublemmas then one may write `?thm//`, but if it is the combination of  `thm1` and `thm2` then we write `?thm1 ?thm2 //`.

thery: very strange to me. Is someone will go to all the code to adopt this convention?

mathcomp/algebra/matrix.v
thery: Initially I was thinking only about the structure lemmas : how is_diag and is_trig propagates to submatrix but maybe adding ad-hoc elimination theorems is ok too. I am not sure there are lot of them in the library.

CohenCyril: > Initially I was thinking only about the structure lemmas : how is_diag and is_trig propagates to submatrix

That's what I thought and did initially, but as I thought, it barely removed one line of proof. In order to become "elementary" I also needed to add the induction. There are a few ad-hoc theorem (e.g. `poly_ind`, `big_ind` and `last_ind`), but it is indeed not a frequent practice, and it is preferred to have "side loading lemmas" instead (e.g. `ubnP`), but in the case of matrices the main enemy is the conversion between `n.+1` and `1 + n`, which I do not know how to deal with as a "side" lemma.

mathcomp/algebra/matrix.v
CohenCyril: I don't know. Do you think I should backtrack on that?

CohenCyril: Some files already started adopting it. Maybe the PR can be accepted independently of this.
We can discuss this convention in a meeting, and then fix it uniformly (or not) depending on the outcome of the discussions.

mathcomp/algebra/matrix.v
thery: So maybe we should not include the induction lemmas.

mathcomp/algebra/matrix.v
thery: My fear is that this non-uniformity will remain but fine with me to accept the PR once we've decided about the inclusion of the induction theorems

mathcomp/algebra/matrix.v
CohenCyril: I was going to conclude the opposite :laughing:

thery: ok let's keep them

thery (approved):

mathcomp/algebra/matrix.v
CohenCyril: I opened issue #580 and I will put it in the list of topics.

PR #572, opened 2020-09-02
Lemmas reindex_omap and bigD1_ord

CohenCyril:
##### Motivation for this change

- New lemma `reindex_omap` generalizes `reindex_onto` to the case
where the inverse function to `h` is partial (i.e. with codomain
`option J`, to cope with a potentially empty `J`.
- New lemma `bigD1_ord` takes out an element in the middle of a
`\big_(i < n)` and reindexes the remainder using `lift`.
- `eq_liftF` and `lift_eqF`
- proof simplificaions

Part of #207

##### Things done/to do

- [x] added corresponding entries in `CHANGELOG_UNRELEASED.md` (do not edit former entries)
- [ ] after merge: open issue to remove the coq < 8.10 compat code.
- ~added corresponding documentation in the headers~

##### Automatic note to reviewers

Read [this Checklist](https://github.com/math-comp/math-comp/wiki/Checklist-for-following,-reviewing-and-playing-with-a-PR#checklist-for-reviewing-a-pr) and make sure there is a milestone.

PR #569, opened 2020-09-02
Extracting a nonzero coefficient from a nonzero matrix

CohenCyril:
##### Motivation for this change

Skips a potentially painful pick step to grab a non zero coefficient
of a nonzero matrix.
Part of #207.

##### Things done/to do

- [x] added corresponding entries in `CHANGELOG_UNRELEASED.md` (do not edit former entries)
- ~added corresponding documentation in the headers~

##### Automatic note to reviewers

Read [this Checklist](https://github.com/math-comp/math-comp/wiki/Checklist-for-following,-reviewing-and-playing-with-a-PR#checklist-for-reviewing-a-pr) and make sure there is a milestone.

PR #574, opened 2020-09-03
Submatrix theory

CohenCyril:

This PR introduces arbitrary sub/reindexed - matrices, and provides a correspondance to all other specific instances in the library (`row`, `row'`, `xrow`, ...). This is an important prerequisite to the theory of minors. We also provide lemma `row_freePn` which states that in a row non-free matrix, one line is a linear combination of the others (i.e. `(row i A <= row' i A)%MS`).

Part of #207.

##### Things done/to do

- [x] Merge dependency: #569
- [x] Merge dependency: #570
- [x] Merge dependency: #572

### DIFF
--- a/mathcomp/algebra/matrix.v
+++ b/mathcomp/algebra/matrix.v
@@ -54,6 +54,13 @@ From mathcomp Require Import div prime binomial ssralg finalg zmodp countalg.
 (*                   should be decomposed.                                    *)
 (*   [u|d]submx A == the up/down submatrices of a column block matrix A.      *)
 (* [u|d][l|r]submx A == the upper left, etc submatrices of a block matrix A.  *)
+(*    mxsub f g A == generic reordered submatrix, given by functions f and g  *)
+(*                   which specify which subset of rows and columns to take   *)
+(*                   and how to reorder them, e.g. picking f and g to be      *)
+(*                   increasing yields traditional submatrices.               *)
+(*                := \matrix_(i, j) A (f i) (g i)                             *)
+(*     rowsub f A := mxsub f id A                                             *)
+(*     colsub g A := mxsub id g A                                             *)
 (* castmx eq_mn A == A : 'M_(m, n) cast to 'M_(m', n') using the equation     *)
 (*                   pair eq_mn : (m = m') * (n = n'). This is the usual      *)
 (*                   workaround for the syntactic limitations of dependent    *)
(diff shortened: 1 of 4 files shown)


mathcomp/algebra/matrix.v
```diff
@@ -54,6 +54,10 @@ From mathcomp Require Import div prime binomial ssralg finalg zmodp countalg.
 (*                   should be decomposed.                                    *)
 (*   [u|d]submx A == the up/down submatrices of a column block matrix A.      *)
 (* [u|d][l|r]submx A == the upper left, etc submatrices of a block matrix A.  *)
+(*    mxsub f g A == generic submatrix given by "subset functions" f and g    *)
```
thery, line 57: I think the name is a bit misleading.
In math, a submatrix is when you delete some lines or 
columns. Here you can do much more.

CohenCyril: The rationale is that `mxsub` will essentially be used with invective (or even increasing) `f` and `g`, but **numerous** (if not almost all) theorems still apply for arbitrary `f`, `g`, so it is a shame to have a separate definition and even more a shame to have theorems that require a spurious injectivity (or monotonicity) hypothesis.

CohenCyril: PS: think of it as `sort`, if you don't give it a proper order relation, it still does something, but it has less properties (it still is parametric though)

thery: yes and no I notive that when I saw `row_perm s = rowsub s`

CohenCyril: > yes and no I notive that when I saw `row_perm s = rowsub s`

this is a curiosity indeed, I am not sure how import this equation is, but I think it should still be in the library. The definition of submatrix is relaxed to (potentially trivial) submatrix, (with potential reordering), even though it does not matter in current practical use cases.

CohenCyril: Moreover I vaguely recall this might be the definition that @ggonthier agreed with, when @vsiles, @maximedenes  and @mortberg were writing the [theory of minors in CoqEAL](https://github.com/CoqEAL/CoqEAL/blob/master/theory/minor.v) (that I wish to end up importing, renaming and completing)

thery: so at least you should change the doc 
```mxsub f g A == generic submatrix  ...```
by
```mxsub f g A == generic sub/reindexed matrix  ...```

CohenCyril: ok

CohenCyril: done.

thery:
With all these force pushed I am a bit lost here. Is there anything that prevents to merge this one?

CohenCyril:
> With all these force pushed I am a bit lost here. Is there anything that prevents to merge this one?

Yes, many merges happened, and I keep rebasing...

CohenCyril:
@thery I rebased again.

CohenCyril:
Now the error is that gitlab CI is saturated with jobs (@gares that would be a good enough reason for me to just drop Coq < 8.10 support :wink:)

PR #575, opened 2020-09-03
Adding mxOver predicate

CohenCyril:

Analogous to `polyOver`, this will be essential when characterizing
real matrices among complex ones.

Part of #207

##### Things done/to do

### DIFF
--- a/mathcomp/algebra/matrix.v
+++ b/mathcomp/algebra/matrix.v
@@ -100,6 +100,7 @@ From mathcomp Require Import div prime binomial ssralg finalg zmodp countalg.
 (*         \adj A == the adjugate matrix of A (\adj A i j = cofactor j i A).  *)
 (*   A \in unitmx == A is invertible (R must be a comUnitRingType).           *)
 (*        invmx A == the inverse matrix of A if A \in unitmx A, otherwise A.  *)
+(* A \is a mxOver S == the matrix A has its coefficients in S.                *)
 (* The following operations provide a correspondence between linear functions *)
 (* and matrices:                                                              *)
 (*     lin1_mx f == the m x n matrix that emulates via right product          *)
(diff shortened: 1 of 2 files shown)


mathcomp/algebra/matrix.v
```diff
@@ -156,7 +157,7 @@ Reserved Notation "''cV[' R ]_ n"   (at level 8, n at level 2). (* only parsing
 Reserved Notation "''M[' R ]_ ( n )"     (at level 8). (* only parsing *)
 Reserved Notation "''M[' R ]_ ( m , n )" (at level 8). (* only parsing *)
 
-Reserved Notation "\matrix_ i E" 
+Reserved Notation "\matrix_ i E"
```
thery, line 160: could we fix all these spaces once and for the all the files in a separate commit?

mathcomp/algebra/matrix.v
```diff
@@ -100,6 +100,7 @@ From mathcomp Require Import div prime binomial ssralg finalg zmodp countalg.
 (*         \adj A == the adjugate matrix of A (\adj A i j = cofactor j i A).  *)
 (*   A \in unitmx == A is invertible (R must be a comUnitRingType).           *)
 (*        invmx A == the inverse matrix of A if A \in unitmx A, otherwise A.  *)
+(* A \is a mxOver S == the matrix A has its coefficients in S.                *)
```
thery, line 103: I would have expected A \in .mxOver S like A \in unitmx

thery (commented):
it seems ok

mathcomp/algebra/matrix.v
CohenCyril: I keep doing this, and some commits keep reintroducing them... at least doing it bit by bit triggers less conflicts...

mathcomp/algebra/matrix.v
CohenCyril: It's more akin to https://github.com/math-comp/math-comp/blob/master/mathcomp/algebra/poly.v#L39-L40

PR #577, opened 2020-09-03
Lemma mul_rVP

CohenCyril:

@affelt-aist

Fixes mathcomp https://github.com/math-comp/real-closed/issues/1

But I am not convinced it is really useful (it's basically
 `row_matrixP` + `rowE`).

##### Things done/to do

### DIFF
--- a/mathcomp/algebra/matrix.v
+++ b/mathcomp/algebra/matrix.v
@@ -1678,6 +1678,9 @@ apply/rowP=> j; rewrite !mxE (bigD1_ord i) //= mxE !eqxx mul1r.
 by rewrite big1 ?addr0 // => i'; rewrite mxE /= lift_eqF mul0r.
 Qed.
 
+Lemma mul_rVP m n A B :((@mulmx 1 m n)^~ A =1 mulmx^~ B) <-> (A = B).
+Proof. by split=> [eqAB|->//]; apply/row_matrixP => i; rewrite !rowE eqAB. Qed.
+
 Lemma row_mul m n p (i : 'I_m) A (B : 'M_(n, p)) :
   row i (A *m B) = row i A *m B.
 Proof. by rewrite !rowE mulmxA. Qed.
--- a/CHANGELOG_UNRELEASED.md
+++ b/CHANGELOG_UNRELEASED.md
@@ -86,6 +86,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - in `matrix.v`, new lemmas `row_usubmx`, `row_dsubmx`, `col_lsubmx`,
   and `col_rsubmx`.
 
+- in `matrix.v` new lemma `mul_rVP`.
+
 ### Changed
 
 - in ssrbool.v, use `Reserved Notation` for `[rel _ _ : _ | _]` to avoid warnings with coq-8.12


mathcomp/algebra/matrix.v
```diff
@@ -1646,6 +1646,9 @@ apply/rowP=> j; rewrite !mxE (bigD1 i) //= mxE !eqxx mul1r.
 by rewrite big1 ?addr0 // => i' ne_i'i; rewrite mxE /= (negPf ne_i'i) mul0r.
 Qed.
 
+Lemma mulmxP m n A B :((@mulmx 1 m n)^~ A =1 mulmx^~ B) <-> (A = B).
```
thery, line 1649: Should the fact that this reflection is for row vectors appear in the name of the lemma?

CohenCyril: Should it be named `mulmxrP` innstead?

CohenCyril: Or `mul_rV_mxP` ?

thery: rV_mulmxP?

CohenCyril: I'm afraid this name suggests we compare row vectors, but we compare full matrices by applying them to row vectors.

CohenCyril: (here `A` and `B` have type `'M_(m, n)`.

thery: but we can't multiply row vectors no?

CohenCyril: ... the lemma says that `A` and `B` are the same if and only if their multiplication with any row vector (on the left, following the diagrammatic convention of the library) yields the same row vector.

thery: ok let's go with `mul_rV_mxP` but I would never have guessed the name.
The theorem will mosty be used from the left to the right to add a row multiplication of the left. 
`mul_rVPl`?

CohenCyril: I'm fine with `mul_rVPl`.

thery: or maybe simply  `mul_rvP` and also a `mul_cVP` for the version on the right.

affeldt-aist: it is also reminiscent of the lemma `row_matrixP`, but `rVmul_matrixP` is maybe a bit long...

thery: `mul_rvPl` and `mul_rvP` are both fine with me. If you do the renaming and change the name of the PR accordingly, I'll merge it.

CohenCyril: :+1:

thery: @affeldt-aist I always thing it difficult to figure out the name of these theorems.
Specially when does on use `matrix` or `mx`

affeldt-aist: > Specially when does on use matrix or mx

I am not sure either, but it looks like the default is to use `matrix` and that the `mx` prefix is reserved for naming lemmas about an identifier of the form `blah(_|)mx` (`mxE` being an obvious exception though).

CohenCyril: > Specially when does on use `matrix` or `mx`

There are actually very few (10) theorems, one notation and zero definition (except for `matrix` itself which is only used with the notation `M`) with `matrix` as a substring, 
- `matrixP`, `row_matrixP`, `matrix_sum_delta`, `card_matrix`,`matrix_eq0`, `matrix0Pn`, `matrix_nonzero1`, `matrix_modl`, `matrix_modr` and `gal_matrix`
- and `\matrix_...`.

I believe we could rename all the lemma to use `mx` instead of `matrix`.

PR #576, opened 2020-09-03
variants of contra lemmas for ordered types

affeldt-aist:
##### Motivation for this change

`contra` lemmas for ordered types as proposed by @CohenCyril in https://github.com/math-comp/analysis/pull/237.

the purpose of this PR is to record the idea for discussion

##### Things done/to do

- [x] added corresponding entries in `CHANGELOG_UNRELEASED.md` (do not edit former entries)
~- [ ] added corresponding documentation in the headers~

##### Automatic note to reviewers

Read [this Checklist](https://github.com/math-comp/math-comp/wiki/Checklist-for-following,-reviewing-and-playing-with-a-PR#checklist-for-reviewing-a-pr) and make sure there is a milestone.

PR #578, opened 2020-09-03
Adding contra lemmas with orders

CohenCyril:

Adding contraposition lemmas for `leq`, `ltn`, `Order.le` and `Order.lt`.

##### Things done/to do

### DIFF
--- a/CHANGELOG_UNRELEASED.md
+++ b/CHANGELOG_UNRELEASED.md
@@ -10,7 +10,25 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Added
 
-- Added contrapostion lemmas involving propositions: `contra_not`, `contraPnot`, `contraTnot`, `contraNnot`, `contraPT`, `contra_notT`, `contra_notN`, `contraPN`, `contraFnot`, `contraPF` and `contra_notF` in ssrbool.v and `contraPeq`, `contra_not_eq`, `contraPneq`, and `contra_neq_not` in eqtype.v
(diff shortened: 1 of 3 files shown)


CohenCyril:
@affeldt-aist would you be ok to merge it with #576 and document?

CohenCyril:
LGTM except I'm an author too :laughing:

CohenCyril:
> LGTM except I'm an author too

I meant: I cannot review so the fact it looks good to me is irrelevant.

pi8027 (approved):

CohenCyril:
(rebased)

affeldt-aist:
I am wondering whether the explanation for the naming of the contra lemmas shouldn't go also into the contributing guide at the root of the repo.

CohenCyril:
> I am wondering whether the explanation for the naming of the contra lemmas shouldn't go also into the contributing guide at the root of the repo.

It would be more appropriate in a general documentation of ssrnat/ssrnum I think. It can be done as a new PR though.

chdoc (approved):
LGTM. However, given the large number of largely boring lemmas, I wonder whether it would be possible to introduce a `contra:` tactic that automatically picks the right contra lemma, or even constructs it from the shape of the goal. However, I'm not sure how well that would work with the `comparable_*` lemmas.

CohenCyril:
> LGTM. However, given the large number of largely boring lemmas, I wonder whether it would be possible to introduce a `contra:` tactic that automatically picks the right contra lemma, or even constructs it from the shape of the goal. However, I'm not sure how well that would work with the `comparable_*` lemmas.

I see how we could do this using canonical structures / typeclasses ... but for now, let's keep it like this.

CohenCyril:
Please open an issue / feature request with your nice idea  :+1:

PR #582, opened 2020-09-08
Fix map_mx_id and a few implicit arguments

CohenCyril:

Missing lemma from previous PR #568.
Part of #207

##### Things done/to do

### DIFF
--- a/mathcomp/algebra/matrix.v
+++ b/mathcomp/algebra/matrix.v
@@ -1100,28 +1100,33 @@ End MapMatrix.
 Arguments map_mx {aT rT} f {m n} A.
 
 Section MultipleMapMatrix.
+Context {R S T : Type} {m n : nat}.
 Local Notation "M ^ phi" := (map_mx phi M).
 
-Lemma map_mx_id (R : Type) m n (M : 'M[R]_(m,n)) : M ^ id = M.
+Lemma map_mx_comp (f : R -> S) (g : S -> T)
+  (M : 'M_(m, n)) : M ^ (g \o f) = (M ^ f) ^ g.
 Proof. by apply/matrixP => i j; rewrite !mxE. Qed.
 
-Lemma map_mx_comp (R S T : Type) m n (M : 'M[R]_(m,n))
-      (f : R -> S) (g : S -> T) : M ^ (g \o f) = (M ^ f) ^ g.
-Proof. by apply/matrixP => i j; rewrite !mxE. Qed.
+Lemma eq_in_map_mx (g f : R -> S) (M : 'M_(m, n)) :
+  (forall i j, f (M i j) = g (M i j)) -> M ^ f = M ^ g.
+Proof. by move=> fg; apply/matrixP => i j; rewrite !mxE. Qed.
 
-Lemma eq_in_map_mx (R S : Type) m n (M : 'M[R]_(m,n)) (g f : R -> S) :
-  (forall i j, f (M i j) = g (M i j)) <-> M ^ f = M ^ g.
-Proof. by rewrite -matrixP; split => fg i j; have := fg i j; rewrite !mxE. Qed.
+Lemma eq_map_mx (g f : R -> S) : f =1 g ->
+  forall (M : 'M_(m, n)), M ^ f = M ^ g.
+Proof. by move=> eq_fg M; apply/eq_in_map_mx. Qed.
 
-Lemma eq_map_mx (R S : Type) m n (M : 'M[R]_(m,n)) (g f : R -> S) :
-  f =1 g -> M ^ f = M ^ g.
-Proof.  by move=> eq_fg; apply/eq_in_map_mx. Qed.
-
-Lemma map_mx_id_in (R : Type) m n (M : 'M[R]_(m,n)) (f : R -> R) :
+Lemma map_mx_id_in (f : R -> R) (M : 'M_(m, n)) :
   (forall i j, f (M i j) = M i j) -> M ^ f = M.
-Proof. by rewrite -[RHS]map_mx_id -eq_in_map_mx. Qed.
+Proof. by move=> fM; apply/matrixP => i j; rewrite !mxE. Qed.
+
+Lemma map_mx_id (f : R -> R) : f =1 id -> forall M : 'M_(m, n), M ^ f = M.
+Proof. by move=> fid M; rewrite map_mx_id_in. Qed.
 
 End MultipleMapMatrix.
+Arguments eq_map_mx {R S m n} g [f].
+Arguments eq_in_map_mx {R S m n} g [f M].
+Arguments map_mx_id_in {R m n} [f M].
+Arguments map_mx_id {R m n} [f].
 
 (*****************************************************************************)
 (********************* Matrix Zmodule (additive) structure *******************)


mathcomp/algebra/matrix.v
```diff
@@ -1067,7 +1067,13 @@ Lemma map_mx_id_in (R : Type) m n (M : 'M[R]_(m,n)) (f : R -> R) :
   (forall i j, f (M i j) = M i j) -> M ^ f = M.
 Proof. by rewrite -[RHS]map_mx_id -eq_in_map_mx. Qed.
 
+Lemma map_mx_eq_id (R : Type) m n (M : 'M[R]_(m,n)) (f : R -> R) :
```
thery, line 1070: In poly, this theorem is simply called `map_poly_id`.

CohenCyril: `map_mx_id` is taken (see [above](https://github.com/math-comp/math-comp/blob/693a9fc82903f19770027d9b26f66de61377f221/mathcomp/algebra/matrix.v#L1051-L1052)), but maybe I could simply replace it, WDYT?

pi8027: For me, it seems better to replace `map_mx_id`.

mathcomp/algebra/matrix.v
```diff
@@ -1067,7 +1067,13 @@ Lemma map_mx_id_in (R : Type) m n (M : 'M[R]_(m,n)) (f : R -> R) :
...
+  f =1 id -> M ^ f = M.
+Proof. by move=> /eq_map_mx->; rewrite map_mx_id. Qed.
```
pi8027, line 1072:
```suggestion
Proof. by move/eq_map_mx->; rewrite map_mx_id. Qed.
```

mathcomp/algebra/matrix.v
thery: can we just remove `map_mx_id`? I think it is not much used and

```
Lemma map_mx_id (R : Type) m n (M : 'M[R]_(m,n)) : M ^ id = M.
Proof. by rewrite map_mx_eq_id . Qed.
```

CohenCyril: I'd rather not remove `map_mx_id`, the specialization to `id` makes `rewrite map_mx_id` possible without pointing out which one is `id` (otherwise patterns might be needed).

pi8027 (commented):
I think `eq_map_mx_id` might be a better name, but I'm not sure. Except that, it looks good to me.

CohenCyril:
> I think `eq_map_mx_id` might be a better name, but I'm not sure. Except that, it looks good to me.

The `eq` in the name is not for the equality in the conclusion, but the equality in the hypothesis. In the same way one reads `map_mx_id` as "`map_mx` applied to `id` is equal to something", one should read `map_mx_eq_id` as "`map_mx` applied to something equal to `id`, is equal to something".

pi8027 (approved):

mathcomp/algebra/matrix.v
thery: Ok but then rename the theorem in `poly` to match the one in `matrix`

pi8027: > Ok but then rename the theorem in `poly` to match the one in `matrix`

Make sense. @CohenCyril Please let me know whether you plan to do this in this PR or in another PR.

CohenCyril: I am looking into it right now...

CohenCyril:
Actually, after performing the change in `poly.v` I realized `map_poly_id` was never used, and I revised the uses of `map_mx_id` in my developments. And in fact, as I set the implicits `(map_mx_eq_id id)` does roughly the same job as `map_mx_id` and is used less often.
So ... I backtrack and I will come back with a new set of lemmas.

mathcomp/algebra/matrix.v
```diff
@@ -1100,28 +1100,33 @@ End MapMatrix.
...
 
-Lemma map_mx_comp (R S T : Type) m n (M : 'M[R]_(m,n))
-      (f : R -> S) (g : S -> T) : M ^ (g \o f) = (M ^ f) ^ g.
-Proof. by apply/matrixP => i j; rewrite !mxE. Qed.
-
-Lemma eq_in_map_mx (R S : Type) m n (M : 'M[R]_(m,n)) (g f : R -> S) :
-  (forall i j, f (M i j) = g (M i j)) <-> M ^ f = M ^ g.
-Proof. by rewrite -matrixP; split => fg i j; have := fg i j; rewrite !mxE. Qed.
+Lemma eq_in_map_mx (g f : R -> S) (M : 'M_(m, n)) :
+  (forall i j, f (M i j) = g (M i j)) -> M ^ f = M ^ g.
```
pi8027, line 1111: This lemma was iff before this change. Is it really OK?

mathcomp/algebra/matrix.v
```diff
@@ -1100,28 +1100,33 @@ End MapMatrix.
...
   (forall i j, f (M i j) = M i j) -> M ^ f = M.
-Proof. by rewrite -[RHS]map_mx_id -eq_in_map_mx. Qed.
+Proof. by move=> fM; apply/matrixP => i j; rewrite !mxE. Qed.
+
+Lemma map_mx_id (f : R -> R) : f =1 id -> forall M : 'M_(m, n), M ^ f = M.
+Proof. by move=> fid M; rewrite map_mx_id_in. Qed.
 
 End MultipleMapMatrix.
+Arguments eq_map_mx {R S m n} g [f].
+Arguments eq_in_map_mx {R S m n} g [f M].
```
pi8027, line 1127: The ordering of `f` and `g` is opposite and `g` is implicit in `eq_map_poly`. Shouldn't it be the same?

mathcomp/algebra/matrix.v
CohenCyril: The iff has no advantage over `matrixP` lemma itself. But as a forward only view, one can use the rewriting left to right with an explicit `g` to give the target function.

mathcomp/algebra/matrix.v
CohenCyril: It's a trade off, such lemma can be used in four different styles:
1. rewriting `eq_map_stuff` using an function equality as argument. (the most used for `eq_map_poly`)
2. rewriting `eq_map_stuff` using a function `g` as argument. (the least used, but still used)
3. `apply: eq_map_stuff` (the second most used, for `eq_map_poly`)
4. `move=> /eq_map_stuff` (the third most used, for `eq_map_poly`)

For the last two, the choice of implicit arguments has no impact whatsoever (except arguments of type `Type` must be either `nonPropType` or maximal implicit, which I made sure of for `eq_map_mx`).

Now, for 1. and 2., there is an impact, let's see it:
1. a. If one lets `f` and `g` implicit, then there is no `_` and one can write `rewrite (eq_map_stuff eq_f_g)` directly.
b. Otherwise, if `g` is explicit, then one needs to add one `_`, which can be bothering but will be nothing compared to what is coming in 2.
2. a. If one lets `f` and `g` implicit, then there is no `_` and one must write `rewrite (@eq_map_stuff _ ... _ g)`. In the case of `eq_map_poly` it's three of them, with `eq_map_mx` without swapping the arguments, it would be five (and with swapping, four). 
b. Otherwise, if `g` is explicit, one can just write `rewrite (eq_map_stuff g)`.

Hence my preference for option `b` for matrices. For polynomials, if I changed something, I suppose it would simply be reordering `g` and `f` so that one needs to put only two `_` to explicit `g`, which would slightly improve the current situation.

pi8027: OK. I agree.

pi8027 (approved):

PR #583, opened 2020-09-08
Adding matrix commutation and its theory

CohenCyril:

This is a prerequisite for theorems about commutation and stability,
essential in co-reduction of endomorphisms.
Part of #207.

##### Things done/to do

### DIFF
--- a/CHANGELOG_UNRELEASED.md
+++ b/CHANGELOG_UNRELEASED.md
@@ -206,6 +206,16 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
     * application of kernel lemmas `mxdirect_sum_geigenspace`,
     * new lemmas: `eigenpolyP`, `eigenvalue_poly`, `eigenspace_sub_geigen`,
   + new `map_mx` lemmas: `map_kermxpoly`, `map_geigenspace`, `eigenpoly_map`.
+- in `matrix.v`, added `comm_mx` and `comm_mxb` the propositional and
+  boolean versions of matrix commutation, `comm_mx A B` is
+  definitionally equal to `GRing.comm A B` when `A B : 'M_n.+1`, this
+  is witnessed by the lemma `comm_mxE`.  New notation `all_comm_mx`
+  stands for `allrel comm_mxb`. New lemmas `comm_mx_sym`, `comm_mx_refl`,
+  `comm_mx0`, `comm0mx`, `comm_mx1`, `comm1mx`, `comm_mxN`, `comm_mxN1`,
+  `comm_mxD`, `comm_mxB`, `comm_mx_sum`, `comm_mxP`, `all_comm_mxP`,
+  `all_comm_mx1`, `all_comm_mx2P`, `all_comm_mx_cons`, `comm_mxC`,
+  `commCmx`, `comm_mx_horner`, and `horner_mxC`. The common arguments
+  of these lemmas `R` and `n` are maximal implicits.
 
   - in `seq.v`, added `in_mask`, `cons_subseq`, `undup_subseq`, `subset_maskP`.
   - in `fintype.v`, added `mask_enum_ord`.
@@ -254,6 +264,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
     `BLeft x` and `BRight x` respectively mean close and open bounds as left bounds,
     and they respectively mean open and close bounds as right bounds.
     This change gives us the canonical "left to right" ordering of interval bounds.
+- In `matrix.v`, generalized `diag_mx_comm` and `scalar_mx_comm` to
+  all `n`, instead of `n'.+1`, thanks to `commmmx`.
 
 - in `interval.v`:
   + Lemmas `mid_in_itv(|oo|cc)` have been generalized from `realFieldType` to
@@ -314,6 +326,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   + `itv_intersectionii` (use `meetxx` instead)
   + `itv_intersectionC` (use `meetC` instead)
   + `itv_intersectionA` (use `meetA` instead)
+
+- in `mxpoly.v`, we deprecate `scalar_mx_comm`, and advise to use
+  `comm_mxC` instead (with maximal implicit arguments `R` and `n`).
 
 ### Infrastructure
 
--- a/mathcomp/algebra/matrix.v
+++ b/mathcomp/algebra/matrix.v
@@ -110,6 +110,9 @@ From mathcomp Require Import div prime binomial ssralg finalg zmodp countalg.
 (*   A \in unitmx == A is invertible (R must be a comUnitRingType).           *)
 (*        invmx A == the inverse matrix of A if A \in unitmx A, otherwise A.  *)
 (* A \is a mxOver S == the matrix A has its coefficients in S.                *)
+(*     comm_mx A B := A *m B = B *m A                                         *)
+(*    comm_mxb A B := A *m B == B *m A                                        *)
+(*  all_comm_mx As := allrel comm_mxb                                         *)
 (* The following operations provide a correspondence between linear functions *)
 (* and matrices:                                                              *)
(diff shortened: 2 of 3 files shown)


CHANGELOG_UNRELEASED.md
```diff
@@ -105,6 +105,16 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - in `mxpoly.v` new lemmas `horner_mx_diag`, `char_poly_trig`,
    `root_mxminpoly`, and `mxminpoly_diag`
 
+- in `matrix.v`, added `commmx` and `commmxb` the propositional and
```
thery, line 108: This duplication is really sad. Is this really needed?

CohenCyril: Unfortunately it is... until we have a structure for abelian categories...

CohenCyril: Did you mean the duplication between `GRing.comm` and `commmx` ?

thery: yep

CohenCyril: Then yes, `GRing.comm` is bound to (non trivial) rings, and matrices form a (non trivial) ring only when they are square and of manifestly positive size. However during proofs in endomorphism reduction, we take restrictions, which are matrices of size `#|V|` (with `V` a matrix space) and it becomes cumbersome to state commutation between restrictions, unless we relax the setting... and this relaxation corresponds to `commmx`...

thery: Is this really needed?

CohenCyril: Do you see an alternative?

CohenCyril: You can take a look [here](https://github.com/math-comp/math-comp/pull/207/files#diff-e50ca2313cb816e520933271b7597638) to see the uses cases (do not forget to expand collapsed code blocks)

thery: It is difficult to say without knowing the application, naively manipulating matrices of size #|V|.-1.+1 would not work?

CohenCyril: > It is difficult to say without knowing the application, naively manipulating matrices of size #|V|.-1.+1 would not work?

This is how I started, but since #|V| occurs in many matrix sizes and ordinals in the context and since the subspace is not always assumed to be nonzero, it introduces a nonnegligible overhead in already complicated proofs, and this overhead collapses when introducing  `commmx`.

affeldt-aist: Shouldn't the reason for introducing `commmx` be documented in the header file? This question is likely to be asked again. I think that it is almost a matter of copy-pasting the explanations above and maybe add a link to the file of the development where the issue appears and is solved.

CohenCyril: Maybe not in the header, but where `commmx` is defined.

CohenCyril: done

CohenCyril:
@thery, @affeldt-aist I find `commmx` particularily unreadable because of the sequence of three `m`, I just thought I could name it `mxcomm` instead. WDYT?

thery:
I'd prefer comm_mx  to mxcom

affeldt-aist:
> I'd prefer comm_mx to mxcomm

It also seems to me more homogeneous because there seems to be relatively few definitions that start with the `mx` prefix (`mxvec`, `mxtrace`) and even in those cases there seems to be a tendency to put `mx` as a suffix (e.g., Lemma `trace_mx11`).

CohenCyril:
fixed @affeldt-aist @thery

mathcomp/algebra/matrix.v
```diff
@@ -2576,20 +2664,26 @@ Proof.
...
 
-Lemma scalar_mx_comm n' a (A : 'M[R]_n'.+1) : GRing.comm A a%:M.
+Lemma scalar_mx_comm n a (A : 'M[R]_n) : comm_mx A a%:M.
 Proof. exact: scalar_mxC. Qed.
 
 Lemma mul_mx_scalar m n a (A : 'M[R]_(m, n)) : A *m a%:M = a *: A.
 Proof. by rewrite scalar_mxC mul_scalar_mx. Qed.
 
+Lemma comm_mxC n a (A : 'M[R]_n) : comm_mx A a%:M.
+Proof. by rewrite /comm_mx mul_mx_scalar mul_scalar_mx. Qed.
```
affeldt-aist, line 2682:
```suggestion
Proof. exact: scalar_mx_comm. Qed.
```

affeldt-aist (commented):
Aren't `comm_mxC` and `scalar_mx_comm` the same lemmas?
Beside, the proof could equally be `exact: scalar_mxC` which the same lemma but stated differentely, but maybe that is intended.

mathcomp/algebra/matrix.v
CohenCyril: They are not exactly the same, on one side `mul_mx_scalar` is more general (non square matrices), and on the other side, `comm_mx` might be used as a head symbol for automatic proof search (or simply human readability purposes).

thery:
@CohenCyril do you want to squash your commit before I merge?

CohenCyril:
@thery 

> @CohenCyril do you want to squash your commit before I merge?

done

mathcomp/algebra/matrix.v
affeldt-aist: Indeed, they are not the same lemmas, my bad.

Still, there is one that is more general than the other. What you are saying is that the specialization with a head symbol might be useful for automation purposes but this is not used yet, right?

thery:
@affeldt-aist  ok to merge?

affeldt-aist:
> @affeldt-aist ok to merge?

Aren't `scalar_mx_comm` and `comm_mxC` the same lemmas?

thery:
@affeldt-aist  ok we wait for @CohenCyril reaction

affeldt-aist:
> @affeldt-aist ok we wait for @CohenCyril reaction

I asked the same question yesterday and was happy with the answer but know I wonder is this answer was really an answer to my question. Anyway, the code has changed in between... Sorry for the poof follow-up. :-(

mathcomp/algebra/matrix.v
CohenCyril: Yes, actually, a hint resolve might just do the trick... maybe it should be added preemptively...

CohenCyril:
> > @affeldt-aist ok we wait for @CohenCyril reaction
> 
> I asked the same question yesterday and was happy with the answer but know I wonder is this answer was really an answer to my question. Anyway, the code has changed in between... Sorry for the poor follow-up. :-(

No you are right, I misread your question...

CohenCyril:
`scalar_mx_comm` and `comm_mxC` are exactly the same now! I guess `scalar_mx_comm` should be deprecated in favor of `comm_mxC`. Let me do that, and thanks for insisting.

mathcomp/algebra/matrix.v
CohenCyril: Actually, my bad, the hint is there.

mathcomp/algebra/mxpoly.v
```diff
@@ -271,7 +271,7 @@ Local Notation n := n'.+1.
 Variable A : 'M[R]_n.
 Implicit Types p q : {poly R}.
 
-Definition horner_mx := horner_morph (fun a => scalar_mx_comm a A).
+Definition horner_mx := horner_morph (fun a => comm_mxC a A).
```
affeldt-aist, line 274: this could be
`Definition horner_mx := horner_morph (fun a => comm_mxC a A).`
provided that
`Arguments comm_mxC {R n} a A`
is added at the end of the `ComMatrix` section in `matrix.v`

CohenCyril: ok! fixed

thery:
@affeldt-aist  ok for merging?

affeldt-aist:
> @affeldt-aist ok for merging?

Yes.

PR #585, opened 2020-09-08
Kernel lemmas

CohenCyril:

- new definitions `kermxpoly g p` (the kernel of polynomial $p(g)$).
  + new elementary theorems: `kermxpolyC`, `kermxpoly1`, `kermxpolyX`, `kermxpoly_min`
  + kernel lemmas: `mxdirect_kermxpoly`, `kermxpolyM`, `kermxpoly_prod`, and `mxdirect_sum_kermx`
  + correspondance between `eigenspace` and `kermxpoly`: `eigenspace_poly`
- generalized eigenspace `geigenspace` and a generalization of eigenvalues called `eigenpoly g` (i.e. polynomials such that `kermxpoly g p` is nonzero, e.g. eigen polynomials of degree 1 are of the form `'X - a%:P` where `a` are eigenvalues).
  + correspondance with `kermx`: `geigenspaceE`
  + application of kernel lemmas `mxdirect_sum_geigenspace`
  + new lemmas: `eigenpolyP`, `eigenvalue_poly`
- new `map_mx` lemmas: `map_kermxpoly`, `map_geigenspace`, `eigenpoly_map`.

Part of #207

##### Things done/to do

### DIFF
--- a/mathcomp/algebra/mxpoly.v
+++ b/mathcomp/algebra/mxpoly.v
@@ -30,6 +30,10 @@ From mathcomp Require Import ssralg zmodp matrix mxalgebra poly polydiv.
 (* degree_mxminpoly A == the (positive) degree of mxminpoly A.                *)
 (* mx_inv_horner A == the inverse of horner_mx A for polynomials of degree    *)
 (*                  smaller than degree_mxminpoly A.                          *)
+(*    kermxpoly g p == the kernel of p(g)                                     *)
+(*  geigenspace g a == the generalized eigenspace of g for eigenvalue a       *)
+(*                  := kermxpoly g ('X ^ n - a%:P) where g : 'M_n             *)
+(*  eigenpoly g p <=> p is an eigen polynomial for g, i.e. kermxpoly g p != 0 *)
 (*  integralOver RtoK u <-> u is in the integral closure of the image of R    *)
 (*                  under RtoK : R -> K, i.e. u is a root of the image of a   *)
 (*                  monic polynomial in R.                                    *)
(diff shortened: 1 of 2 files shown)


mathcomp/algebra/mxpoly.v
```diff
@@ -786,6 +790,155 @@ Qed.
...
+  by rewrite mulrAC [q * p]mulrC rmorphM/= mulrA -!mulmxE mulmx_ker mul0mx.
+rewrite -[_ * _ * q]mulrA [u * _]mulrC.
+by rewrite rmorphM mulrA -!mulmxE mulmx_ker mul0mx.
+Qed.
+
+Lemma kermxpoly_prod n (g : 'M_n)
+    (I : finType) (P : {pred I}) (p_ : I -> {poly K}) :
+  {in P &, forall i j, j != i -> coprimep (p_ i) (p_ j)} ->
+  (kermxpoly g (\prod_(i | P i) p_ i) :=: \sum_(i | P i) kermxpoly g (p_ i))%MS.
+Proof.
```
thery, line 853: Don't understand why this proof is so long.
Thought it was a simple elimination with ind2.
Unfortunately ind2 seems to weak.

mathcomp/algebra/mxpoly.v
```diff
@@ -30,6 +30,10 @@ From mathcomp Require Import ssralg zmodp matrix mxalgebra poly polydiv.
 (* degree_mxminpoly A == the (positive) degree of mxminpoly A.                *)
 (* mx_inv_horner A == the inverse of horner_mx A for polynomials of degree    *)
 (*                  smaller than degree_mxminpoly A.                          *)
+(*    kermxpoly g p == the kernel of p(g)                                     *)
+(*  geigenspace g a == the generalized eigenspace of g for eigenvalue a       *)
+(*                  := kermxpoly ('X ^ n - a%:P) where g : 'M_n               *)
```
CohenCyril, line 35:
```suggestion
(*                  := kermxpoly g ('X ^ n - a%:P) where g : 'M_n             *)
```

mathcomp/algebra/mxpoly.v
CohenCyril: :laughing: I had exactly the same thoughts, in this order... but then I realized that it is not a simple elimination, because the parisize coprimality hypothesis is used to derive coprimality with partial products, and simple inductions do not take into account partial big operators. So maybe there is a higher order induction scheme for this, but I did not figure it out, and went for an induction on the sequence instead.

CohenCyril:
@thery are you ok with a merge?

PR #588, opened 2020-09-09
Fix big meet and join notations for dual_display

pi8027:

These `Local Notation`s were removed in #454 and this removal broke the following `\join^d_` and `\meet^d_` notations.

##### Things done/to do

### DIFF
--- a/CHANGELOG_UNRELEASED.md
+++ b/CHANGELOG_UNRELEASED.md
@@ -133,6 +133,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   deprecated and will be removed two releases from now.
 
 - in `eqtype.v` new lemmas `contra_not_neq`, `contra_eq_not`.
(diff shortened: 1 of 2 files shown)


pi8027:
Since the removal does not break compilation, I think this fix is "indirect" in some sense. So it is probably better to redefine these big notations using `\big[dual_join/dual_bottom]_` and `\big[dual_meet/dual_top]_`. BTW, why we do not have `0^d` and `1^d` notations in the `order_scope`?

CohenCyril:
> Since the removal does not break compilation, I think this fix is "indirect" in some sense. So it is probably better to redefine these big notations using `\big[dual_join/dual_bottom]_` and `\big[dual_meet/dual_top]_`.

I am ok with the current fix but with a comment explaining the reason for the local notations (I think I did not see the point and deleted them because of that).

> BTW, why we do not have 0^d and 1^d notations in the order_scope?

Forgetfulness? Let's add them.

pi8027:
Done.

CohenCyril (approved):
LGTM

PR #492, opened 2020-05-01
New `big_uncond` and `big_rmcond -> big_rmcond_in`

CohenCyril:
##### Motivation for this change
- Lemma `big_rmcond` has been renamed `big_rmcomd_in` and the variant which does not require an `eqType` has been added and named `big_uncond`.
- The name `big_rmcond` is deprecated and will be reused for `big_uncond` in the next version of math-comp
- Additionally `big_uncond_in` is made available for uniformity, but is already deprecated.

##### Things done/to do

- [x] added corresponding entries in `CHANGELOG_UNRELEASED.md`
- [ ] after merge: open an issue to deal with the deprecations in 1.13 and 1.14.
- ~added corresponding documentation in the headers~

##### Automatic note to reviewers

Read [this Checklist](https://github.com/math-comp/math-comp/wiki/Checklist-for-following,-reviewing-and-playing-with-a-PR#checklist-for-reviewing-a-pr) and make sure there is a milestone.

PR #592, opened 2020-09-10
Fix naming inconsistencies between `imset` and `map` lemmas. 

### LINKED ISSUE #508: mem_imset mem_map naming/statement inconsistency
while cleaning, I noticed the following oddity:
```coq
Lemma mem_imset (aT rT : finType) (f : aT -> rT) (D : {pred aT}) (x : aT) : 
  x \in D -> f x \in [set f x | x in D]

Lemma map_f (T1 T2 : eqType) (f : T1 -> T2) (s : seq T1) (x : T1) : 
   x \in s -> f x \in [seq f i | i <- s]
Lemma mem_map (T1 T2 : eqType) (f : T1 -> T2),
  injective f -> forall (s : seq T1) (x : T1), (f x \in [seq f i | i <- s]) = (x \in s)
```
Wouldn't it be more consistent to rename `mem_imset` to `imset_f` and add the following lemma:
```coq
Lemma mem_imset (aT rT : finType) (f : aT -> rT) (A : {set aT}) (x : aT) :
  injective f -> (f x \in f @: A) = (x \in A).
```

chdoc:

fixes #508 (the first stage)

##### Things done/to do

- [ ] after merging, add issues to remove the deprecated lemmas and rename the `mem_imsert_eq` lemmas. 

### DIFF
--- a/CHANGELOG_UNRELEASED.md
+++ b/CHANGELOG_UNRELEASED.md
@@ -40,7 +40,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - in `finset.v`, new lemmas: `properC`, `properCr`, `properCl`
 - in `ssrnat.v`, new lemmas: `subn_minl`, `subn_maxl`
 - in `ssrnat.v`, new lemma: `oddS`
-
+- in `finset.v`, new lemmas: `mem_imset_eq`, `mem_imset2_eq`.
+  These lemmas will lose the `_eq` suffix in the next release, when the shortende names will become availabe (cf. Renamed section)
 
 - Added a factory `distrLatticePOrderMixin` in order.v to build a
   `distrLatticeType` from a `porderType`.
@@ -237,6 +238,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Renamed
 
 - `big_rmcond` -> `big_rmcond_in` (cf Changed section)
+- `mem_imset` -> `imset_f` (with deprecation alias, cf. Added section)
+- `mem_imset2` -> imset2_f` (with deprecation alias, cf. Added section)
 
 - in `interval.v`, we deprecate, rename, and relocate to `order.v` the following:
   + `lersif_(trans|anti)` -> `lteif_(trans|anti)`
--- a/mathcomp/character/classfun.v
+++ b/mathcomp/character/classfun.v
@@ -2355,7 +2355,7 @@ Proof.
 have [[injg defR] [injh defS]] := (isomP isoG, isomP isoH).
 rewrite !morphimEdom in defS defR; apply/cfun_inP=> s.
 rewrite -{1}defS => /imsetP[x Hx ->] {s}; have Gx := subsetP sHG x Hx.
-rewrite {1}eq_hg ?(cfResE, cfIsomE) // -defS -?eq_hg ?mem_imset // -defR.
+rewrite {1}eq_hg ?(cfResE, cfIsomE) // -defS -?eq_hg ?imset_f // -defR.
 by rewrite (eq_in_imset eq_hg) imsetS.
 Qed.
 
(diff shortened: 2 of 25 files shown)


chdoc:
@CohenCyril @gares can I get an assignee?

CHANGELOG_UNRELEASED.md
```diff
@@ -155,10 +157,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   `big_uncond` (cf Added) but it will be renamed `big_mkcond` in the
   next release.
 
-
 ### Renamed
 
 - `big_rmcond` -> `big_rmcond_in` (cf Changed section)
+- `mem_imset` -> `imset_f` (with deprecation alias, cf. Added section)
+- `mem_imset2` -> imset2_f` (with deprecation alias, cf. Added section)
```
CohenCyril, line 164: tick (`` ` ``) missing

CohenCyril:
@ggonthier @thery could you review this piece of code?

CHANGELOG_UNRELEASED.md
chdoc: missing tick added

CHANGELOG_UNRELEASED.md
```diff
@@ -40,6 +40,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - in `finset.v`, new lemmas: `properC`, `properCr`, `properCl`
 - in `ssrnat.v`, new lemmas: `subn_minl`, `subn_maxl`
 - in `ssrnat.v`, new lemma: `oddS`
+- in `finset.v`, new lemmas: `mem_imset_eq`, `mem_imset2_eq`.
+  Ideal names would be `mem_imset` and `mem_imset2`, but these names have just been deprecated from their previous meaning.
```
thery, line 44: Don't like this idea of temporary name. @maximedenes how coq deals with this kind of situation?

CohenCyril: This strategy was discussed [here](https://github.com/math-comp/math-comp/wiki/Minutes-June-30-2020] I cannot recall whether @maximedenes attended though...

thery: I wasn't there either :wink: still not a fan of temporary names

gares: You can say "these names will lose the _eq suffix in the next release", it may be of interest to the user. The reason why is not really helping IMO, and here you are not anyway suggesting we plan to use the ideal names in the future.

gares: I don't recall a case like this in Coq.

thery: If everybody like temporary names, then @CohenCyril could already make the next pull request for 1.13 so we are sure theses names wil disappear.

chdoc: > Don't like this idea of temporary name. @maximedenes how coq deals with this kind of situation?

I don't like the temporary names either, but I don't see another way without causing a breaking change. And this seems to minor an issue to cause a breaking change. 

> You can say "these names will lose the _eq suffix in the next release", it may be of interest to the user. The reason why is not really helping IMO, and here you are not anyway suggesting we plan to use the ideal names in the future.

I was merely following the formulations used by Cyril in #492 

> If everybody like temporary names, then @CohenCyril could already make the next pull request for 1.13 so we are sure theses names wil disappear.

This is an "after merge" Item, again as in #492

chdoc: It can't be a pull request yet, because the things to be removed aren't there yet ...

thery: @chdoc sorry I am reviewing too many pull requests by cyril I thought it was one of his too.

CohenCyril: > It can't be a pull request yet, because the things to be removed aren't there yet ...

Actually, it could be, I did not think about it, but it just needs to be a pr on top of this one :laughing:

thery:
@chdoc if you change the wording as sugested by @gares I think we can merge this PR

chdoc:
Reworded and rebased.

CohenCyril:
@thery waiting for your "green tick" ;)

thery:
:heavy_check_mark:

CohenCyril:
... I meant "Approve PR" and stuff, but that works for me too.

PR #593, opened 2020-09-12
lemma used in mathcomp-analysis about big and comparability/real

affeldt-aist:

Move a lemma from MathComp-Analysis to MathComp.

##### Things done/to do

### DIFF
--- a/CHANGELOG_UNRELEASED.md
+++ b/CHANGELOG_UNRELEASED.md
@@ -32,7 +32,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - in `ssralg.v`, new lemma `sumr_const_nat` and `iter_addr_0`
 - in `ssrnum.v`, new lemma `ler_sum_nat`
 - in `ssrnum.v`, new lemmas `max_real`, `min_real`, `bigmax_real`, `bigmin_real`
-`
+- in `order.v`, new lemms `comparable_big`
+
 - in `seq.v`, new lemmas: `take_uniq`, `drop_uniq`
 - in `fintype.v`, new lemmas: `card_geqP`, `card_gt1P`, `card_gt2P`,
   `card_le1_eqP` (generalizes `fintype_le1P`),
--- a/CHANGELOG_UNRELEASED.md
+++ b/CHANGELOG_UNRELEASED.md
@@ -31,7 +31,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
     `contraPleq`, `contraPltn`, `contra_not_leq`, `contra_not_ltn`, `contra_leq_not`, `contra_ltn_not`
 - in `ssralg.v`, new lemma `sumr_const_nat` and `iter_addr_0`
 - in `ssrnum.v`, new lemma `ler_sum_nat`
-
+- in `ssrnum.v`, new lemmas `max_real`, `min_real`, `bigmax_real`, `bigmin_real`
+`
 - in `seq.v`, new lemmas: `take_uniq`, `drop_uniq`
 - in `fintype.v`, new lemmas: `card_geqP`, `card_gt1P`, `card_gt2P`,
   `card_le1_eqP` (generalizes `fintype_le1P`),
--- a/CHANGELOG_UNRELEASED.md
+++ b/CHANGELOG_UNRELEASED.md
@@ -275,6 +275,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   + and the same holds for the dual `>=<^d`, `><^d`, the product
   `>=<^p`, `><^p`, and lexicographic `>=<^l`, `><^l`.
   The previous meanings can be obtained through `>=<%O x` and `><%O x`.
+- In `srnum.v`,
+  + `>=< y` stands for `[pred x | x >=< y]`
 
 
 ### Renamed
--- a/CHANGELOG_UNRELEASED.md
+++ b/CHANGELOG_UNRELEASED.md
@@ -31,8 +31,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
     `contraPleq`, `contraPltn`, `contra_not_leq`, `contra_not_ltn`, `contra_leq_not`, `contra_ltn_not`
 - in `ssralg.v`, new lemma `sumr_const_nat` and `iter_addr_0`
 - in `ssrnum.v`, new lemma `ler_sum_nat`
-- in `ssrnum.v`, new lemmas `max_real`, `min_real`, `bigmax_real`, `bigmin_real`
-- in `order.v`, new lemms `comparable_big`
+- in `ssrnum.v`, new lemmas `big_real`, `sum_real`, `prod_real`,
+  `max_real`, `min_real`, and `bigmax_real`, `bigmin_real`.
+- in `order.v`, new lemmas `comparable_bigl` and `comparable_bigr`.
 
 - in `seq.v`, new lemmas: `take_uniq`, `drop_uniq`
 - in `fintype.v`, new lemmas: `card_geqP`, `card_gt1P`, `card_gt2P`,
@@ -264,6 +265,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - The `class_of` records of structures are turned into primitive records to
   prevent prevent potential issues of ambiguous paths and convertibility of
   structure instances.
+- In `order.v`, rephrased `comparable_(min|max)[rl]` in terms of
+  `{in _, forall x y, _}`, hence reordering the arguments. Made them
+  hints for smoother combination with `comparable_big[lr]`.
+
 
 ### Renamed
 
--- a/CHANGELOG_UNRELEASED.md
+++ b/CHANGELOG_UNRELEASED.md
@@ -32,7 +32,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - in `ssralg.v`, new lemma `sumr_const_nat` and `iter_addr_0`
 - in `ssrnum.v`, new lemma `ler_sum_nat`
 - in `ssrnum.v`, new lemmas `big_real`, `sum_real`, `prod_real`,
-  `max_real`, `min_real`, and `bigmax_real`, `bigmin_real`.
+  `max_real`, `min_real`, `bigmax_real`, and `bigmin_real`.
 - in `order.v`, new lemmas `comparable_bigl` and `comparable_bigr`.
 
 - in `seq.v`, new lemmas: `take_uniq`, `drop_uniq`
--- a/CHANGELOG_UNRELEASED.md
+++ b/CHANGELOG_UNRELEASED.md
@@ -269,6 +269,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   `{in _, forall x y, _}`, hence reordering the arguments. Made them
   hints for smoother combination with `comparable_big[lr]`.
 
+- In `order.v`,
+  + `>=< y` stands for `[pred x | x >=< y]`
+  + `>< y`  stands for `[pred x | x >< y]`
+  + and the same holds for the dual `>=<^d`, `><^d`, the product
+  `>=<^p`, `><^p`, and lexicographic `>=<^l`, `><^l`.
+  The previous meanings can be obtained through `>=<%O x` and `><%O x`.
+
 
 ### Renamed
 
--- a/mathcomp/algebra/ssrnum.v
+++ b/mathcomp/algebra/ssrnum.v
@@ -1786,8 +1786,8 @@ Lemma min_real : {in real &, forall x y, min x y \is real}.
 Proof. by move=> x y ? ?; case: real_leP. Qed.
 
 Lemma bigmax_real I x0 (r : seq I) (P : pred I) (f : I -> R):
-  x0 \is real -> {in P, forall t : I, f t \is real} ->
-  \big[max/x0]_(t <- r | P t) f t \is real.
+  x0 \is real -> {in P, forall i : I, f i \is real} ->
+  \big[max/x0]_(i <- r | P i) f i \is real.
 Proof.
 by move=> x0r Pr; elim/big_rec : _ => // i x Pi xr; rewrite max_real ?Pr.
 Qed.
--- a/mathcomp/algebra/ssrnum.v
+++ b/mathcomp/algebra/ssrnum.v
@@ -1479,6 +1479,19 @@ rewrite sqrf_eq1 => /orP[/eqP //|]; rewrite -ger0_def le0r oppr_eq0 oner_eq0.
 by move/(addr_gt0 ltr01); rewrite subrr ltxx.
 Qed.
 
+Lemma big_real x0 op I (P : pred I) F (s : seq I) :
+  {in real &, forall x y, op x y \is real} -> x0 \is real ->
+  {in P, forall i, F i \is real} -> \big[op/x0]_(i <- s | P i) F i \is real.
+Proof. exact: comparable_bigr. Qed.
+
+Lemma sum_real I (P : pred I) (F : I -> R) (s : seq I) :
+  {in P, forall i, F i \is real} -> \sum_(i <- s | P i) F i \is real.
+Proof. by apply/big_real; [apply: rpredD | apply: rpred0]. Qed.
+
+Lemma prod_real I (P : pred I) (F : I -> R) (s : seq I) :
+  {in P, forall i, F i \is real} -> \prod_(i <- s | P i) F i \is real.
+Proof. by apply/big_real; [apply: rpredM | apply: rpred1]. Qed.
+
 Section NormedZmoduleTheory.
 
 Variable V : normedZmodType R.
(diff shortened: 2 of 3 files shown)


mathcomp/algebra/ssrnum.v
```diff
@@ -3711,6 +3711,17 @@ Implicit Types x y z t : R.
 Lemma num_real x : x \is real. Proof. exact: num_real. Qed.
 Hint Resolve num_real : core.
 
+Lemma max_real x y : x \is real -> y \is real -> max x y \is real.
+Proof.
+by rewrite !realE => /orP[|] x0 /orP[|] y0;
+  rewrite le_maxr le_maxl x0 y0 !(orbT,orTb).
+Qed.
```
CohenCyril, line 3718:
```suggestion
Lemma max_real x y : x \is real -> y \is real -> max x y \is real.
Proof. by case: leP. Qed.

Lemma min_real x y : x \is real -> y \is real -> min x y \is real.
Proof. by case: leP. Qed.
```

CHANGELOG_UNRELEASED.md
```diff
@@ -30,8 +30,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
     `contra_leq`, `contra_ltn`, `contra_leq_ltn`, `contra_ltn_leq`,
     `contraPleq`, `contraPltn`, `contra_not_leq`, `contra_not_ltn`, `contra_leq_not`, `contra_ltn_not`
 - in `ssralg.v`, new lemma `sumr_const_nat` and `iter_addr_0`
-- in `ssrnum.v`, new lemma `ler_sum_nat`
-
+- in `ssrnum.v`, new lemmas `ler_sum_nat`, `max_real`, `bigmax_real`
+`
```
pi8027, line 34: "Do not edit former entries."

mathcomp/algebra/ssrnum.v
```diff
@@ -3711,6 +3711,17 @@ Implicit Types x y z t : R.
...
+Lemma max_real x y : x \is real -> y \is real -> max x y \is real.
+Proof. by case: leP. Qed.
+
+Lemma min_real x y : x \is real -> y \is real -> min x y \is real.
+Proof. by case: leP. Qed.
+
+Lemma bigmax_real (T : choiceType) x (D : seq T) (f : T -> R):
+  x \is real -> (forall t : T, t \in D -> f t \is real) ->
+  \big[max/x]_(t <- D) f t \is real.
+Proof. by move=> ?; elim/big_ind : _. Qed.
```
pi8027, line 3723: In this section, `R` is a `realDomainType`. They should be located in a section that assumes that `R` is a `numDomainType` (e.g., somewhere after `Lemma real_ltgt0P`).

pi8027 (changes requested):
I guess `max_real` and `min_real` can be stated as follows as in `realD`.
```coq
Lemma max_real : {in real &, forall x y, max x y \is real}.
Lemma min_real : {in real &, forall x y, min x y \is real}.
```
Also I'm not sure if `bigmax_real` really requires `T : choiceType` (`eqType` might be enough).

mathcomp/algebra/ssrnum.v
pi8027: So `case: leP` should not be applicable to prove them. Use `real_leP` instead.

mathcomp/algebra/ssrnum.v
```diff
@@ -1769,6 +1769,20 @@ move=> hx; rewrite -[X in `|X|]subr0; case: (@real_ltgtP 0 x);
...
 Qed.
 
+Lemma max_real : {in real &, forall x y, max x y \is real}.
+Proof. by move=> x y ? ?; case: real_leP. Qed.
+
+Lemma min_real : {in real &, forall x y, min x y \is real}.
+Proof. by move=> x y ? ?; case: real_leP. Qed.
+
+Lemma bigmax_real (T : eqType) x (D : seq T) (f : T -> R):
+  x \is real -> (forall t : T, t \in D -> f t \is real) ->
```
pi8027, line 1779:
```suggestion
  x \is real -> {in D, forall t : T, f t \is real} ->
```
Should we also have `bigmin_real`?

CHANGELOG_UNRELEASED.md
```diff
@@ -31,7 +31,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
     `contraPleq`, `contraPltn`, `contra_not_leq`, `contra_not_ltn`, `contra_leq_not`, `contra_ltn_not`
 - in `ssralg.v`, new lemma `sumr_const_nat` and `iter_addr_0`
 - in `ssrnum.v`, new lemma `ler_sum_nat`
-
+- in `ssrnum.v`, new lemmas `max_real`, `bigmax_real`
+`
```
pi8027, line 35:
```suggestion
- in `ssrnum.v`, new lemmas `max_real`, `min_real`, `bigmax_real`

```

mathcomp/algebra/ssrnum.v
CohenCyril: Yes we should.

CohenCyril: There are two possible phrasing for `forall t : T, t \in D -> f t \is real` :
- either `{in D, forall t : T, f t \is real}`,
- or `{homo f : x / D >-> mem real}`

mathcomp/algebra/ssrnum.v
```diff
@@ -1769,6 +1769,28 @@ move=> hx; rewrite -[X in `|X|]subr0; case: (@real_ltgtP 0 x);
...
+Qed.
+
+Lemma bigmin_real (T : eqType) x (D : seq T) (f : T -> R):
+  x \is real -> {in D, forall t : T, f t \is real} ->
+  \big[min/x]_(t <- D) f t \is real.
+Proof.
+rewrite big_seq_cond; elim/big_ind : _ => [//| |? /andP[t _] xr]; last exact.
+by move=> y z yr zr ? xr; rewrite min_real // ?(yr,zr).
+Qed.
+
```
CohenCyril, line 1793: I suspect (and hope) this to be a one-liner in `T^d`

mathcomp/algebra/ssrnum.v
```diff
@@ -1769,6 +1769,28 @@ move=> hx; rewrite -[X in `|X|]subr0; case: (@real_ltgtP 0 x);
...
+by move=> y z yr zr ? xr; rewrite max_real // ?(yr,zr).
+Qed.
+
+Lemma bigmin_real (T : eqType) x (D : seq T) (f : T -> R):
+  x \is real -> {in D, forall t : T, f t \is real} ->
+  \big[min/x]_(t <- D) f t \is real.
+Proof.
+rewrite big_seq_cond; elim/big_ind : _ => [//| |? /andP[t _] xr]; last exact.
+by move=> y z yr zr ? xr; rewrite min_real // ?(yr,zr).
+Qed.
```
CohenCyril, line 1792: I suspect (and hope) this to be a one-liner in `T^d`, e.g.
```suggestion
Proof. exact: (@bigmax_real [eqType of T^d]). Qed.
```

mathcomp/algebra/ssrnum.v
```diff
@@ -1769,6 +1769,28 @@ move=> hx; rewrite -[X in `|X|]subr0; case: (@real_ltgtP 0 x);
...
+Lemma min_real : {in real &, forall x y, min x y \is real}.
+Proof. by move=> x y ? ?; case: real_leP. Qed.
+
+Lemma bigmax_real (T : eqType) x (D : seq T) (f : T -> R):
+  x \is real -> {in D, forall t : T, f t \is real} ->
+  \big[max/x]_(t <- D) f t \is real.
+Proof.
+rewrite big_seq_cond; elim/big_ind : _ => [//| |? /andP[t _] xr]; last exact.
+by move=> y z yr zr ? xr; rewrite max_real // ?(yr,zr).
+Qed.
```
CohenCyril, line 1784: The best practice is to have a conditional big with a `in` on the predicate, this allows the index type (renamed from `T` to `I`) to be any type (not necessarily an `eqType`). One can recover the desired behavior thanks to `big_seq_cond`.
```suggestion
Lemma bigmax_real I x0 (r : seq I) (P : pred I) (f : I -> R):
  x0 \is real -> {in P, forall t : I, f t \is real} ->
  \big[max/x0]_(t <- r | P t) f t \is real.
Proof.
by move=> x0r Pr; elim/big_rec : _ => // i x Pi xr; rewrite max_real ?Pr.
Qed.
```

mathcomp/algebra/ssrnum.v
CohenCyril, line 1792: Same remark:
```suggestion
Lemma bigmin_real I x0 (r : seq I) (P : pred I) (f : I -> R):
  x0 \is real -> {in P, forall t : I, f t \is real} ->
  \big[min/x0]_(t <- r | P t) f t \is real.
Proof.
by move=> x0r Pr; elim/big_rec : _ => // i x Pi xr; rewrite min_real ?Pr.
Qed.
```

CohenCyril:
I think a similar result should appear in `order.v` for comparable elements.

affeldt-aist:
> I think a similar result should appear in `order.v` for comparable elements.

I guess that's not what you want, I'll investigate more tomorrow.

CohenCyril:
> > I think a similar result should appear in `order.v` for comparable elements.
> 
> I guess that's not what you want, I'll investigate more tomorrow.

not at all indeed ... It should look like:
```coq
{in P, forall x, x0 >=< x} -> {in P &, forall i j, f x >=< f y} ->
all (>=< \big[max/x0]_(i <- s | P i) f i) (x :: s).
```
With perhaps a particular case for `x` and for `s` when `I` is an `eqType`.

pi8027 (approved):
LGTM

mathcomp/ssreflect/order.v
```diff
@@ -3129,6 +3129,32 @@ Lemma nmono_leif (f : T -> T) C :
...
+Qed.
+
+Lemma all_comparable_bigmin I x0 (P : pred I) f (s : seq I) x xs :
+  {in x :: xs, forall y, x0 >=< y} ->
+  {in x :: xs, forall y : T, {in P, forall t, f t >=< y}} ->
+  all (>=< \big[min/x0]_(i <- s | P i) f i) (x :: xs).
+Proof.
+by move=> x0x Pf; apply/allP=> ? ?; rewrite comparable_bigmin ?x0x//; exact: Pf.
+Qed.
+
```
CohenCyril, line 3157: Given the nice phrasing you found for `comparable_bigmin` these `all` variants seem useless to me. If needed, it's easy to prove on the spot without induction.
```suggestion
```

mathcomp/ssreflect/order.v
```diff
@@ -3129,6 +3129,32 @@ Lemma nmono_leif (f : T -> T) C :
...
 
+Lemma comparable_bigmax I x0 (P : pred I) f (s : seq I) x :
+   x0 >=< x -> {in P, forall t, f t >=< x} ->
+  \big[max/x0]_(i <- s | P i) f i >=< x.
+Proof. by move=> ? ?; elim/big_ind : _ => // y z; exact: comparable_maxl. Qed.
+
+Lemma comparable_bigmin I x0 (P : pred I) f (s : seq I) x :
+   x0 >=< x -> {in P, forall t, f t >=< x} ->
+  \big[min/x0]_(i <- s | P i) f i >=< x.
+Proof. by move=> ? ?; elim/big_ind : _ => // y z; exact: comparable_minl. Qed.
```
CohenCyril, line 3140:
```suggestion
Lemma comparable_bigmax x0 I (P : pred I) (s : seq I) F x :
   x0 >=< x -> {in P, forall i, F i >=< x} ->
  \big[max/x0]_(i <- s | P i) F i >=< x.
Proof. by move=> ? ?; elim/big_ind : _ => // y z; exact: comparable_maxl. Qed.

Lemma comparable_bigmin x0 I (P : pred I) (s : seq I) F x :
   x0 >=< x -> {in P, forall i, f i >=< x} ->
  \big[min/x0]_(i <- s | P i) f i >=< x.
Proof. by move=> ? ?; elim/big_ind : _ => // y z; exact: comparable_minl. Qed.
```

mathcomp/algebra/ssrnum.v
```diff
@@ -1769,6 +1769,26 @@ move=> hx; rewrite -[X in `|X|]subr0; case: (@real_ltgtP 0 x);
...
+Proof.
+by move=> x0r Pr; elim/big_rec : _ => // i x Pi xr; rewrite max_real ?Pr.
+Qed.
+
+Lemma bigmin_real I x0 (r : seq I) (P : pred I) (f : I -> R):
+  x0 \is real -> {in P, forall t : I, f t \is real} ->
+  \big[min/x0]_(t <- r | P t) f t \is real.
+Proof.
+by move=> x0r Pr; elim/big_rec : _ => // i x Pi xr; rewrite min_real ?Pr.
+Qed.
```
CohenCyril, line 1790: Naive question is this simply `exact: comparable_bigmax`, `exact: comparable_bigmin` now?

CohenCyril:
Subsidiary question: shouldn't we have a generic `comparable_big` which generalizes both `comparable_bigmax` and  `comparable_bigmin` under the assumption that the iterated operation preserves comparability?

mathcomp/algebra/ssrnum.v
affeldt-aist: Almost. Modulo `comparabler0` rewrite which make the script longer than a single line.

affeldt-aist:
> Subsidiary question: shouldn't we have a generic `comparable_big` which generalizes both `comparable_bigmax` and `comparable_bigmin` under the assumption that the iterated operation preserves comparability?

Indeed. But with the new `comparable_big`, the old `comparable_bigmax`
is not exactly `comparable_big comparable_maxl` because of the order of arguments of `comparable_maxl`.
(BTW, I changed the notation `>=< x` because it looked counter-intuitive to me.)

CohenCyril:
> > Subsidiary question: shouldn't we have a generic `comparable_big` which generalizes both `comparable_bigmax` and `comparable_bigmin` under the assumption that the iterated operation preserves comparability?
> 
> Indeed. But with the new `comparable_big`, the old `comparable_bigmax`
> is not exactly `comparable_big comparable_maxl` because of the order of arguments of `comparable_maxl`.

Then swap the arguments ;)

> (BTW, I changed the notation `>=< x` because it looked counter-intuitive to me.)

You are right!

mathcomp/algebra/ssrnum.v
CohenCyril: But now with `comparable_big` isn't it better?

mathcomp/ssreflect/order.v
```diff
@@ -3129,6 +3129,16 @@ Lemma nmono_leif (f : T -> T) C :
   forall x y, (f x <= f y ?= iff C) = (y <= x ?= iff C).
 Proof. by move=> mf x y; rewrite /leif !eq_le !mf. Qed.
 
+Section comparable_big.
+Variables op : T -> T -> T.
+Hypothesis op_comparable : forall z, {in >=< z &, forall x y, op x y >=< z}.
```
CohenCyril, line 3134: This hypothesis is too weak...

CohenCyril:
BTW, making `comparable_minl` and `comparable_maxl` hints make `rewrite comparable_big//` simplify side conditions automatically if one rephrases them. I took the liberty to patch your PR, but I cannot seem to push in your branch, did you authorize pushs from maintainers?

CohenCyril (approved):
I patched this PR and I am ok with my own patch @pi8027 could you take a second look?

CohenCyril:
and ... rebased

mathcomp/algebra/ssrnum.v
CohenCyril: it is !

mathcomp/ssreflect/order.v
CohenCyril: better now

CHANGELOG_UNRELEASED.md
```diff
@@ -31,6 +31,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
     `contraPleq`, `contraPltn`, `contra_not_leq`, `contra_not_ltn`, `contra_leq_not`, `contra_ltn_not`
 - in `ssralg.v`, new lemma `sumr_const_nat` and `iter_addr_0`
 - in `ssrnum.v`, new lemma `ler_sum_nat`
+- in `ssrnum.v`, new lemmas `big_real`, `sum_real`, `prod_real`,
+  `max_real`, `min_real`, and `bigmax_real`, `bigmin_real`.
```
pi8027, line 35:
```suggestion
  `max_real`, `min_real`, `bigmax_real`, and `bigmin_real`.
```

mathcomp/ssreflect/order.v
```diff
@@ -1132,7 +1132,7 @@ Notation "< y :> T" := (< (y : T)) (only parsing) : order_scope.
 Notation "> y" := (lt y) : order_scope.
 Notation "> y :> T" := (> (y : T)) (only parsing) : order_scope.
 
-Notation ">=< y" := (comparable y) : order_scope.
+Notation ">=< y" := ([pred x | comparable x y]) : order_scope.
```
pi8027, line 1135: The ordering of arguments of `comparable` has been changed here. This should be recorded in CHANGELOG in my opinion. Otherwise, this should be `comparable y x`.

BTW, I don't see why we define `>=< _` twice in here and L.1157 (but not from this PR).

CohenCyril: > The ordering of arguments of `comparable` has been changed here. This should be recorded in CHANGELOG in my opinion. Otherwise, this should be `comparable y x`.

Yes, it must be documented and the change is intentional.

> BTW, I don't see why we define `>=< _` twice in here and L.1157 (but not from this PR).

I think this is the unfortunate result of a copy paste from `<=` and `>=` and the fact the symbol  `>=<` is symmetric.

pi8027 (changes requested):
The definition of `>=< y` in ssrnum should also be changed.

CohenCyril:
> The definition of `>=< y` in ssrnum should also be changed.

@affeldt-aist do you think you can do it?

affeldt-aist:
> > The definition of `>=< y` in ssrnum should also be changed.
> 
> @affeldt-aist do you think you can do it?

I'll try to by monday.

pi8027 (approved):
LGTM.

pi8027:
@CohenCyril Can you confirm the last commit and merge this PR?

PR #596, opened 2020-09-24
mulrIb

CohenCyril:

Adding `mulrIb` a specific injectivity case of `*+` when the lhs is a boolean.

##### Things done/to do

### DIFF
--- a/mathcomp/algebra/ssralg.v
+++ b/mathcomp/algebra/ssralg.v
@@ -835,6 +835,9 @@ Qed.
 Lemma mulrnAC x m n : x *+ m *+ n = x *+ n *+ m.
 Proof. by rewrite -!mulrnA mulnC. Qed.
 
+Lemma mulrIb (x : V) : x != 0 -> injective (fun b : bool => x *+ b).
+Proof. by move=> x_neq0 [[]|[/esym|]]// /eqP; rewrite (negPf x_neq0). Qed.
+
 Lemma iter_addr_0 n (m : V) : iter n (+%R m) 0 = m *+ n.
 Proof. by elim: n => //= n ->; rewrite mulrS. Qed.
 
--- a/CHANGELOG_UNRELEASED.md
+++ b/CHANGELOG_UNRELEASED.md
@@ -161,6 +161,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   (e.g. with `raddf_eq0`), and lemma `inj_row_free` characterizing
   `row_free` matrices `A`  through `v *m A = 0 -> v = 0` for all `v`.
 
+- in `ssralg.v`, new lemma `mulrIb`.
 
 ### Changed
 


chdoc:
This looks a bit bespoke to me, what's the motivation or where is this useful?

CohenCyril:
https://github.com/CohenCyril/signdet/blob/master/signdet.v#L318, ideally that would be `mulrn` is injective, but that's true only if the characteristic of the ring is `0`.

chdoc (approved):
LGTM

chdoc:
@CohenCyril I can't merge this, so what should I do now? Who, apart from you, is most familiar with that part of the library?

CohenCyril:
woops, I think @thery, @gares, @pi8027, @strub might all be good candidates

thery:
@CohenCyril sorry too cryptic for me, don't see how this could be a reusable theorem.

CohenCyril:
> @CohenCyril sorry too cryptic for me, don't see how this could be a reusable theorem.

Well I do think it is reusable. Let me try to explain it another way.

The construction `x *+ b` is `x` if `b` otherwise `0`, it is frequently used in combination with membership (e.g. `x \in G`) or Kronecker symbol (`i == j`), the latter gives a "Dirac" function `fun j => x *+ (i == j)` which is thus injective through `mulrIb`.
This is used pervasively in the representation of diagonal matrices `\matrix_(i, j) (d 0 i *+ (i == j))`, and this lemma helps identifying nonzero values of the matrix with elements of the diagonal.

In other words again, this lemma, explains these "Dirac/weight function" are injective on  `b` as long as `x` is nonzero, which allows to infer the validity of `b` (i.e.  `i == j`, or `i \in G`, depending on the use cases) from the nullity of `x`.

thery:
@CohenCyril I've understood what it is your use case, but I just doubt it will be reused in other places.

CohenCyril:
What is the conclusion?

thery:
not a fan

mathcomp/algebra/ssralg.v
```diff
@@ -835,6 +835,9 @@ Qed.
 Lemma mulrnAC x m n : x *+ m *+ n = x *+ n *+ m.
 Proof. by rewrite -!mulrnA mulnC. Qed.
 
+Lemma mulrIb (x : V) : x != 0 -> injective (fun b : bool => x *+ b).
```
pi8027, line 838: I have no opinion on how it is useful, but this should be exported from `GRing.Theory`.

chdoc:
@thery @CohenCyril As the assignee for this PR, I would be happy if we could make progress towards either merging or closing this PR. Personally, I don't think this one-liner will hurt, so I would be in favor of merging it, even if it may be slightly bespoke. 

@CohenCyril How often is this lemma used in your developments?

thery:
@chdoc as I said I am not a big fan, but it will not keep me from sleeping at night if it is merged. My only point if If it is a general construction it should at least be documented. 
As a naive user I would  have missed the theorem, as I would have coded the gard as
`(b%:R * e)` as I feel more at ease manipulating `*` than `*+`

PR #597, opened 2020-09-24
Injectivity for additive functions and mulmxr.

CohenCyril:

Injectivity for additive functions and mulmxr.

##### Things done/to do


affeldt-aist:
Lemmas `raddf_inj` and `raddf_eq0` look like they could be merged in a single lemma stated with an equivalence but i suspect that it would make it less practical, is that correct?

CohenCyril:
> Lemmas `raddf_inj` and `raddf_eq0` look like they could be merged in a single lemma stated with an equivalence but i suspect that it would make it less practical, is that correct?

Yes indeed. From injectivity one may want to rewrite, but to prove it, one would rather prove a single implication.

PR #521, opened 2020-06-04
In and on

CohenCyril:
##### Motivation for this change

These new lemmas describe more precisely the possible interactions between `in` and `on`.

##### Things done/to do

- [x] added corresponding entries in `CHANGELOG_UNRELEASED.md`
- ~added corresponding documentation in the headers~
- [ ] backport to Coq

##### Automatic note to reviewers

Read [this Checklist](https://github.com/math-comp/math-comp/wiki/Checklist-for-following,-reviewing-and-playing-with-a-PR#checklist-for-reviewing-a-pr) and make sure there is a milestone.

PR #358, opened 2019-06-13
Stability proofs of sort

pi8027:
##### Motivation for this change

This PR adds the following lemmas which state the stability of `path.merge` and `path.sort` in various ways.

```coq
sorted_merge
     : forall (T : Type) (leT : rel T),
       transitive leT ->
       forall s t : seq T, sorted leT (s ++ t) -> merge leT s t = s ++ t
merge_stable_path
     : forall (T : Type) (leT leT' : rel T),
       total leT ->
       forall (x : T) (s1 s2 : seq T),
       all (fun y : T => all (leT' y) s2) s1 ->
       path [rel x0 y | leT x0 y && (leT y x0 ==> leT' x0 y)] x s1 ->
       path [rel x0 y | leT x0 y && (leT y x0 ==> leT' x0 y)] x s2 ->
       path [rel x0 y | leT x0 y && (leT y x0 ==> leT' x0 y)] x
         (merge leT s1 s2)
merge_stable_sorted
     : forall (T : Type) (leT leT' : rel T),
       total leT ->
       forall s1 s2 : seq T,
       all (fun x : T => all (leT' x) s2) s1 ->
       sorted [rel x y | leT x y && (leT y x ==> leT' x y)] s1 ->
       sorted [rel x y | leT x y && (leT y x ==> leT' x y)] s2 ->
       sorted [rel x y | leT x y && (leT y x ==> leT' x y)] (merge leT s1 s2)
sorted_sort
     : forall (T : Type) (leT : rel T),
       transitive leT -> forall s : seq T, sorted leT s -> sort leT s = s
sort_stable
     : forall (T : Type) (leT leT' : rel T),
       total leT -> transitive leT' ->
       forall s : seq T,
       sorted leT' s ->
       sorted [rel x y | leT x y && (leT y x ==> leT' x y)] (sort leT s)
filter_sort
     : forall (T : Type) (leT : rel T),
       total leT -> transitive leT ->
       forall (p : pred T) (s : seq T),
       [seq x <- sort leT s | p x] = sort leT [seq x <- s | p x]
mask_sort
     : forall (T : Type) (leT : rel T),
       total leT -> transitive leT ->
       forall (s : seq T) (m : bitseq),
       {m_s : bitseq | mask m_s (sort leT s) = sort leT (mask m s)}
sorted_mask_sort
     : forall (T : Type) (leT : rel T),
       total leT -> transitive leT ->
       forall (s : seq T) (m : bitseq),
       sorted leT (mask m s) ->
[...]

PR #328, opened 2019-04-12
Modified the definition of sort to work on any type

ybertot:
The sorting algorithm does not really rely on the element type being an eqType.
It may make sense for the proofs, but not for the algorithm per se.  Avoiding the reference to eqType
makes it easier to use the parametricity plugin on this algorithm.

EDIT by Cyril:
- [x] `CHANGELOG_UNRELEASE.md` updated
- [x] documentation updated

Issue #319, opened 2019-04-03
Failure of automatic lemma invocation


ybertot:
Adding a lemma and placing a hint on this lemma, I am able to construct to goals that are apparently the same, but one is solved automatically while the other is not.

```coq

From mathcomp Require Import all_ssreflect.

Set Implicit Arguments.
Unset Strict Implicit.
Unset Printing Implicit Defensive.

Lemma iter_orbit (T : finType) f i (x : T) : iter i f x \in orbit f x.
Proof. by rewrite -fconnect_orbit fconnect_iter. Qed.

(* Just using the following command fails. *)

Hint Resolve iter_orbit : core.

(* replacing with the following Hint directive does no reveal the problem. *)
(* Hint Extern 0 (is_true (iter _ _ _ \in orbit _ _)) =>
  apply iter_orbit : core. *)


Lemma sample (P : finType) a (E : {set P}) (b : P) (f : P -> P) :
   E =i orbit f b -> iter a f b \in E.
Proof.
by move=> A; rewrite A.
Qed.

Lemma bug (P : finType) a (E : {set P}) (b : P) (f : P -> P) : 
   orbit f b =i E -> iter a f b \in E.
Proof.
move => A; rewrite -A.
simpl.  (* whether we include this or not, the problem appears. *)
Fail by [].
by simple apply iter_orbit.
Qed.
```

ggonthier:
This is a known issue, and essentially a Coq one, so if you want to draw attention to it you you also raise it over there. Essentially here the `SSReflect` `done` tactic call `trivial` to apply hints, and `trivial` only applies hints that match strictly syntactically, apparently even more so than `simple apply`.
   This tends to work very poorly with `MathComp` precisely because we have many lemmas with implicit parameters that can take convertible but syntactically different forms depending on the scripting history. This is why `MathComp` is _very_ selective in the hints it adds (going so far as to declare multiple hints to cover all cases, in `fingroup`).
   You can turn on `Set Printing All` to see what's going on here; reading the `ssrbool` documentation of the `\in` notation (now in `coq/plugins/ssreflect`, not `mathcomp`) will give you the full context, but I'll summarise:
- `x \in A` denotes the composition of _two_ polymorphic constants:
          `@in_mem ?Tx x (@mem ?Tp ?pA A)`
[...]

PR #601, opened 2020-09-30
Add `_in` variants of the sorting lemmas

pi8027:

This PR adds `_in` variants of the sorting lemmas that restrict the domain where totality, transitivity, etc. hold by a predicate.
- [done] ~This generalization has not been done for the stability lemmas yet.~
- [postponed] ~This PR also attempts to add the `sort_map_in` lemma taken from PRs #328 and #358, but this is not general enough yet.~

CC: @CohenCyril

##### Things done/to do

- [x] merge dependency: #632
- [x] merge dependency: #646
- [x] merge dependency: #650
- [ ]  backport to Coq (including ones from #521)

##### PR overlays

- coq-community/lemma-overloading#63
- imdea-software/fcsl-pcm#24

### DIFF
--- a/mathcomp/ssreflect/path.v
+++ b/mathcomp/ssreflect/path.v
@@ -879,6 +879,31 @@ Arguments map_sort {T T' f leT' leT}.
 Arguments merge_map {T T' f leT}.
 Arguments sort_map {T T' f leT}.
 
+Section SortSeq_in.
+
+Variables (T : Type) (P : {pred T}) (leT : rel T).
+
+Hypothesis leT_total : {in P &, total leT}.
+
+Lemma sort_sorted_in s : all P s -> sorted leT (sort leT s).
+Proof.
+move=> /all_sigP [{}s <-]; rewrite sort_map sorted_map.
+by apply: sort_sorted; move=> [x Px] [y Py]; apply: leT_total.
+Qed.
+
+Hypothesis leT_tr : {in P & &, transitive leT}.
+
+Lemma sorted_sort_in s : all P s -> sorted leT s -> sort leT s = s.
+Proof.
+move=> /all_sigP [{}s <-]; rewrite sort_map sorted_map.
+by move=> /sorted_sort -> // [x Px] [y Py] [z Pz]; apply: leT_tr.
+Qed.
+
+End SortSeq_in.
+
+Arguments sort_sorted_in {T P leT} leT_total {s}.
+Arguments sorted_sort_in {T P leT} leT_tr {s}.
+
 Section EqSortSeq.
 
 Variables (T : eqType) (leT : rel T).
@@ -1069,6 +1094,26 @@ apply/(@irr_sorted_eq _ lt_lex); rewrite /lt_lex /leN //=.
 - by move=> ?; rewrite !(mem_filter, mem_sort).
 Qed.
 
+Lemma sort_stable_in T (P : {pred T}) (leT leT' : rel T) :
+  {in P &, total leT} -> {in P & &, transitive leT'} ->
+  forall s : seq T, all P s -> sorted leT' s ->
+  sorted [rel x y | leT x y && (leT y x ==> leT' x y)] (sort leT s).
+Proof.
+move=> leT_total leT_tr _ /all_sigP [s <-]; rewrite sort_map !sorted_map.
+apply: sort_stable; first by move=> [x Px] [y Py]; apply: leT_total.
+by move=> [x Px] [y Py] [z Pz]; apply: leT_tr.
+Qed.
+
+Lemma filter_sort_in T (P : {pred T}) (leT : rel T) :
+  {in P &, total leT} -> {in P & &, transitive leT} ->
+  forall p s, all P s -> filter p (sort leT s) = sort leT (filter p s).
+Proof.
+move=> leT_total leT_tr p _ /all_sigP [s <-].
+rewrite !(sort_map, filter_map) filter_sort //.
+  by move=> [x Px] [y Py]; apply: leT_total.
+by move=> [x Px] [y Py] [z Pz]; apply: leT_tr.
+Qed.
+
 Section Stability_mask.
 
 Variables (T : Type) (leT : rel T).
@@ -1090,6 +1135,30 @@ Lemma sorted_mask_sort s m :
 Proof. by move/(sorted_sort leT_tr) <-; exact: mask_sort. Qed.
 
 End Stability_mask.
+
+Section Stability_mask_in.
+
+Variables (T : Type) (P : {pred T}) (leT : rel T).
+Hypothesis leT_total : {in P &, total leT}.
+Hypothesis leT_tr : {in P & &, transitive leT}.
+
+Lemma mask_sort_in s m :
+  all P s -> {m_s : bitseq | mask m_s (sort leT s) = sort leT (mask m s)}.
+Proof.
+move=> /all_sigP [{}s <-]; case: (mask_sort (leT := relpre sval leT) _ _ s m).
+- by move=> [x Px] [y Py]; apply: leT_total.
+- by move=> [x Px] [y Py] [z Pz]; apply: leT_tr.
+- by move=> m' m'E; exists m'; rewrite -map_mask !sort_map -map_mask m'E.
+Qed.
+
+Lemma sorted_mask_sort_in s m :
+  all P s -> sorted leT (mask m s) -> {m_s | mask m_s (sort leT s) = mask m s}.
+Proof.
+move=> ? /(sorted_sort_in leT_tr _) <-; last exact: all_mask.
+exact: mask_sort_in.
+Qed.
+
+End Stability_mask_in.
 
 Section Stability_subseq.
 
@@ -1113,6 +1182,37 @@ by case: eqP => // _; rewrite {1}/sort /= lexy /=.
 Qed.
 
 End Stability_subseq.
+
+Section Stability_subseq_in.
+
+Variables (T : eqType) (leT : rel T).
+
+Lemma subseq_sort_in t s :
+  {in s &, total leT} -> {in s & &, transitive leT} ->
+  subseq t s -> subseq (sort leT t) (sort leT s).
+Proof.
+move=> leT_total leT_tr /subseqP [m _ ->].
+have [m' <-] := mask_sort_in leT_total leT_tr m (allss _).
+exact: mask_subseq.
+Qed.
+
+Lemma sorted_subseq_sort_in t s :
+  {in s &, total leT} -> {in s & &, transitive leT} ->
+  subseq t s -> sorted leT t -> subseq t (sort leT s).
+Proof.
+move=> ? leT_tr ? /(sorted_sort_in leT_tr) <-; last exact/allP/mem_subseq.
+exact: subseq_sort_in.
+Qed.
+
+Lemma mem2_sort_in s :
+  {in s &, total leT} -> {in s & &, transitive leT} ->
+  forall x y, leT x y -> mem2 s x y -> mem2 (sort leT s) x y.
+Proof.
+move=> leT_total leT_tr x y lexy; rewrite !mem2E.
+by move/subseq_sort_in; case: (_ == _); rewrite /sort /= ?lexy; apply.
+Qed.
+
+End Stability_subseq_in.
 
 (* Function trajectories. *)
 
--- a/mathcomp/ssreflect/path.v
+++ b/mathcomp/ssreflect/path.v
@@ -882,21 +882,20 @@ Arguments sort_map {T T' f leT}.
 Section SortSeq_in.
 
 Variables (T : Type) (P : {pred T}) (leT : rel T).
+Notation le_sT := (relpre (val : sig P -> _) leT).
 
 Hypothesis leT_total : {in P &, total leT}.
+Let le_sT_total : total le_sT := in2_sig leT_total.
 
 Lemma sort_sorted_in s : all P s -> sorted leT (sort leT s).
-Proof.
-move=> /all_sigP [{}s <-]; rewrite sort_map sorted_map.
-by apply: sort_sorted; move=> [x Px] [y Py]; apply: leT_total.
-Qed.
+Proof. by move=> /all_sigP[? ->]; rewrite sort_map sorted_map sort_sorted. Qed.
 
 Hypothesis leT_tr : {in P & &, transitive leT}.
+Let le_sT_tr : transitive le_sT := in3_sig leT_tr.
 
 Lemma sorted_sort_in s : all P s -> sorted leT s -> sort leT s = s.
 Proof.
-move=> /all_sigP [{}s <-]; rewrite sort_map sorted_map.
-by move=> /sorted_sort -> // [x Px] [y Py] [z Pz]; apply: leT_tr.
+by move=> /all_sigP [{}s ->]; rewrite sort_map sorted_map => /sorted_sort->.
 Qed.
 
 End SortSeq_in.
@@ -1099,19 +1098,16 @@ Lemma sort_stable_in T (P : {pred T}) (leT leT' : rel T) :
   forall s : seq T, all P s -> sorted leT' s ->
   sorted [rel x y | leT x y && (leT y x ==> leT' x y)] (sort leT s).
 Proof.
-move=> leT_total leT_tr _ /all_sigP [s <-]; rewrite sort_map !sorted_map.
-apply: sort_stable; first by move=> [x Px] [y Py]; apply: leT_total.
-by move=> [x Px] [y Py] [z Pz]; apply: leT_tr.
+move=> /in2_sig leT_total /in3_sig leT_tr _ /all_sigP[s ->].
+by rewrite sort_map !sorted_map; apply: sort_stable.
 Qed.
 
 Lemma filter_sort_in T (P : {pred T}) (leT : rel T) :
   {in P &, total leT} -> {in P & &, transitive leT} ->
   forall p s, all P s -> filter p (sort leT s) = sort leT (filter p s).
 Proof.
-move=> leT_total leT_tr p _ /all_sigP [s <-].
-rewrite !(sort_map, filter_map) filter_sort //.
-  by move=> [x Px] [y Py]; apply: leT_total.
-by move=> [x Px] [y Py] [z Pz]; apply: leT_tr.
+move=> /in2_sig leT_total /in3_sig leT_tr p _ /all_sigP[s ->].
+by rewrite !(sort_map, filter_map) filter_sort.
 Qed.
 
 Section Stability_mask.
@@ -1142,20 +1138,21 @@ Variables (T : Type) (P : {pred T}) (leT : rel T).
 Hypothesis leT_total : {in P &, total leT}.
 Hypothesis leT_tr : {in P & &, transitive leT}.
 
+Notation le_sT := (relpre (val : sig P -> _) leT).
+Let le_sT_total : total le_sT := in2_sig leT_total.
+Let le_sT_tr : transitive le_sT := in3_sig leT_tr.
+
 Lemma mask_sort_in s m :
   all P s -> {m_s : bitseq | mask m_s (sort leT s) = sort leT (mask m s)}.
 Proof.
-move=> /all_sigP [{}s <-]; case: (mask_sort (leT := relpre sval leT) _ _ s m).
-- by move=> [x Px] [y Py]; apply: leT_total.
-- by move=> [x Px] [y Py] [z Pz]; apply: leT_tr.
-- by move=> m' m'E; exists m'; rewrite -map_mask !sort_map -map_mask m'E.
+move=> /all_sigP [{}s ->]; case: (mask_sort (leT := le_sT) _ _ s m) => //.
+by move=> m' m'E; exists m'; rewrite -map_mask !sort_map -map_mask m'E.
 Qed.
 
 Lemma sorted_mask_sort_in s m :
   all P s -> sorted leT (mask m s) -> {m_s | mask m_s (sort leT s) = mask m s}.
 Proof.
-move=> ? /(sorted_sort_in leT_tr _) <-; last exact: all_mask.
-exact: mask_sort_in.
+move=> ? /(sorted_sort_in leT_tr _) <-; [exact: mask_sort_in | exact: all_mask].
 Qed.
 
 End Stability_mask_in.
--- a/mathcomp/ssreflect/path.v
+++ b/mathcomp/ssreflect/path.v
@@ -882,7 +882,7 @@ Arguments sort_map {T T' f leT}.
 Section SortSeq_in.
 
 Variables (T : Type) (P : {pred T}) (leT : rel T).
-Notation le_sT := (relpre (val : sig P -> _) leT).
+Let le_sT := relpre (val : sig P -> _) leT.
 
 Hypothesis leT_total : {in P &, total leT}.
 Let le_sT_total : total le_sT := in2_sig leT_total.
@@ -1138,7 +1138,7 @@ Variables (T : Type) (P : {pred T}) (leT : rel T).
 Hypothesis leT_total : {in P &, total leT}.
 Hypothesis leT_tr : {in P & &, transitive leT}.
 
-Notation le_sT := (relpre (val : sig P -> _) leT).
+Let le_sT := relpre (val : sig P -> _) leT.
 Let le_sT_total : total le_sT := in2_sig leT_total.
 Let le_sT_tr : transitive le_sT := in3_sig leT_tr.
 
--- a/mathcomp/ssreflect/path.v
+++ b/mathcomp/ssreflect/path.v
@@ -549,8 +549,8 @@ Lemma sorted_eq_in s1 s2 :
   {in s1 & &, transitive leT} -> {in s1 &, antisymmetric leT} ->
   sorted leT s1 -> sorted leT s2 -> perm_eq s1 s2 -> s1 = s2.
 Proof.
-move=> /in3_sig leT_tr /in2_sig/(_ _ _ _)/val_inj leT_anti ss1 ss2 s1s2.
-move: ss1 ss2 (s1s2); have /all_sigP[s1' ->] := allss s1.
+move=> /in3_sig leT_tr /in2_sig/(_ _ _ _)/val_inj leT_anti + + /[dup] s1s2.
+have /all_sigP[s1' ->] := allss s1.
 have /all_sigP[{s1s2}s2 ->] : all (mem s1) s2 by rewrite -(perm_all _ s1s2).
 by rewrite !sorted_map => ss1' ss2 /(perm_map_inj val_inj)/(sorted_eq leT_tr)->.
 Qed.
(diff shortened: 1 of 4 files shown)


mathcomp/ssreflect/path.v
```diff
@@ -407,6 +407,8 @@ End EqPath.
 
 Section EqHomoPath.
 
+(* TODO: Generalize these lemmas for non-eqType T by introducing a predicate  *)
+(* P : {pred T}.                                                              *)
 Variables (T : eqType) (T' : Type) (f : T -> T') (leT : rel T) (leT' : rel T').
 
 Lemma sub_path_in (e e' : rel T) x s : {in x :: s &, subrel e e'} ->
```
pi8027, line 414: I think that the following generalization of `sub_path_in` is useful:
```coq
Lemma sub_path_in (T T' : Type) (P : {pred T}) (e e' : rel T) x s :
  P x -> all P x -> {in P &, subrel e e'} -> path e x s -> path e' x s.
```
Does it make sense? What should be the naming convention for these two variants of the lemma? (I have the same issue in, e.g., `sort_sorted_in`.)

pi8027:
There are naming inconsistencies in `eq_(path|sorted)(_in)` lemmas. :(
```coq
eq_path
     : forall (T : Type) (e e' : T -> T -> bool), e =2 e' -> path e =2 path e'

eq_path_in
     : forall (T : eqType) (e e' : rel T) (x : T) (s : seq T),
       {in x :: s &, e =2 e'} -> path e x s = path e' x s

eq_sorted
     : forall (T : eqType) (leT : rel T),
       transitive leT ->
       antisymmetric leT ->
       forall s1 s2 : seq T,
       sorted leT s1 -> sorted leT s2 -> perm_eq s1 s2 -> s1 = s2

eq_sorted_in
     : forall (T : eqType) (leT : rel T) (s1 : seq_predType T) (s2 : seq T),
       {in s1 & &, transitive leT} ->
       {in s1 &, antisymmetric leT} ->
       sorted leT s1 -> sorted leT s2 -> perm_eq s1 s2 -> s1 = s2
```

pi8027:
Also `path`, `cycle`, and `sorted` share some similar properties but many of them are actually missing as lemmas, e.g., the `cycle` variant of `eq_path_in` and `(homo|mono)_path(_in)`, and the `sorted` variants of `eq_path(_in)`. I guess it would be nice to move `sorted` after the definition of `cycle` to get a better view.

mathcomp/ssreflect/path.v
```diff
@@ -407,7 +475,9 @@ End EqPath.
 
 Section EqHomoPath.
 
-Variables (T : eqType) (T' : Type) (f : T -> T') (leT : rel T) (leT' : rel T').
+(* TODO: Generalize these lemmas for non-eqType T by introducing a predicate  *)
+(* P : {pred T}.                                                              *)
+Variables (T : eqType) (T' : Type) (f : T -> T').
 
 Lemma sub_path_in (e e' : rel T) x s : {in x :: s &, subrel e e'} ->
    path e x s -> path e' x s.
```
pi8027, line 483: Since my previous comment has become outdated, I repeat it. I think that the following generalization of `sub_path_in` is useful:
```coq
Lemma sub_path_in (T T' : Type) (P : {pred T}) (e e' : rel T) x s :
  P x -> all P x -> {in P &, subrel e e'} -> path e x s -> path e' x s.
```
This generalization may apply to all the lemmas in this section. Does it make sense? What should be the naming convention for these two variants of the lemma? (I have the same issue in the new lemmas, e.g., `sort_sorted_in`.)

pi8027:
I don't think I can make more progress without some feedback. May I have some comment on the above points (naming inconsistencies, a possible generalization of `sub_path_in`, and its naming convention)?

mathcomp/ssreflect/path.v
CohenCyril: Yes!

CohenCyril: > There are naming inconsistencies in `eq_(path|sorted)(_in)` lemmas. :(

I'm not sure what the inconsistency is... `path` being a predicate `eq_path` is its congruence lemma, on the other hand, `sorted` being a unary predicate `eq_sorted` characterizes the equality of sorted sequences.

CohenCyril:
> There are naming inconsistencies in `eq_(path|sorted)(_in)` lemmas. :(

Nevermind, yes there is an inconsistency...

pi8027:
> Nevermind, yes there is an inconsistency... I guess `eq_sorted` should be renamed `sorted_eq` then...

Sure. So we will be able to add `eq_sorted(_in)` consistent with `eq_path(_in)` in 1.13.

mathcomp/ssreflect/path.v
```diff
@@ -771,22 +998,25 @@ case: s => [|x s] //; have [s1 pp qq] := perm_iota_sort leT x (x :: s).
...
-End EqHomoSortSeq.
-
-Arguments homo_sorted_in {T T' f leT leT'}.
-Arguments mono_sorted_in {T T' f leT leT'}.
+(* TODO: Generalize the following lemma for non-eqType T'. *)
+Lemma sort_map_in (T T' : eqType) f (leT : rel T) (leT' : rel T') (s : seq T) :
+  {in s &, antisymmetric (relpre f leT')} ->
+  {in s & &, transitive (relpre f leT')} -> {in s &, total leT} ->
+  {in s &, {homo f : x y / leT x y >-> leT' x y}} ->
+  sort leT' [seq f x | x <- s] = [seq f x | x <- sort leT s].
```
pi8027, line 1006: Also, I had a hard time to generalize this lemma. The point is that the current proof uses `perm_eq` of `seq T'`, so `T'` is required to be an `eqType` but it shouldn't. `T` can also be generalized to a `Type` by introducing a predicate `P : {pred T}`.

mathcomp/ssreflect/path.v
CohenCyril: I think I'd state it as 
```coq
Lemma sub_path_in (T T' : Type) (P : {pred T}) (e e' : rel T) x s :
  all P (x :: s) -> {in P &, subrel e e'} -> path e x s -> path e' x s.
```
so that we can apply it to
```coq
Lemma all_seq {T : eqType} {s : seq T} : all (mem s) s. Proof. exact/allP. Qed.
```
and get back the previous behaviour...

CohenCyril: On second thought, I'm wondering whether the more general versions should be in a separate module, because applying `all_seq` by hand all the time seems a little bit... costly :confused:

pi8027: It seems that both `all (mem s) s` and `all (pred_of_seq s) s` appear in goals, and then the above `all_seq` can be problematic. :(
```coq
Require Import all_ssreflect.

Lemma allss (T : eqType) (s : seq T) : all (mem s) s. Proof. exact/allP. Qed.

Goal forall (T : eqType) (x : T) s, all (pred_of_seq (x :: s)) (x :: s).
Proof.
move=> T x s.
Fail apply: allss. (* ! *)
exact: (allss _ (x :: s)).
Qed.
```
I think that the above `Fail apply:` unfolds `pred_of_seq (x :: s)` first and then fails to unify it with `mem _`.

pi8027: I found a solution to make `allss` work both for `all (mem s) s` and `all (pred_of_seq s) s` and declared it as a hint. See `seq.v`.

pi8027:
Deprecating `eq_sorted_irr` breaks fcsl-pcm and lemma-overloading, because the deprecation facility does not support the `(ident := term)` syntax for explicit applications. https://github.com/imdea-software/fcsl-pcm/blob/1d5b78248fb1d5bec3553d64f92656e973f2b6de/pcm/unionmap.v#L3573 https://github.com/coq-community/lemma-overloading/blob/05a9f467049822eb14b3784ddaa21ddc0252c92a/theories/heaps.v#L1199

mathcomp/ssreflect/path.v
pi8027: Except for this point, this PR seems to be almost good enough to merge. So I would remove this lemma for now and try to add it later in another PR if there is no immediate solution.

pi8027:
I will write CHANGELOG entries later today. This PR is ready for review except for that.

pi8027:
> I will write CHANGELOG entries later today. This PR is ready for review except for that.

Done.

mathcomp/ssreflect/seq.v
```diff
@@ -1428,6 +1430,9 @@ Arguments count_memPn {T x s}.
 Arguments uniqPn {T} x0 {s}.
 Arguments uniqP {T} x0 {s}.
 
+Hint Extern 0 (is_true (all _ _)) =>
+  exact: (allss : forall T s, all (mem_seq s) s) : core.
```
chdoc, line 1434: Is it really necessary to add a global external hint for this?

chdoc (commented):
I gave it first pass, and this looks largely reasonable to me. However, I haven't taken the time yet to look at the individual proofs. Moreover, it looks to me that this PR changes (much) more than just generalize the sorting lemmas. I would advocate for splitting this PR into more focused (and hence smaller) PRs, so that each part can be given proper consideration.

pi8027:
> Moreover, it looks to me that this PR changes (much) more than just generalize the sorting lemmas. I would advocate for splitting this PR into more focused (and hence smaller) PRs, so that each part can be given proper consideration.

@chdoc Sure. I separated off the first part of this PR as #632.

mathcomp/ssreflect/order.v
```diff
@@ -2750,12 +2750,15 @@ rewrite andbF; apply/and5P=> [[ne_nm lenm _ _ le_ms]]; case/negP: ne_nm.
...
-Lemma eq_sorted_lt s1 s2 : sorted lt s1 -> sorted lt s2 -> s1 =i s2 -> s1 = s2.
-Proof. by apply: eq_sorted_irr => //; apply: lt_trans. Qed.
+Lemma sorted_eq_lt s1 s2 : sorted <%O s1 -> sorted <%O s2 -> s1 =i s2 -> s1 = s2.
+Proof. by apply: sorted_eq_irr => //; apply: lt_trans. Qed.
 
-Lemma eq_sorted_le s1 s2 : sorted le s1 -> sorted le s2 ->
-  perm_eq s1 s2 -> s1 = s2.
-Proof. by apply: eq_sorted; [apply: le_trans|apply: le_anti]. Qed.
+Lemma sorted_eq_le s1 s2 :
+  sorted <=%O s1 -> sorted <=%O s2 -> perm_eq s1 s2 -> s1 = s2.
```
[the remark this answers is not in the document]
pi8027: I replaced `eq_sorted` with `sorted_eq` blindly but `sorted_(le|lt)_eq` seems to make much sense here because they state properties of lists sorted by `<=%O` and `<%O`. So should we rename `sorted_eq_irr(_in)` to `sorted_irr_eq(_in)`? What do you think, @CohenCyril?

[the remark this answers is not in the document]
CohenCyril: These names do not make sense to me, `sorted` and `lt`/`le` must be next to each other.
I liked `eq_sorted_(le|lt)`, but I would be fine with `sorted_(le|lt)_eq` too.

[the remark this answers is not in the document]
CohenCyril: (and same rationale with `sorted_eq_irr`)

mathcomp/ssreflect/seq.v
CohenCyril: I guess it follows from a problem from @pi8027 encountered and that I cannot find on github :disappointed: 
However, it is forbidden to use `exact:` in a `Hint Extern`, since `exact` might call `ssrdone` which calls makes use of level 0 hints, potentially causing a loop in some weird cases. You may use `solve [apply: stuff]` instead.
(BTW I notice `exact:` is used in `ssrnum.v` which I accidentally let through...)

CohenCyril: It should probably be documented as a comment in the code though...

pi8027: > I guess it follows from a problem from @pi8027 encountered and that I cannot find on github 

The problem I observed in the discussion following from https://github.com/math-comp/math-comp/pull/601#discussion_r501516139 was that both `all (mem s) s` and `all (pred_of_seq s) s` appear in goals and do not reduce to each other. So stating `allss` in the form of one of them makes `apply: allss` failing for the other case. Since both `mem` and `pred_of_seq` reduce to `mem_seq`, I had to put this explicit type annotation here. Since the `Hint` facility lets us use syntactic matching to find the rules to apply, I had also to use `Hint Extern` to give that pattern explicitly (see #319).

> However, it is forbidden to use `exact:` in a `Hint Extern`, since `exact` might call `ssrdone` which calls use of level 0 hints, potentially causing a loop in some weird cases. You may use `solve [apply: stuff]` instead.
> (BTW I notice `exact:` is used in `ssrnum.v` which I accidentally let through...)

Sure. I will fix them.

mathcomp/ssreflect/order.v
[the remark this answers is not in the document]
pi8027: Fixed.

mathcomp/ssreflect/path.v
```diff
@@ -649,85 +661,150 @@ rewrite /sort; move: [::] {2}_.+1 (ltnSn (size s)./2) => ss n.
...
-elim=> //= x2 s2 IHs2 x /andP[le_x_x1 ord_s1] /andP[le_x_x2 ord_s2].
-case: ifP => le_x21 /=; first by rewrite le_x_x1 {}IHs1 //= le_x21.
-by rewrite le_x_x2 IHs2 //=; have:= leT_total x1 x2; rewrite le_x21 /= => ->.
+rewrite sortE -[s in RHS]/(flatten [::] ++ s).
+elim: s [::] => [|x s ih] ss /=.
+  by elim: ss [::] => //= s ss ih s'; rewrite ih !all_cat all_merge; bool_congr.
+rewrite {}ih -cat_rcons !all_cat -cats1; congr andb.
+elim: ss [:: x] => {x s} [|[|x s] ss ih] s' //=;
+  rewrite ?ih !all_cat ?all_merge /= -?andbA; bool_congr.
 Qed.
```
chdoc, line 672: This is another instance where it would be really useful to be able to use the fact that `sort s` and `s` have the same elements, even if one does not have a decidable equality (#638). This isn't a blocker, I'm merely collecting examples for the aforementioned issue.

chdoc:
@pi8027 , is this PR ready for review now?

pi8027:
> @pi8027 , is this PR ready for review now?

@chdoc Yes.

mathcomp/ssreflect/path.v
pi8027: As in https://github.com/math-comp/math-comp/pull/632#discussion_r520492284, `mkseq_nth` is useful to avoid induction here.
```coq
Lemma all_sort (T : Type) (P : {pred T}) (leT : rel T) s :
  all P (sort leT s) = all P s.
Proof.
case: s => // x s; move: (x :: s) => {s} s.
rewrite -(mkseq_nth x s) sort_map !all_map.
by apply: perm_all; rewrite perm_sort.
Qed.
```

pi8027:
Now I think `perm_iota_sort` is not convenient enough to perform the `mkseq` technique for `sort`. I actually rather use `mkseq_nth` often. Let me investigate before review.

mathcomp/ssreflect/path.v
chdoc: 3 (short) lines rather than 6, that's much better. Also, the use of `mkseq_nth` makes it easier to find this should there ever be list membership without decidable equality in mathcomp. Great!

CohenCyril:
There are two ways to add the `_in` variants,
- one way is exactly what you did: prove the `_in` versions first, and then apply them to the not `_in` counterparts.
- the other one is to keep the original code as such and extend it to the `_in` variants, by applying the original lemmas to `sig P` and using parametricity results (such as `path_map`, `sorted_map`, `sort_map`, etc)
I wonder if the second approach wouldn't be more compact, maybe at the cost of some reusable machinery... (like `inW` etc), @pi8027 did you try it?

pi8027:
@CohenCyril I thought about the second possibility but did not try it yet. Let me try. Also, I intend to separate off the renaming of `eq_sorted` lemmas as another PR to make it easy to compare these two approaches.

pi8027:
CI fails since coq.inria.fr is down🤦‍♂️

pi8027:
Progress report: before working on another approach to add the `_in` lemmas suggested by @CohenCyril, I'm discovering some possible simplifications and generalizations of existing proofs (namely the first commit of this PR). I will probably open it as another PR.

mathcomp/ssreflect/path.v
```diff
@@ -375,6 +424,42 @@ Arguments mono_path {T T' f e e' x s}.
...
+
+Lemma ltn_index s : sorted leT s ->
+  {in s &, forall x y, index x s < index y s -> leT x y}.
+Proof. by apply: ltn_index_in; apply: in3W. Qed.
+
+Hypothesis leT_refl : reflexive leT.
+
+Lemma leq_index s : sorted leT s ->
+  {in s &, forall x y, index x s <= index y s -> leT x y}.
+Proof. by apply: leq_index_in; [apply: in3W | apply: in1W]. Qed.
```
pi8027, line 459: The names of these lemmas `(ltn|leq)_index` are too short in my opinion. Shouldn't we rename them to `sorted_(ltn|leq)_index`, so that these properties specialized to a `porderType` can be called `(le|lt)_sorted_(ltn|leq)_index`?

CohenCyril: Yes indeed, note that `lt_sorted_` versions, or more generally either "irreflexive versions" or "reflexive + uniq" versions should be stronger, i.e. we should get an equality i.e. a `mono`, (and also `mono` for `sorted_le_nth` under the same conditions). I was planning to add it one day, and I will do it in another PR (unless you desire to do it instead, if so please tell me), and from this perspective the renaming you suggest is a nice anticipation.

pi8027: Sure. I will rename them, but will not add the irreflexive versions of them by myself (at least for now). I'm also preparing a PR to address #622 at the same time, but I will not touch this part (including `sorted_(le|lt)_nth` lemmas).

pi8027: > ... from this perspective the renaming you suggest is a nice anticipation and maybe it should even go further renaming the current `sorted_lt_nth` to `sorted_ltn_nth` (so that `sorted_lt_nth` becomes available for the specialization of `sorted_irr_nth` to `<%O`).

`lt`/`le` of `sorted_(lt|le)_nth` is about indices, not about elements of `s`. So I think `irr` should come before `sorted` (e.g., `irr_sorted_lt_nth`), but it would still be nice to replace these `lt` and `le` with `ltn` and `leq` respectively.

pi8027: Addressed in #650.

pi8027:
>  the other one is to keep the original code as such and extend it to the `_in` variants, by applying the original lemmas to `sig P` and using parametricity results (such as `path_map`, `sorted_map`, `sort_map`, etc)

This one seems quite successful (see https://github.com/pi8027/math-comp/commit/4c794d28703abe14c60fce540374a8ad7a5808af). Since these lemmas will not likely to be used frequently and their proofs are reasonably short, we probably don't have to add them in `path.v`. But, at least, I would like to document this proof technique somewhere. Would `sort_map(_in)` be a good example?

CohenCyril:
> This one seems quite successful (see [pi8027@4c794d2](https://github.com/pi8027/math-comp/commit/4c794d28703abe14c60fce540374a8ad7a5808af)). Since these lemmas will not likely to be used frequently and their proofs are reasonably short, we probably don't have to add them in `path.v`. But, at least, I would like to document this proof technique somewhere. Would `sort_map(_in)` be a good example?

Actually, I think the best documentation would come from factoring out the proofs that transitivity, totality, reflexivity, irreflexivity, antisymmetry (and the ones I maybe forgot) "in" `P`, propagate to `sig P` and documenting the proof technique there and pointing to `sort_map(_in)` for an example of usage.

CohenCyril:
cf https://github.com/CohenCyril/math-comp/tree/sorting_in-param

CohenCyril:
My impression is that in previous versions of this, you had many more `_in` variants. I cannot find the previous changelog anymore to compare, but is that right?

pi8027:
> My impression is that in previous versions of this, you had many more `_in` variants. I cannot find the previous changelog anymore to compare, but is that right?

Right. Here I have a previous version: https://github.com/pi8027/math-comp/commit/024cb1b6455fa40ea7cb902909a311033fb30108. I will put some of them back, but I'm not sure if we need all of them.

mathcomp/ssreflect/path.v
```diff
@@ -475,9 +527,48 @@ Qed.
...
+Proof.
+move=> /in3_sig leT_tr /in1_sig leT_irr; case/all_sigP: (allss s) => s' ->.
+by rewrite sorted_map (map_inj_uniq val_inj); exact: sorted_uniq.
+Qed.
+
+Lemma sorted_eq_in s1 s2 :
+  {in s1 & &, transitive leT} -> {in s1 &, antisymmetric leT} ->
+  sorted leT s1 -> sorted leT s2 -> perm_eq s1 s2 -> s1 = s2.
+Proof.
+move=> /in3_sig leT_tr /in2_sig leT_asym + + /[dup] perm_s12.
```
pi8027, line 552: The behavior of this intro pattern has been changed between Coq 8.11 and Coq 8.12.🙁

In Coq 8.11:
```
  perm_s12 : perm_eq s1 s2
  ============================
  sorted leT s1 -> sorted leT s2 -> perm_eq s1 s2 -> s1 = s2
```

In Coq 8.12:
```
  perm_s12 : perm_eq s1 s2
  ============================
  perm_eq s1 s2 -> sorted leT s1 -> sorted leT s2 -> s1 = s2
```

CohenCyril: CC @gares

CohenCyril: The behaviour in 8.12 is not the right one... This kind of calls for a backport...

mathcomp/ssreflect/path.v
```diff
@@ -475,9 +527,48 @@ Qed.
...
+
+Lemma sorted_eq_in s1 s2 :
+  {in s1 & &, transitive leT} -> {in s1 &, antisymmetric leT} ->
+  sorted leT s1 -> sorted leT s2 -> perm_eq s1 s2 -> s1 = s2.
+Proof.
+move=> /in3_sig leT_tr /in2_sig leT_asym + + perm_s12; move: (perm_s12).
+case/all_sigP: (allss s1) => s1' ->; move: (allss s1).
+rewrite (perm_all _ perm_s12) => /all_sigP [s2' ->] /(perm_map_inj val_inj).
+rewrite !sorted_map => {}perm_s12 s1_sorted s2_sorted; congr map.
+by apply: sorted_eq s1_sorted s2_sorted perm_s12 => // ? ? /leT_asym /val_inj.
```
CohenCyril, line 556:
```suggestion
move=> /in3_sig leT_tr /in2_sig/(_ _ _ _)/val_inj leT_anti ss1 ss2 s1s2.
move: ss1 ss2 (s1s2); have /all_sigP[s1' ->] := allss s1.
have /all_sigP[{s1s2}s2 ->] : all (mem s1) s2 by rewrite -(perm_all _ s1s2).
by rewrite !sorted_map => ss1' ss2 /(perm_map_inj val_inj)/(sorted_eq leT_tr)->.
```

mathcomp/ssreflect/path.v
```diff
@@ -475,9 +527,48 @@ Qed.
...
+Qed.
+
+Lemma irr_sorted_eq_in s1 s2 :
+  {in s1 & &, transitive leT} -> {in s1, irreflexive leT} ->
+  sorted leT s1 -> sorted leT s2 -> s1 =i s2 -> s1 = s2.
+Proof.
+move=> /in3_sig leT_tr /in1_sig leT_irr + + eq_s12; move: (eq_s12).
+case/all_sigP: (allss s1) => s1' ->; move: (allss s1).
+rewrite (eq_all_r eq_s12) => /all_sigP [s2' ->].
+rewrite !sorted_map => {}eq_s12 s1_sorted s2_sorted; congr map.
```
CohenCyril, line 565: Even if it is slightly longer, this documents the proof in a better way.
```suggestion
move=> /in3_sig leT_tr /in1_sig leT_irr + + s1s2; move: (s1s2).
have /all_sigP[s1' ->] := allss s1.
have /all_sigP[s2' ->] : all (mem s1) s2 by rewrite -(eq_all_r s1s2).
rewrite !sorted_map => {}s1s2 s1_sorted s2_sorted; congr map.
```

mathcomp/ssreflect/path.v
```diff
@@ -233,6 +242,31 @@ Qed.
...
+move=> Pxs; apply/idP/idP => [xs|/andP[/path_min_sorted<-//]].
+by rewrite (order_path_min_in leT_tr) //; apply: path_sorted xs.
+Qed.
+
+Lemma sorted_ltn_nth_in x0 s : all P s -> sorted leT s ->
+  {in [pred n | n < size s] &, {homo nth x0 s : i j / i < j >-> leT i j}}.
+Proof.
+elim: s => // x s ihs Pxs path_xs [|i] [|j] //=; rewrite -!topredE /= !ltnS.
+- by move=> _ js _; apply/all_nthP/js/order_path_min_in.
+- by apply/ihs/path_sorted/path_xs; case/andP: Pxs.
```
CohenCyril, line 257:
```suggestion
  by move=> _ js _; apply/all_nthP/js/order_path_min_in.
by apply/ihs/path_sorted/path_xs; case/andP: Pxs.
```

mathcomp/ssreflect/path.v
```diff
@@ -913,14 +1027,26 @@ Proof. by apply: perm_mem; rewrite perm_sort. Qed.
...
   forall s1 s2, reflect (sort leT s1 = sort leT s2) (perm_eq s1 s2).
 Proof.
-move=> leT_total leT_tr leT_asym s1 s2.
-apply: (iffP idP) => eq12; last by rewrite -perm_sort eq12 perm_sort.
-apply: (sorted_eq leT_tr leT_asym); rewrite ?sort_sorted //.
-by rewrite perm_sort (permPl eq12) -perm_sort.
+by move=> *; apply: perm_sort_inP; [apply: in2W | apply: in3W | apply: in2W].
 Qed.
 
 End EqSortSeq.
```
CohenCyril, line 1052: I'd rather keep `perm_sortP` untouched, and add
```coq
Lemma perm_sort_inP (T : eqType) (leT : rel T) (s1 s2 : seq T) :
  {in s1 &, total leT} -> {in s1 & &, transitive leT} ->
  {in s1 &, antisymmetric leT} ->
  reflect (sort leT s1 = sort leT s2) (perm_eq s1 s2).
Proof.
move=> /in2_sig leT_total /in3_sig leT_tr /in2_sig/(_ _ _ _)/val_inj leT_asym.
apply: (iffP idP) => s1s2; last by rewrite -(perm_sort leT) s1s2 perm_sort.
move: (s1s2); have /all_sigP[s1' ->] := allss s1.
have /all_sigP[{s1s2}s2 ->] : all (mem s1) s2 by rewrite -(perm_all _ s1s2).
by rewrite !sort_map => /(perm_map_inj val_inj) /(perm_sortP leT_total)->.
Qed.
```
after the end of the section.

CohenCyril (approved):
Great!

CohenCyril:
@chdoc I'm happy with the current state of this PR, if you are too, I will merge.

chdoc (approved):
LGTM!

PR #626, opened 2020-10-30
Generalizing inj_card_onto and inj_card_bij.

CohenCyril:

This weakens the hypothesis of the two above lemmas

##### Things done/to do

### DIFF
--- a/CHANGELOG_UNRELEASED.md
+++ b/CHANGELOG_UNRELEASED.md
@@ -319,6 +319,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   + `>=< y` stands for `[pred x | x >=< y]`
 
 - in `ssrint.v`, generalized `mulpz` for any `ringType`.
+- In `fintype.v`, lemmas `inj_card_onto` and `inj_card_bij` take a
+  weaker hypothesis (i.e. `#|T| >= #|T'|` instead of `#|T| = #|T'|`
+  thanks to `leq_card` under injectivity assumption).
 
 ### Renamed
 
--- a/mathcomp/ssreflect/fintype.v
+++ b/mathcomp/ssreflect/fintype.v
@@ -1300,7 +1300,9 @@ Qed.
 
 Lemma leq_card : #|T| <= #|T'|. Proof. exact: (leq_card_in (in2W _)). Qed.
 
-Hypothesis card_range : #|T| = #|T'|.
+Hypothesis card_range : #|T| >= #|T'|.
+
+Let eq_card : #|T| = #|T'|. Proof. by apply/eqP; rewrite eqn_leq leq_card. Qed.
 
 Lemma inj_card_onto y : y \in codom f.
 Proof. by move: y; apply/subset_cardP; rewrite ?card_codom ?subset_predT. Qed.


chdoc:
I'm a bit surprised that this change does not cause any breakage whatsoever. One reason might be that the lemma, regardless of the premise, is too weak to be of much use. I only found one use for this lemma so far, and that was to prove the following strengthening. (I'm including the comment I added when proving this 18 months ago :smile:)
```coq
(** TOTHINK: [#|P| <= #|aT|] would suffice, the other direction is
implied. But the same is true for [inj_card_onto]. *)
Lemma inj_card_onto_pred (aT rT : finType) (f : aT -> rT) (P : pred rT) : 
  injective f -> (forall x, f x \in P) -> #|aT| = #|P| -> {in P, forall y, y \in codom f}.
Proof.
move=> f_inj fP E y yP.
pose rT' := { y : rT | P y}; pose f' (x:aT) : rT' := Sub (f x) (fP x).
have/inj_card_onto f'_inj : injective f' by  move => x1 x2 []; apply: f_inj.
rewrite -[#|P|]card_sig in E; have/mapP[x _ [] ->] := f'_inj E (Sub y yP). 
exact: codom_f. 
Qed.
```
This one I use a handful of times. I guess one could also have a predicate `A` on the argument type and weaken this to {in A&, injective f} and f a homomorphism from A to P.

CohenCyril:
@chdoc note that two lemmas were generalized, and the second one `inj_card_bij`, is used slightly more often. I guess one reason no breakage occurs is that every proof solving `#|T| = #|T'|` can solve `#|T| >= #|T'|` just as easily.
As for your variation on `inj_card_onto`, I'd say it would be a nice addition indeed, in a subsequent PR.
I suggest the following statements + proofs (rather than going all the way through `inj_card_onto`):
```coq
Lemma inj_image_eq (aT rT : finType) (f : aT -> rT) 
    (D : {pred aT}) (P : {pred rT}) : 
    {in D &, injective f} -> image f D \subset P -> #|P| <= #|D| ->
  image f D =i P.
Proof.
move=> f_inj fP leP; apply/subset_cardP => //; apply/eqP.
by rewrite eqn_leq subset_leq_card//= (leq_trans leP)// card_in_image.
Qed.

Lemma inj_codom_eq (aT rT : finType) (f : aT -> rT) (P : {pred rT}) : 
  injective f -> codom f \subset P -> #|P| <= #|aT| -> codom f =i P.
Proof. by move=> f_inj; apply/inj_image_eq; apply: in2W. Qed.
```

chdoc:
> I guess one reason no breakage occurs is that every proof solving `#|T| = #|T'|` can solve `#|T| >= #|T'|` just as easily.

I was a bit worried that there could be places where the the equality is shown by showing `#|T| >= #|T'|` and `#|T| <= #|T'|` separately, possibly duplicating your argument. But neither lemma is used withing mathcomp (outside of fintype), so even if this is the case, it would be of no concern to this PR.

chdoc (approved):
LGTM!

pi8027 (approved):
LGTM.

chdoc:
@pi8027 then please merge, because I cannot :smirk:

PR #630, opened 2020-11-04
New lemmas about allpairs

CohenCyril:
##### Motivation for this change

Fixes #497 en passant.

The former convention seemed to implicitly consit in dropping suffix
`l` when the `r` side does not exist as an equality (but rather as a
`=i` or now a `perm_eq`). If we want to change this convention, I can
open another PR after this one is merged, but it will involve even
more deprecation.

##### Things done/to do

- [x] added corresponding entries in `CHANGELOG_UNRELEASED.md` (do not edit former entries)
- ~added corresponding documentation in the headers~

##### Automatic note to reviewers

Read [this Checklist](https://github.com/math-comp/math-comp/wiki/Checklist-for-following,-reviewing-and-playing-with-a-PR#checklist-for-reviewing-a-pr) and make sure there is a milestone.

PR #632, opened 2020-11-05
Reorganize, generalize, and add lemmas about `path`, `cycle`, and `sorted`

pi8027:

Part of #601.

- Add `allss` and `all_mask` lemmas in `seq.v`.
- Since `path`, `cycle`, and `sorted` share similar properties, these lemmas have been relocated in the same place to improve the visibility. Some missing lemmas have also been discovered and added.
- Generalize `sub_path_in`, `sub_sorted_in`, and `eq_path_in` for non-`eqType` `T` by introducing a predicate `P : {pred T}`.

##### Things done/to do

### DIFF
--- a/mathcomp/ssreflect/path.v
+++ b/mathcomp/ssreflect/path.v
@@ -137,55 +137,272 @@ Qed.
 Lemma rotr_cycle p : cycle (rotr n0 p) = cycle p.
 Proof. by rewrite -rot_cycle rotrK. Qed.
 
+Definition sorted s := if s is x :: s' then path x s' else true.
+
+Lemma path_sorted x s : path x s -> sorted s.
+Proof. by case: s => //= y s /andP[]. Qed.
+
+Lemma path_min_sorted x s : all (e x) s -> path x s = sorted s.
+Proof. by case: s => //= y s /andP [->]. Qed.
+
 End Path.
 
-Lemma eq_path e e' : e =2 e' -> path e =2 path e'.
-Proof. by move=> ee' x p; elim: p x => //= y p IHp x; rewrite ee' IHp. Qed.
+Section SubPath_in.
 
-Lemma eq_cycle e e' : e =2 e' -> cycle e =1 cycle e'.
-Proof. by move=> ee' [|x p] //=; apply: eq_path. Qed.
+Variable (P : {pred T}) (e e' : rel T).
+Hypothesis (ee' : {in P &, subrel e e'}).
 
-Lemma sub_path e e' : subrel e e' -> forall x p, path e x p -> path e' x p.
-Proof. by move=> ee' x p; elim: p x => //= y p IHp x /andP[/ee'-> /IHp]. Qed.
+Lemma sub_path_in x s : all P (x :: s) -> path e x s -> path e' x s.
+Proof.
+by elim: s x => //= y s ihs x /and3P [? ? ?] /andP [/ee' -> //]; apply/ihs/andP.
+Qed.
 
-Lemma rev_path e x p :
+Lemma sub_cycle_in s : all P s -> cycle e s -> cycle e' s.
+Proof.
+case: s => //= x s /andP [Px Ps].
+by apply: sub_path_in; rewrite /= all_rcons Px.
+Qed.
+
+Lemma sub_sorted_in s : all P s -> sorted e s -> sorted e' s.
+Proof. by case: s => //; apply: sub_path_in. Qed.
+
+End SubPath_in.
+
+Section EqPath_in.
+
+Variable (P : {pred T}) (e e' : rel T).
+Hypothesis (ee' : {in P &, e =2 e'}).
+
+Let e_e' : {in P &, subrel e e'}. Proof. by move=> ? ? ? ?; rewrite ee'. Qed.
+Let e'_e : {in P &, subrel e' e}. Proof. by move=> ? ? ? ?; rewrite ee'. Qed.
+
+Lemma eq_path_in x s : all P (x :: s) -> path e x s = path e' x s.
+Proof. by move=> Pxs; apply/idP/idP; apply: sub_path_in Pxs. Qed.
+
+Lemma eq_cycle_in s : all P s -> cycle e s = cycle e' s.
+Proof. by move=> Ps; apply/idP/idP; apply: sub_cycle_in Ps. Qed.
+
+End EqPath_in.
+
+Section SubPath.
+
+Variables e e' : rel T.
+
+Lemma sub_path : subrel e e' -> forall x p, path e x p -> path e' x p.
+Proof. by move=> ? ? ?; apply/sub_path_in/all_predT; apply: in2W. Qed.
+
+Lemma sub_cycle : subrel e e' -> subpred (cycle e) (cycle e').
+Proof. by move=> ee' [] // ? ?; apply: sub_path. Qed.
+
+Lemma sub_sorted : subrel e e' -> subpred (sorted e) (sorted e').
+Proof. by move=> ee' [] //=; apply: sub_path. Qed.
+
+Lemma eq_path : e =2 e' -> path e =2 path e'.
+Proof. by move=> ? ? ?; apply/eq_path_in/all_predT; apply: in2W. Qed.
+
+Lemma eq_cycle : e =2 e' -> cycle e =1 cycle e'.
+Proof. by move=> ee' [] // ? ?; apply: eq_path. Qed.
+
+End SubPath.
+
+Section Transitive_in.
+
+Variables (P : {pred T}) (leT : rel T).
+Hypothesis leT_tr : {in P & &, transitive leT}.
+
+Lemma path_mask_in x m s :
+  all P (x :: s) -> path leT x s -> path leT x (mask m s).
+Proof.
+elim: m s x => [|[] m ih] [|y s] x //=.
+  by case/and3P=> ? ? ? /andP [-> /ih ->] //; apply/andP.
+case/andP=> Px Pys /andP [xy ys]; case/andP: (Pys) => Py Ps.
+case: (mask _ _) (all_mask m Ps) (ih s y Pys ys) => //=.
+by move=> z t /andP [Pz Pt] /andP [] /(leT_tr Py Px Pz xy) ->.
+Qed.
+
+Lemma path_filter_in x a s :
+  all P (x :: s) -> path leT x s -> path leT x (filter a s).
+Proof. by move=> Pxs; rewrite filter_mask; exact: path_mask_in. Qed.
+
+Lemma cycle_mask_in m s : all P s -> cycle leT s -> cycle leT (mask m s).
+Proof.
+case: (resize_mask m s) => {m} m sizeE ->.
+elim: m s sizeE => [|[]m ih] [|x s] //= [sizeE] /andP [Px Ps].
+- rewrite -!cats1 -(mask_cat [:: true] [:: x]) //.
+  by apply: path_mask_in; rewrite /= all_cat /= Px Ps.
+- move=> xsx; apply: ih => // {sizeE}; case: s xsx Ps => //= y s.
+  rewrite !rcons_path => /and3P [xy -> sx] /andP [Py Ps] /=.
+  apply: leT_tr sx xy => //.
+  by elim: s y Ps Py => //= z s ih y /andP [Pz Ps] _; apply: ih.
+Qed.
+
+Lemma cycle_filter_in a s : all P s -> cycle leT s -> cycle leT (filter a s).
+Proof. move=> Ps; rewrite filter_mask; exact: cycle_mask_in. Qed.
+
+Lemma sorted_mask_in m s : all P s -> sorted leT s -> sorted leT (mask m s).
+Proof.
+elim: m s => [|[] m ih] [|x s] //= Pxs; first exact: path_mask_in.
+by move/path_sorted/ih; apply; case/andP: Pxs.
+Qed.
+
+Lemma sorted_filter_in a s : all P s -> sorted leT s -> sorted leT (filter a s).
+Proof. rewrite filter_mask; exact: sorted_mask_in. Qed.
+
+End Transitive_in.
+
+Section Transitive.
+
+Variable (leT : rel T).
+Hypothesis leT_tr : transitive leT.
+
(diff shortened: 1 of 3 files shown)


mathcomp/ssreflect/path.v
```diff
@@ -137,55 +137,284 @@ Qed.
...
+Arguments mono_cycle {T T' f e e'}.
+Arguments mono_sorted {T T' f e e'}.
+
+Section HomoPath_in.
+
+Variables (T T' : Type) (P : {pred T}) (f : T -> T') (e : rel T) (e' : rel T').
+
+Lemma homo_path_in x s : {in P &, {homo f : x y / e x y >-> e' x y}} ->
+  all P (x :: s) -> path e x s -> path e' (f x) (map f s).
+Proof. by move=> f_mono; rewrite path_map; apply: sub_path_in. Qed.
```
chdoc, line 379: I'm a bit surprised/sad that the lemmas in this section are proved after their more specific counterparts. On the other hand, all lemmas in the `HomoPath` section are already one-liners. So maybe we can keep it as is.

chdoc (commented):
This already looks very solid to me. However, I wonder whether using `subseq` rather than explicit masks would simplify things. Put differently, I wonder whether there is additional  complexity introduced by not requiring an `eqType`, and whether this is is justified by some application or model.

mathcomp/ssreflect/path.v
```diff
@@ -517,25 +715,22 @@ elim: s [::] => /= [|x s ihs] ss allss.
...
 Qed.
 
-Lemma order_path_min x s : transitive leT -> path leT x s -> all (leT x) s.
+Lemma order_path_min x s : transitive leT -> path x s -> all (leT x) s.
 Proof.
 move=> leT_tr; elim: s => //= y [//|z s] ihs /andP[xy yz]; rewrite xy {}ihs//.
 by move: yz => /= /andP [/(leT_tr _ _ _ xy) ->].
 Qed.
 
 Hypothesis leT_tr : transitive leT.
```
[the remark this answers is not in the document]
pi8027: Unfortunately, it breaks `abelian.v`. https://github.com/math-comp/math-comp/pull/358#discussion_r295259267

mathcomp/ssreflect/path.v
```diff
@@ -137,55 +137,284 @@ Qed.
...
+Qed.
+
+Lemma path_filter_in x a s :
+  all P (x :: s) -> path leT x s -> path leT x (filter a s).
+Proof. by move=> Pxs; rewrite filter_mask; exact: path_mask_in. Qed.
+
+Lemma cycle_mask_in m s : all P s -> cycle leT s -> cycle leT (mask m s).
+Proof.
+case: (resize_mask m s) => {m} m sizeE ->.
+elim: m s sizeE => [|[]m ih] [|x s] //= [sizeE] /andP [Px Ps].
```
CohenCyril, line 230: I'd rather spend one more character to get better insight into the case analysis.
```suggestion
elim: m s sizeE => [|[|m] ih] [|x s] //= [sizeE] /andP [Px Ps].
```

mathcomp/ssreflect/path.v
```diff
@@ -137,55 +137,284 @@ Qed.
...
+Section Transitive_in.
+
+Variables (P : {pred T}) (leT : rel T).
+Hypothesis leT_tr : {in P & &, transitive leT}.
+
+Lemma path_mask_in x m s :
+  all P (x :: s) -> path leT x s -> path leT x (mask m s).
+Proof.
+elim: m s x => [|[] m ih] [|y s] x //=.
+  by case/and3P=> ? ? ? /andP [-> /ih ->] //; apply/andP.
```
CohenCyril, line 217: same here
```suggestion
elim: m s x => [|[|m] ih] [|y s] x //=.
  by move=> /and3P[? ? ?] /andP[-> /ih ->] //; apply/andP.
```

mathcomp/ssreflect/path.v
```diff
@@ -137,55 +137,284 @@ Qed.
...
+  apply: leT_tr sx xy => //.
+  by elim: s y Ps Py => //= z s ih y /andP [Pz Ps] _; apply: ih.
+Qed.
+
+Lemma cycle_filter_in a s : all P s -> cycle leT s -> cycle leT (filter a s).
+Proof. move=> Ps; rewrite filter_mask; exact: cycle_mask_in. Qed.
+
+Lemma sorted_mask_in m s : all P s -> sorted leT s -> sorted leT (mask m s).
+Proof.
+elim: m s => [|[] m ih] [|x s] //= Pxs; first exact: path_mask_in.
```
CohenCyril, line 244:
```suggestion
elim: m s => [|[|m] ih] [|x s] //= Pxs; first exact: path_mask_in.
```

mathcomp/ssreflect/path.v
pi8027: Using general ones to prove specific ones does not help us much here, but we can place them in the opposite order (also in the same section). Let me do so.

mathcomp/ssreflect/seq.v
```diff
@@ -1930,6 +1932,11 @@ elim: m s => [|b m IHm] [|x s] //; rewrite has_mask_cons /= andbC.
 by case: (a x) => //= /IHm.
 Qed.
 
+Lemma all_mask a m s : all a s -> all a (mask m s).
+Proof.
+by elim: m s => [|[] m IHm] [|x s] //= /andP [? /IHm ->]; rewrite ?andbT.
+Qed.
```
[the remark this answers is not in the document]
pi8027: @chdoc Can you open a new issue for this?

mathcomp/ssreflect/path.v
pi8027: `[]` is the case analysis for the first element of `m : bitseq` here. Your suggestion does not work.

CohenCyril: Ahah right!

CohenCyril: I got confused between the convention for natural numbers and masks.

CohenCyril, line 230: New suggestion then:
```suggestion
elim: m s sizeE => [|[] m ih] [|x s] //= [sizeE] /andP[Px Ps].
```

mathcomp/ssreflect/path.v
pi8027: Done.

pi8027:
> This already looks very solid to me. However, I wonder whether using `subseq` rather than explicit masks would simplify things. Put differently, I wonder whether there is additional complexity introduced by not requiring an `eqType`, and whether this is is justified by some application or model.

In general, I would like to make `_in` lemmas general enough to derive their specific counterparts immediately. Since `(homo|mono)_(path|cycle|sorted)` lemmas have not been derived from their `_in` counterparts, this probably does not apply to this PR. But, IMO, this is quite crucial to keep PR #601 as simple as possible, since many existing lemmas which I generalize do not require an `eqType`.

Also, I observed that `homo_sorted_in` generalized in this PR is useful to prove the following lemma (I actually managed to prove it):
```coq
Lemma sort_map_in
      (T : Type) (T' : eqType) (P : {pred T}) (f : T -> T') leT leT' :
  {in P &, antisymmetric (relpre f leT')} ->
  {in P & &, transitive (relpre f leT')} -> {in P &, total leT} ->
  {in P &, {homo f : x y / leT x y >-> leT' x y}} ->
  {in all P, forall s : seq T,
        sort leT' [seq f x | x <- s] = [seq f x | x <- sort leT s]}.
```

mathcomp/ssreflect/seq.v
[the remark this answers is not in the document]
pi8027: Oh, but proving `has_mask` and `all_mask` using the `subseq` lemmas specializes `T` to an `eqType`. I don't think it is a good idea.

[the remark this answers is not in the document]
chdoc: I'm of two minds about this. On the one hand, I agree that one should not require an `eqType` where none is required. On the other hand, I find it deeply unsatisfying to resort to induction just because `mathcomp` defines list membership (and hence list inclusion) only for `eqType`. Using the `In` predicate, whose use I'm really not advocating here, it's not too hard to prove:
```coq
Lemma In_mask x m s : In x (mask m s) -> In x s. 
Lemma In_all a s s' : (forall x, In x s -> In x s') -> all a s' -> all a s.

Lemma all_mask a m s : all a s -> all a (mask m s).
Proof. by apply/In_all => ?; exact: In_mask. Qed.
```
Again, this is not about this specific lemma, this is about the general principle about lists as "sets" or "collections" without equality and how to reason about them. @CohenCyril do you have an opinion on the matter?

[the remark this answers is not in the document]
pi8027: I think we should open a new issue anyway then.

mathcomp/ssreflect/path.v
```diff
@@ -137,55 +137,272 @@ Qed.
...
+Lemma cycle_mask_in m s : all P s -> cycle leT s -> cycle leT (mask m s).
+Proof.
+case: (resize_mask m s) => {m} m sizeE ->.
+elim: m s sizeE => [|[] m ih] [|x s] //= [sizeE] /andP[Px Ps].
+- rewrite -!cats1 -(mask_cat [:: true] [:: x]) //.
+  by apply: path_mask_in; rewrite /= all_cat /= Px Ps.
+- move=> xsx; apply: ih => // {sizeE}; case: s xsx Ps => //= y s.
+  rewrite !rcons_path => /and3P [xy -> sx] /andP [Py Ps] /=.
+  apply: leT_tr sx xy => //.
+  by elim: s y Ps Py => //= z s ih y /andP [Pz Ps] _; apply: ih.
```
chdoc, line 236: I'm afraid, but the lemma you removed from this PR was not the only place where the lack of an `eqType` structure makes you resort to induction where one would expect monotonicity to be used. For instance, the last line could be:
```
by apply/(@allP _ _ (y::s)); rewrite /= ?mem_last ?Py.
```
if `T` was an `eqType`. 

@CohenCyril , what's your position on this?

mathcomp/ssreflect/seq.v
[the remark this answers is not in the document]
CohenCyril: Until mathcomp introduces support for non decidable membership*, it's perfectly ok not to rely on a lemma involving `mem` to show something that does not involve it. If it really matters is is suggested one goes through `mkseqP` or `mkseq_nth` lemmas to index the biggest list with natural numbers, and hence reduce the reasoning to the eqType `nat`, (which is the best we can do when we are not able to distinguish one element from another).

*supporting non decidable predicate would be a radical modification of mathcomp and could introduce a lot of code duplication if not done properly. But on the other hand it would be extremely useful for mathcomp-analysis, and it could be a good topic of discussion for a mathcomp meeting or rather a working group on the topic.

mathcomp/ssreflect/path.v
CohenCyril: It's ok to "duplicate" short proofs.

CohenCyril: However, maybe this proof could start with `have [{}s] := mkseqP s`.

pi8027: I realized the fact `leT` relates all the elements of `s` if `leT` is transitive and `cycle leT s` holds. So `cycle_(filter|mask)_in` lemmas may not be that useful. Shouldn't we rather remove them for now?

mathcomp/ssreflect/seq.v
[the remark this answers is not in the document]
CohenCyril: BTW you could reuse old material with the following proof, but that would be counterproductive...
```coq
Lemma all_mask T (a : pred T) m s : all a s -> all a (mask m s).
Proof.
case: s => [|x s]; first by rewrite mask0.
have [n f {x s}_] := mkseqP x (x :: s); rewrite -map_mask !all_map.
by move=> /allP fan; apply/allP => x /mem_mask/fan.
Qed.
```

mathcomp/ssreflect/path.v
CohenCyril: Sure, maybe there should be a proof that `cycle leT s -> allrel leT s` ?

pi8027: Good. Let me do that in another PR since #630 seems to be useful for it.

mathcomp/ssreflect/seq.v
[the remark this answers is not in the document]
chdoc: I opened issue #638 to organize the larger discussion on non-decidable membership and I will ignore the issue as it comes to this PR.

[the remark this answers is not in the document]
pi8027: @chdoc Thanks!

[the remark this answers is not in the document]
pi8027: Ah, I discovered another one-liner!
```suggestion
Lemma all_mask a m s : all a s -> all a (mask m s).
Proof. by apply/contraTT; rewrite -!has_predC; apply: has_mask. Qed.
```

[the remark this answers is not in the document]
pi8027: I decided to put `all_mask` before `has_mask` and shorten `has_mask` similarly.

mathcomp/ssreflect/seq.v
```diff
@@ -1920,15 +1922,17 @@ Lemma mask_cat m1 m2 s1 s2 :
   size m1 = size s1 -> mask (m1 ++ m2) (s1 ++ s2) = mask m1 s1 ++ mask m2 s2.
 Proof. by move: m1 s1; apply: seq_ind2 => // -[] m1 x1 s1 /= _ ->. Qed.
 
+Lemma all_mask a m s : all a s -> all a (mask m s).
+Proof.
+by elim: m s => [|[] m IHm] [|x s] //= /andP [? /IHm ->]; rewrite ?andbT.
+Qed.
```
CohenCyril, line 1928:
```suggestion
Proof. by elim: s m => [|x s IHs] [|[] m]//= /andP[ax /IHs->]; rewrite ?ax. Qed.
```

CohenCyril: (the induction on `s` changes nothing but I have my own tastes :wink:

CohenCyril (approved):
LGTM!

chdoc (approved):
LGTM, too!

PR #646, opened 2020-11-12
Rename `eq_sorted` lemmas to `sorted_eq` and generalize `sort_le_id`

pi8027:

Part of #601.

- This PR renames `eq_sorted(|_irr|_le|_lt)` lemmas to `(|irr_|le_|lt_)sorted_eq` to address the following naming inconsistency.
  ```coq
  eq_path
       : forall (T : Type) (e e' : rel T), e =2 e' -> path e =2 path e'

  eq_cycle
       : forall (T : Type) (e e' : rel T), e =2 e' -> cycle e =1 cycle e'

  eq_sorted
       : forall (T : eqType) (leT : rel T),
         transitive leT ->
         antisymmetric leT ->
         forall s1 s2 : seq T,
         sorted leT s1 -> sorted leT s2 -> perm_eq s1 s2 -> s1 = s2
  ```
  The plan is to deprecate the current `eq_sorted` lemma in 1.12, and replace it with the following lemma in 1.13:
  ```coq
  eq_sorted
       : forall (T : Type) (e e' : rel T), e =2 e' -> sorted e =1 sorted e'
  ```
- Lemma `sort_le_id` has been generalized from `orderType` to `porderType`.

##### Things done/to do

- [ ] after merge: open an issue to deal with the deprecations in 1.13

### DIFF
--- a/mathcomp/ssreflect/order.v
+++ b/mathcomp/ssreflect/order.v
@@ -2741,7 +2741,7 @@ Proof. by rewrite andbC lt_le_asym. Qed.
 
 Definition lte_anti := (=^~ eq_le, lt_asym, lt_le_asym, le_lt_asym).
 
-Lemma lt_sorted_uniq_le s : sorted lt s = uniq s && sorted le s.
+Lemma lt_sorted_uniq_le s : sorted <%O s = uniq s && sorted <=%O s.
 Proof.
 case: s => //= n s; elim: s n => //= m s IHs n.
 rewrite inE lt_neqAle negb_or IHs -!andbA.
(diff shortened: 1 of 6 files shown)


mathcomp/ssreflect/order.v
```diff
@@ -2741,7 +2741,7 @@ Proof. by rewrite andbC lt_le_asym. Qed.
 
 Definition lte_anti := (=^~ eq_le, lt_asym, lt_le_asym, le_lt_asym).
 
-Lemma lt_sorted_uniq_le s : sorted lt s = uniq s && sorted le s.
+Lemma lt_sorted_uniq_le s : sorted <%O s = uniq s && sorted <=%O s.
```
CohenCyril, line 2744: @pi8027 this suggest an even more consistent naming for the former `eq_sorted_lt` would be `lt_sorted_eq`.

CohenCyril: This would mean the new names would be `le_sorted_eq`, `lt_sorted_eq` and `irr_sorted_eq`.

pi8027: That makes sense!

pi8027: Done.

CohenCyril (approved):
LGTM

chdoc (approved):
LGTM!

PR #649, opened 2020-11-13
Add sorting lemmas specialized for `porderType` and `orderType`

### LINKED ISSUE #622: Missing sorting lemmas specialized for `orderType`
> > Since `gtn n m` is convertible to `m < n`, we can rather use `ltn_trans` and `ltnn`. But if there are any use cases that `ltn_trans` and `ltnn` are not applicable, I would like to see.
> 
> I came across it when sorting in descending order and afterwards using the lemmas about `sort` (an example is `sorted_uniq`) which often leaves you with a goal `transitive gtn`. I know that `transitive gtn` can easily be derived from `ltn_trans`, but I think that this should be closed with just a single `apply` (or ideally just providing the argument directly to `sorted_uniq`), which I don't think it can without introductions using only `ltn_trans`.

I see. Let me suggest another solution. Some sorting related lemmas specialized to `orderType` are also available in the order library. You may use them with dual ordering, e.g.,
```coq
Import Order.Theory.

Check (@sort_le_sorted _ [orderType of nat^d]).
(*
sort_le_sorted
     : forall s : seq [orderType of nat^d], sorted <=^d%O (sort <=^d%O s)
*)
Check (fun s : seq nat^d => sort_le_sorted s).
(*
[eta sort_le_sorted]
     : forall s : seq nat^d, sorted <=^d%O (sort <=^d%O s)
*)
```
where `<=^d%O` means `geq` here. Unfortunately, many lemmas including `sorted_uniq` are missing in order.v, but we can add them anyway.

_Originally posted by @pi8027 in https://github.com/math-comp/math-comp/issues/505#issuecomment-657971174_

pi8027:

Closes #622

`sorted_(ltn|leq)_(nth|index)` lemmas specialized for `porderType` are still missing on purpose. https://github.com/math-comp/math-comp/pull/601#discussion_r523141733

##### Things done/to do

### DIFF
--- a/mathcomp/ssreflect/order.v
+++ b/mathcomp/ssreflect/order.v
@@ -2759,21 +2759,59 @@ Proof. by rewrite andbC lt_le_asym. Qed.
 
 Definition lte_anti := (=^~ eq_le, lt_asym, lt_le_asym, le_lt_asym).
 
+Lemma le_path_min x s : path <=%O x s -> all (>= x) s.
+Proof. exact/order_path_min/le_trans. Qed.
+
+Lemma lt_path_min x s : path <%O x s -> all (> x) s.
+Proof. exact/order_path_min/lt_trans. Qed.
+
+Lemma le_path_sortedE x s : path <=%O x s = all (>= x) s && sorted <=%O s.
+Proof. exact/path_sortedE/le_trans. Qed.
+
+Lemma lt_path_sortedE x s : path <%O x s = all (> x) s && sorted <%O s.
+Proof. exact/path_sortedE/lt_trans. Qed.
+
+Lemma le_sorted_pairwise s : sorted <=%O s = pairwise <=%O s.
+Proof. exact/sorted_pairwise/le_trans. Qed.
+
+Lemma lt_sorted_pairwise s : sorted <%O s = pairwise <%O s.
+Proof. exact/sorted_pairwise/lt_trans. Qed.
+
+Lemma le_path_pairwise x s : path <=%O x s = pairwise <=%O (x :: s).
+Proof. exact/path_pairwise/le_trans. Qed.
+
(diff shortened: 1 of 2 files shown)


mathcomp/ssreflect/order.v
```diff
@@ -3741,6 +3795,30 @@ Hint Resolve sort_le_sorted : core.
...
+
+Lemma subseq_sort_le : {homo sort <=%O : s1 s2 / @subseq T s1 s2}.
+Proof. exact/subseq_sort/le_trans/le_total. Qed.
+
+Lemma sorted_subseq_sort_le s1 s2 :
+  subseq s1 s2 -> sorted <=%O s1 -> subseq s1 (sort <=%O s2).
+Proof. exact/sorted_subseq_sort/le_trans/le_total. Qed.
+
+Lemma mem2_sort_le s x y : x <= y -> mem2 s x y -> mem2 (sort <=%O s) x y.
+Proof. exact/mem2_sort/le_trans/le_total. Qed.
```
pi8027, line 3820: Since `<=%O` is antisymmetric, these stability lemmas may not be that useful. I have no idea whether they are needed or not.

pi8027:
Now I think it makes a lot of sense to wait for a major update of real-closed to see what we need here.

pi8027:
I will resume this PR after #727 and #738 get merged.

pi8027:
> `sorted_(ltn|leq)_(nth|index)` lemmas specialized for `porderType` are still missing on purpose. [#601 (comment)](https://github.com/math-comp/math-comp/pull/601#discussion_r523141733)

`(lt|le)_sorted_(ltn|leq)_nth` are actually introduced in #738. We should probably double-check them.

CohenCyril (approved):
LGTM

PR #653, opened 2020-11-19
[WIP] Make linear subspaces canonically a `tbLatticeType`

### LINKED ISSUE #613: Modular lattice of linear subspaces
The subspaces `{vspace vT}` of a vector space `vT` (defined in `vector.v`) seem to form a lattice. [A Wikipedia article](https://en.wikipedia.org/wiki/Linear_subspace) says that "the operations intersection and sum make the set of all (linear) subspaces a bounded modular lattice". So we may apply the lattice theory to reason about vector subspaces by defining this lattice structure instance. Also, the modular lattice structures are intermediate ones between `(tb)LatticeType` and `(tb)DistrLatticeType`. So it would be nice to have them in `order.v`. Does this make sense? Is there any good literature on this topic?

pi8027:

This is my first attempt to make linear subspaces `{vspace vT}` canonically a `tbLatticeType`. In `vector.v`, some operators for linear subspace (`subsetv`, `fullv`, `addv`, and `capv`) have been redefined as notations for order operators. *The first obstacle is that `falgebra.v` does not compile since both `prodv` and `capv` (`Order.meet`) distributes over `addv` (`Order.join`) and so `addv_addoid` (in `falgebra.v`) is now redundant with `join_addoid` (in `order.v`).*

Closes #613

##### Things done/to do

### DIFF
--- a/algebra/vector.v
+++ b/algebra/vector.v
@@ -2,7 +2,7 @@
 (* Distributed under the terms of CeCILL-B.                                  *)
 From HB Require Import structures.
 From mathcomp Require Import ssreflect ssrfun ssrbool eqtype ssrnat seq choice.
-From mathcomp Require Import fintype tuple finfun bigop nmodule.
+From mathcomp Require Import fintype tuple finfun bigop nmodule order.
 From mathcomp Require Import rings_modules_and_algebras divalg matrix mxalgebra.
 
 (******************************************************************************)
@@ -129,7 +129,9 @@ Reserved Notation "\dim A" (at level 10, A at level 8, format "\dim  A").
 
 Delimit Scope vspace_scope with VS.
 
-Import GRing.Theory.
+Import Order.LTheory GRing.Theory.
+
+Fact vspace_display : Order.disp_t. Proof. exact: Order.Disp tt tt. Qed.
 
 (* Finite dimension vector space *)
 Record semivector_axiom_def (R : nzSemiRingType) n (V : lSemiModType R) := {
@@ -295,7 +297,6 @@ Implicit Types (u : vT) (X : seq vT) (U V : {vspace vT}).
 HB.instance Definition _ := [Choice of {vspace vT} by <:].
 
 Definition dimv U := \rank (vs2mx U).
-Definition subsetv U V := (vs2mx U <= vs2mx V)%MS.
 Definition vline u := mx2vs (v2r u).
 
 (* Vspace membership is defined as line inclusion. *)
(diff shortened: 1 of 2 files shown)


mathcomp/algebra/vector.v
```diff
@@ -2045,6 +2041,239 @@ Qed.
...
+
+Lemma addvC : commutative addV. Proof. exact: joinC. Qed.
+
+Lemma addvA : associative addV. Proof. exact: joinA. Qed.
+
+Lemma addv_idPl {U V}: reflect (U + V = U)%VS (V <= U)%VS.
+Proof. exact: join_idPr. Qed.
+
+Lemma addv_idPr {U V} : reflect (U + V = V)%VS (U <= V)%VS.
+Proof. exact: join_idPl. Qed.
```
pi8027, line 2099: It seems like the names `join_idP(l|r)` are transposed. Should be fixed in 1.12. @CohenCyril https://github.com/math-comp/math-comp/blob/0606b6bf22e258dc3b7cf440f10c108f785904b5/mathcomp/ssreflect/order.v#L3642-L3645

mathcomp/algebra/vector.v
```diff
@@ -596,50 +662,8 @@ End BigSum.
...
-
-Lemma capvf : right_id fullv capV.
-Proof. by move=> U; apply/capv_idPl/subvf. Qed.
-
-Canonical capv_monoid := Monoid.Law capvA capfv capvf.
-Canonical capv_comoid := Monoid.ComLaw capvC.
-
 Lemma memv_cap w U V : (w \in U :&: V)%VS = (w \in U) && (w \in V).
-Proof. by rewrite !memvE subv_cap. Qed.
+Proof. by rewrite !memvE lexI. Qed.
```
pi8027, line 666: If `_ \in _` for linear subspaces is defined by using `Order.le` as in intervals, this proof can probably be `exact: lexI`. (I think it requires some more work.) https://github.com/math-comp/math-comp/blob/0606b6bf22e258dc3b7cf440f10c108f785904b5/mathcomp/algebra/interval.v#L493-L494

mathcomp/algebra/vector.v
CohenCyril: Well spotted!

pi8027: See #671

pi8027:
> *The first obstacle is that `falgebra.v` does not compile since both `prodv` and `capv` (`Order.meet`) distributes over `addv` (`Order.join`) and so `addv_addoid` (in `falgebra.v`) is now redundant with `join_addoid` (in `order.v`).*

Redefining `Monoid.add_law` as follows would probably solve this issue. But, if there is a multiplication-like operator that distributes over two different addition-like operators, it does not work. Is there any plan to extend the `bigop` hierarchy? https://github.com/math-comp/math-comp/blob/ee410eb85f36f5e366f807207a7dee8f3e1bdf3a/mathcomp/ssreflect/bigop.v#L349-L354
```coq
(* We have to swap its parameter (multiplication) and the first projection    *)
(* (addition) so that its canonical instances can be indexed by constants of  *)
(* multiplication rather than addition.                                       *)
Structure mul_distr_law (add : T -> T -> T) := MulDistrLaw {
  mul_distr_operator : T -> T -> T;
  (* The following two axioms are part of `mul_law`. A telescope cannot be used here anymore. *)
  _ : left_zero idm mul_distr_operator;
  _ : right_zero idm mul_distr_operator;

  _ : left_distributive mul_distr_operator add;
  _ : right_distributive mul_distr_operator add
}.
```

pi8027:
> But, if there is a multiplication-like operator that distributes over two different addition-like operators, it does not work.

I self answer this: `muln` distributes over `maxn` and `addn`, and so it does not work.

PR #660, opened 2020-11-20
Using Arguments simpl never instead of nosimpl

### LINKED ISSUE #94: Use Arguments simpl never functions
(EDIT by Cyril)
The following definitions are defined using `nosimpl`, they could be defined using `Arguments` option `simpl never` instead

CohenCyril:

Fixes #94 

##### Things done/to do

### DIFF
--- a/mathcomp/character/classfun.v
+++ b/mathcomp/character/classfun.v
@@ -443,14 +443,13 @@ Definition isometry_from_to mCFD tau mCFR :=
    prop_in2 mCFD (inPhantom (isometry tau))
   /\ prop_in1 mCFD (inPhantom (forall phi, in_mem (tau phi) mCFR)).
 
-End Predicates.
+Definition orthogonal S1 S2 := @ortho_rec S1 S2.
 
-(* Outside section so the nosimpl does not get "cooked" out. *)
-Definition orthogonal gT D S1 S2 := nosimpl (@ortho_rec gT D S1 S2).
+End Predicates.
 
 Arguments cfker {gT D%g} phi%CF.
 Arguments cfaithful {gT D%g} phi%CF.
-Arguments orthogonal {gT D%g} S1%CF S2%CF.
+Arguments orthogonal {gT D%g} S1%CF S2%CF : simpl never.
 Arguments pairwise_orthogonal {gT D%g} S%CF.
 Arguments orthonormal {gT D%g} S%CF.
 Arguments isometry {gT rT D%g R%g} tau%CF.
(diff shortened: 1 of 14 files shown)


CohenCyril:
CC @gares note that `simpl never` does not have the same behavior as `nosimpl` when adding a definition on top of a `simpl never` one, we might go through the unfolding. (Hence the need to fix `pfactor` again in `prime.v`)

mathcomp/character/classfun.v
```diff
@@ -445,12 +445,11 @@ Definition isometry_from_to mCFD tau mCFR :=
 
 End Predicates.
 
-(* Outside section so the nosimpl does not get "cooked" out. *)
-Definition orthogonal gT D S1 S2 := nosimpl (@ortho_rec gT D S1 S2).
+Definition orthogonal gT D S1 S2 := @ortho_rec gT D S1 S2.
```
pi8027, line 448: I think this one should be relocated to the above `Predicates` section.

gares (approved):

ybertot:
I am workiing on a patch to fourcolor so that it is compatible with this change.  There are good chances that the patch will be backward compatible.

CohenCyril:
I will investigate `odd_order`

CohenCyril:
@gares, the problem we encountered is the one I mentioned for e.g. `pfactor`, except it occurs for a local definition.
This seems to be an essential shortcoming of `simpl never`...

CohenCyril:
@ybertot @gares, Unfortunately, I think the consequence of this exploration is that this PR cannot be merged because it introduces a regression in the behaviour of symbols that are no simplifiable... and that a bugfix should be submitted on the Coq side (which I am going to do).

ybertot:
I came to a similar conclusion.

gares:
> a bug report should be submitted on the Coq side (which I am going to do).

please do, since I can't exactly understand what the problem is from the comments here

ybertot:
I have been trying to construct a minimal example, but I can't reproduce the problem in a few lines yet.

ybertot:
In the meantime shouldn't this PR be re-affected to milestone 1.13.0?

CohenCyril:
> I have been trying to construct a minimal example, but I can't reproduce the problem in a few lines yet.

I did it here https://github.com/coq/coq/issues/13428

CohenCyril:
> In the meantime shouldn't this PR be re-affected to milestone 1.13.0?

I am afraid way later than 1.13 since, assuming someone fixes this for Coq 8.13, we would have to wait until Coq 8.15 is out.
But we must test the feature ASAP this time :cry:

PR #661, opened 2020-11-20
Tuning simplifications using Arguments nomatch

### LINKED ISSUE #95: Use Arguments to tune simplification of functions like nth, subn
Many functions consume two inductive arguments in parallel.
`Arguments` lets one block simplification when only one of the two is concrete, improving usability.

CohenCyril:

Fixes #95 

In particular this solves the notorious problem of ``p`_0`` expanding too eagerly.

##### Things done/to do

### DIFF
--- a/CHANGELOG_UNRELEASED.md
+++ b/CHANGELOG_UNRELEASED.md
@@ -294,6 +294,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - in `path.v`, added `size_merge_sort_push`, which documents an
   invariant of `merge_sort_push`.
 
+- in `seq.v ` added lemma `mask0s`.
+
 ### Changed
 
 - in `ssrbool.v`, use `Reserved Notation` for `[rel _ _ : _ | _]` to avoid warnings with coq-8.12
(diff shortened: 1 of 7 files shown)


gares (approved):

gares:
Can you tell me why ! was not ok?

CohenCyril:
> Can you tell me why ! was not ok?

Almost all definitions (except for `eqseq`) which are involved in this tuning have at least one "default" case, where one argument does not need to be a constructor for a ground value to be output. e.g. in `take` if the list is `[::]`, then whatever the natural number, the output is `[::]`. The "right" criterion is that no match should be exposed by doing so.

Except for `mask` where there is an extra match on a boolean that we would like to expose. So I sacrificed expansion when the mask is `[::]` and not the list `s`, for the sake of still having a form of controlled expansion, which would still expose the match on the boolean.

This hints at the fact that both `!` and `nomatch` are too restrictive in different fashion. Somehow, we would like to say which `match` are allowed/forbidden instead. e.g. `Arguments mask _ !m !s : simpl nomatch`, will simplify iff it does not (directly) expose a match on `m` or `s`. Moreover `nomatch` will prevent expension even if the appearance of the match has nothing to do with the unfolding of the constant itself. e.g.
```coq
Definition mask_if mask b (x : T) m' s' :=  if b then x :: mask m' s' else mask m' s'
Fixpoint mask m s {struct m} :=
  match m, s with
  | b :: m', x :: s' => mask_if mask b x m' s'
  | _, _ => [::]
  end.
Arguments mask !m !s.
Arguments mask_if /.
```
will still no simplify `mask (b :: m) (x :: s)` because doing one loop of the fixpoint and then expanding `mask_if` will result in a match :cry: :cry: :cry:

CohenCyril:
@affeldt-aist the CI passed, I rebased once more, when it passes again, you can merge.

gares:
I'm OK and aware that "if" is problematic with no match.

PR #665, opened 2020-11-20
Generalize `allrel` to take two lists as arguments

### LINKED ISSUE #662: A generalization of `allrel` that takes two lists
```coq
Definition allrel {T S : Type} (r : T -> S -> bool) xs ys :=
   all id [seq r x y | x <- xs, y <- ys].
```
Since `all (fun x => all (fun y => ...) ys) xs` appears at least twice, this generalization would be useful. https://github.com/math-comp/math-comp/blob/676a9266ad77232ab198c86a6a3a3f3f6ba53cc0/mathcomp/ssreflect/path.v#L976-L978 https://github.com/math-comp/math-comp/blob/676a9266ad77232ab198c86a6a3a3f3f6ba53cc0/mathcomp/character/classfun.v#L429-L430

Incidentally, the following ones can be reformulated using the current `allrel`. https://github.com/math-comp/math-comp/blob/676a9266ad77232ab198c86a6a3a3f3f6ba53cc0/mathcomp/field/falgebra.v#L1080-L1082 https://github.com/math-comp/math-comp/blob/676a9266ad77232ab198c86a6a3a3f3f6ba53cc0/mathcomp/field/separable.v#L291-L292

pi8027:

Closes #662. The unary case is renamed to `all1rel`.

##### Things done/to do


mathcomp/ssreflect/seq.v
```diff
@@ -3460,45 +3463,94 @@ Arguments perm_consP {T x s t}.
...
+
+Lemma all1rel_cons x xs :
+  all1rel r (x :: xs) = [&& r x x, all (r x) xs & all1rel r xs].
+Proof.
+rewrite allrel_consl allrel_consr /= !andbA [r x x && _]andbC -andbA andbACA.
+rewrite -[RHS]andbA; congr andb; rewrite -all_predI.
+by elim: xs => //= y xs ->; rewrite rsym andbb.
+Qed.
+
+End All1Rel.
```
pi8027, line 3553: I think most use-cases of `all1rel` require `r` to be symmetric. This is the case at least in `all_comm_mx`, `ahom_in`, and `Derivation`. But, if it is not the case in general, I would like to put the suffix `_sym` to the names of these lemmas.

mathcomp/ssreflect/seq.v
```diff
@@ -132,8 +132,11 @@ From mathcomp Require Import ssreflect ssrfun ssrbool eqtype ssrnat.
 (*                    i.e. self expanding definition for                      *)
 (*                      [seq f x y | x <- s, y <- t]                          *)
 (*               := [:: f x_1 y_1; ...; f x_1 y_m; f x_2 y_1; ...; f x_n y_m] *)
-(*     allrel r s := all id [seq r x y | x <- xs, y <- xs]                    *)
-(*                == the proposition r x y holds for all possible x, y in xs  *)
+(* allrel r xs ys := all [pred x | all (r x) ys] xs                           *)
+(*                == the proposition r x y holds for all x and y respectively *)
+(*                   in xs and ys.                                            *)
+(*   all1rel r xs := allrel r xs xs                                           *)
```
CohenCyril, line 138: I prefer `all2rel` or `pairwise` than `all1rel`.
@gares, @chdoc, you participated in the first discussion on `allrel`, what do you say about this change and my suggestion?

mathcomp/ssreflect/path.v
```diff
@@ -963,24 +963,25 @@ Hypothesis (leT_total : total leT) (leT'_tr : transitive leT').
...
 elim: s1 s2 x => //= x s1 ih1; elim => //= y s2 ih2 h.
-rewrite all_predI -andbA => /and4P [xy' xs2 ys1 s1s2].
-case/andP => hx xs1 /andP [] hy ys2; case: ifP => xy /=; rewrite (hx, hy) /=.
-- by apply: ih1; rewrite ?all_predI ?ys1 //= xy xy' implybT.
-- by apply: ih2; have:= leT_total x y; rewrite ?xs2 //= xy => /= ->.
+rewrite allrel_consl allrel_consr /= -andbA.
+move=> /and4P [xy' xs2 ys1 s1s2] /andP [hx xs1] /andP [hy ys2].
+case: ifP => xy /=; rewrite (hx, hy) /=.
+- by apply: ih1; rewrite ?allrel_consr ?ys1 //= xy xy' implybT.
+- by apply: ih2; have:= leT_total x y; rewrite ?allrel_consl ?xs2 ?xy //= => ->.
```
CohenCyril, line 974: I'm a bit concerned that the proof looks longer...

mathcomp/ssreflect/seq.v
```diff
@@ -3460,45 +3463,94 @@ Arguments perm_consP {T x s t}.
...
+Implicit Types (x : T) (y : S) (xs : seq T) (ys : seq S).
 
-Lemma allrel0 (T : Type) (r : rel T) : allrel r [::].
-Proof. by []. Qed.
+Definition allrel xs ys := all [pred x | all (r x) ys] xs.
 
-Lemma allrel_map (T T' : Type) (f : T' -> T) (r : rel T) xs :
-  allrel r (map f xs) = allrel (relpre f r) xs.
-Proof. by rewrite /allrel allpairs_mapl allpairs_mapr. Qed.
+Lemma allrel0s ys : allrel [::] ys. Proof. by []. Qed.
```
CohenCyril, line 3471: I had the impression that the convention here was `l` `r` rather than positional (as in `allpairs0l`)

mathcomp/ssreflect/seq.v
CohenCyril: These lemma only make sense when the relation is symmetric anyway, otherwise it can all be obtained by `allrel_cons2` which would be the lemma combining `allrel_consr` and `allrel_consl`.

mathcomp/ssreflect/path.v
CohenCyril, line 974:
```suggestion
rewrite allrel_cons2 => /and4P [xy' xs2 ys1 s1s2] /andP [hx xs1] /andP [hy ys2].
case: ifP => xy /=; rewrite (hx, hy) /=.
- by apply: ih1; rewrite ?allrel_consr ?ys1 //= xy xy' implybT.
- by apply: ih2; have:= leT_total x y; rewrite ?allrel_consl ?xs2 ?xy //= => ->.
```

mathcomp/ssreflect/seq.v
CohenCyril: PS: and maybe that's what shortens the proof of `merge_stable_path` once more, as I tried to exhibit in [another comment](https://github.com/math-comp/math-comp/pull/665/files#r527591337)

mathcomp/ssreflect/seq.v
```diff
@@ -3460,45 +3463,94 @@ Arguments perm_consP {T x s t}.
...
-Variable (T : nonPropType) (r : rel T).
-Implicit Types (xs : seq T) (x y z : T).
-Hypothesis (rxx : reflexive r) (rsym : symmetric r).
+Lemma allrel_consl x xs ys : allrel (x :: xs) ys = all (r x) ys && allrel xs ys.
+Proof. by []. Qed.
+
+Lemma allrel_consr xs y ys :
+  allrel xs (y :: ys) = all (r^~ y) xs && allrel xs ys.
+Proof. exact: all_predI. Qed.
+
```
CohenCyril, line 3481:
```suggestion

Lemma allrel_cons2 x y xs ys :
￼  allrel (x :: xs) (y :: ys) = [&& r x y, all (r x) ys, all (r^~ y) xs & allrel xs ys].
￼Proof. by rewrite /= all_predI/= -andbA. Qed. (* or something like that *)

```

pi8027: I think we should rather provide a multi-rule rewriting lemma consists of `allrel_(cons|cat)(l|r)`, etc.

mathcomp/ssreflect/seq.v
```diff
@@ -132,8 +132,11 @@ From mathcomp Require Import ssreflect ssrfun ssrbool eqtype ssrnat.
 (*                    i.e. self expanding definition for                      *)
 (*                      [seq f x y | x <- s, y <- t]                          *)
 (*               := [:: f x_1 y_1; ...; f x_1 y_m; f x_2 y_1; ...; f x_n y_m] *)
-(*     allrel r s := all id [seq r x y | x <- xs, y <- xs]                    *)
-(*                == the proposition r x y holds for all possible x, y in xs  *)
+(* allrel r xs ys := all [pred x | all (r x) ys] xs                           *)
+(*                == the proposition r x y holds for all x and y respectively *)
+(*                   in xs and ys.                                            *)
```
chdoc, line 137: How about shortening the description to something along the lines of:
```suggestion
(* allrel r xs ys := all [pred x | all (r x) ys] xs                           *)
(*                == r x y holds whenever x is in xs and y is in ys           *)
```
I don't feel the need to use `proposition` here, as it it actually a boolean.

mathcomp/ssreflect/seq.v
chdoc: I'm in favor of the new definition, I wasn't fond of the `all id ...` to begin with. As to to the naming, don't have much to say.

CohenCyril: On second thoughts, I prefer to reserve `pairwise` for
```coq
Fixpoint pairwise {T} (r : rel T) (xs : seq T) :=
  if xs is x :: xs then all (r x) xs && pairwise r xs else true.
```
That might come up in mathcomp 1.13 and that will be equivalent to `all2rel` under `reflexive r` and `symmetric r` only...

mathcomp/ssreflect/seq.v
CohenCyril: We could have both, but the multirule rule may not lead to the same results modulo simpl :confused: so I'd rather let the user form them themselves at their own risk.

mathcomp/ssreflect/seq.v
CohenCyril: What about `selfrel` ?

mathcomp/ssreflect/seq.v
```diff
@@ -3460,45 +3463,94 @@ Arguments perm_consP {T x s t}.
...
+Arguments allrel_consl {T S} r x xs ys.
+Arguments allrel_consr {T S} r xs y ys.
+Arguments allrel1s {T S} r x ys.
+Arguments allrels1 {T S} r xs y.
+Arguments allrel_catl {T S} r xs xs' ys.
+Arguments allrel_catr {T S} r xs ys ys'.
+Arguments allrel_allpairsE {T S} r xs ys.
+
+Notation all1rel r xs := (allrel r xs xs).
+
```
CohenCyril, line 3516: Missing 
```coq
Lemma eq_allrel {T S : Type} (r r': T -> S -> bool) xs ys :
  r =2 r' -> allrel r xs ys = allrel r' xs ys.
Proof. (*...*) Qed.

Lemma eq_in_allrel {T S : Type} (P : {pred T}) (Q : {pred S}) (r r': T -> S -> bool) xs ys :
  all P xs -> all Q ys -> {in P & Q, r =2 r'} -> allrel r xs ys = allrel r' xs ys.
Proof. (*...*) Qed.
```

mathcomp/ssreflect/seq.v
pi8027: @CohenCyril I thought your initial suggestion was to name the general version `all2rel` (https://github.com/math-comp/math-comp/issues/662#issuecomment-730807158). This one is fine with me. But, since I preferred to give a shorter name for a general version, I renamed the original `allrel` to `all1rel` ("1" stands for unary in the sense of that `all1rel` takes one list). Also, I'm not fond of `selfrel`...

CohenCyril: I prefer to keep `allrel` for the general version. But I was considering renaming `all1rel` to `all2rel` (expressing all 2 are related..., while I cannot read `all1rel` in english, besides interpreting `1` as the arity...).
However, it does not matter much to me, we can keep `all1rel`, at least for 1.12.

CohenCyril (changes requested):
LGTM modulo changelog entry!

mathcomp/ssreflect/seq.v
pi8027: > I prefer to keep `allrel` for the general version. But I was considering renaming `all1rel` to `all2rel` (expressing all 2 are related..., while I cannot read `all1rel` in english, besides interpreting `1` as the arity...).

Sure. Now I think it is fine to rename `all1rel` to `all2rel`.

CHANGELOG_UNRELEASED.md
```diff
@@ -88,11 +88,16 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
...
-  + lemmas `allrel1`, `allrel2` and `allrel_cons`,
-    under assumptions of reflexivity and symmetry of `r`.
+- in `seq.v`, new high-order relation `allrel r xs ys` which asserts that
+  `r x y` holds whenever `x` is in `xs` and `y` is in `ys`, new notation
+  `all2rel r xs (:= allrel r xs xs)` which asserts that `r` holds on all
+  pairs of elements of `xs`, and
+  + lemmas `allrel0(l|r)`, `allrel_cons(l|r|2)`, `allrel1(l|r)`,
+    `allrel_cat(l|r)`, `allrel_allpairsE`, `eq_in_allrel`, `eq_allrel`,
+    `allrelC`, `allrel_map(l|r)`, `allrelP`,
+  + lemmas `all1rel1`, `all1rel2`, and `all1rel_cons`
```
CohenCyril, line 98:
```suggestion
  + lemmas `all2rel1`, `all2rel2`, and `all2rel_cons`
```

mathcomp/ssreflect/seq.v
pi8027: BTW, `pairwise` is also equivalent to `sorted` under transitivity. Reformulating some lemmas about `sorted` using `pairwise` may allow us to drop the transitivity assumption.

CohenCyril: That would indeed be a nice thing to do... for mathcomp 1.13.0 :wink:

CohenCyril (approved):
Ready to merge, modulo rebase.

PR #670, opened 2020-11-21
Fix deprecation notations in path.v

### LINKED ISSUE #668: wrong deprecation notations in path.v
The recently merged PR #650 introduces various deprecation notations. At least one of them causes compatibility issues:

```coq
Notation "@ 'sorted_le_nth'" := (deprecate sorted_le_nth sorted_leq_nth)
  (at level 10, only parsing) : fun_scope.
Notation sorted_le_nth :=
  (fun leT_tr leT_refl x0 s_sorted =>
     @sorted_le_nth _ _ leT_tr leT_refl x0 _ s_sorted _ _) (only parsing).
```
Using  `@sorted_le_nth` (rather than `@sorted_leq_nth`) in the second notation completely messes up the arguments, as can be seen below:
```coq
Lemma foo s (srt_s : sorted leq s) : 
  forall i j, i < j -> i < size s -> j < size s -> nth 0 s i <= nth 0 s j.
Proof.
move => i j /ltnW i_j i_s j_s.
apply: sorted_le_nth. 
(* compare with apply: sorted_leq_nth. *)
```

All of these notations should be carefully rechecked.

Ping! @pi8027 @CohenCyril

pi8027:

Fix #668.

~Since the first argument of `sorted_leq_nth` is maximally implicit, `@sorted_le_nth` does not simulate explicit application yet. We probably have to turn some of them into notations without deprecation as in #640 for now.~

##### Things done/to do

- ~[ ] added corresponding entries in `CHANGELOG_UNRELEASED.md` (do not edit former entries)~

### DIFF
--- a/mathcomp/ssreflect/path.v
+++ b/mathcomp/ssreflect/path.v
@@ -1446,13 +1446,21 @@ Notation "@ 'eq_sorted'" :=
   (deprecate eq_sorted sorted_eq) (at level 10, only parsing) : fun_scope.
 Notation "@ 'eq_sorted_irr'" := (deprecate eq_sorted_irr irr_sorted_eq)
   (at level 10, only parsing) : fun_scope.
-Notation "@ 'sorted_lt_nth'" := (deprecate sorted_lt_nth sorted_ltn_nth)
+Notation "@ 'sorted_lt_nth'" :=
+  (fun (T : Type) (leT : rel T) (leT_tr : transitive leT) =>
+     deprecate sorted_lt_nth sorted_ltn_nth leT_tr)
   (at level 10, only parsing) : fun_scope.
-Notation "@ 'sorted_le_nth'" := (deprecate sorted_le_nth sorted_leq_nth)
+Notation "@ 'sorted_le_nth'" :=
+  (fun (T : Type) (leT : rel T) (leT_tr : transitive leT) =>
+     deprecate sorted_le_nth sorted_leq_nth leT_tr)
   (at level 10, only parsing) : fun_scope.
-Notation "@ 'ltn_index'" := (deprecate ltn_index sorted_ltn_index)
+Notation "@ 'ltn_index'" :=
+  (fun (T : eqType) (leT : rel T) (leT_tr : transitive leT) =>
+     deprecate ltn_index sorted_ltn_index leT_tr)
   (at level 10, only parsing) : fun_scope.
-Notation "@ 'leq_index'" := (deprecate leq_index sorted_leq_index)
+Notation "@ 'leq_index'" :=
+  (fun (T : eqType) (leT : rel T) (leT_tr : transitive leT) =>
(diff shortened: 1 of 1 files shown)


mathcomp/ssreflect/path.v
```diff
@@ -1462,15 +1462,12 @@ Notation eq_sorted :=
...
-     @sorted_le_nth _ _ leT_tr leT_refl x0 _ s_sorted _ _) (only parsing).
-Notation ltn_index :=
-  (fun leT_tr s_sorted => @ltn_index _ _ leT_tr _ s_sorted _ _) (only parsing).
+  (fun leT_tr leT_refl x0 => @sorted_le_nth leT_tr leT_refl x0 _)
+  (only parsing).
+Notation ltn_index := (fun leT_tr => @ltn_index _ _ leT_tr _) (only parsing).
 Notation leq_index :=
-  (fun leT_tr leT_refl s_sorted =>
-     @leq_index _ _ leT_tr leT_refl _ s_sorted _ _) (only parsing).
+  (fun leT_tr leT_refl => @leq_index _ _ leT_tr leT_refl _) (only parsing).
```
chdoc, line 1471: This one still does not behave the same as `sorted_leq_index` wrt to apply:
```coq
Goal forall (s : seq nat) (srt_s : sorted leq s), {in s &, forall x y, index x s <= index y s -> x <= y}.
move => s srt_s. 
Fail apply: leq_index.
apply: sorted_leq_index. 
```
This is literally the first example I tried.

chdoc (changes requested):
The sorted_(le|lt)_nth notations seem to work now.

mathcomp/ssreflect/path.v
pi8027: So we should remove the first two placeholders here as well.

pi8027:
@chdoc @CohenCyril As in the description of this PR, some deprecation notations for explicit application such as `@sorted_le_nth` do not let us write the first two implicit arguments `T` and `leT`. IIUC, this is a limitation of the deprecation mechanism. I think we should rather turn them into notations without deprecation as in #640 for now. What do you think?

CohenCyril:
> @chdoc @CohenCyril As written in the description of this PR, some deprecation notations for explicit application such as `@sorted_le_nth` do not let us write the first two implicit arguments `T` and `leT`. IIUC, this is a limitation of the deprecation mechanism of MathComp. I think we should rather turn them into notations without deprecation as in #640 for now. What do you think?

How about setting an intermediate notation e.g. `sorted_leq_nth_subdef` with no maximal implicit?

pi8027:
@CohenCyril If we pass such `sorted_leq_nth_subdef` as the second argument of `deprecate`, the deprecation message says `use sorted_leq_nth_subdef instead`.

pi8027:
OK, I found a solution. This should be fine.

chdoc (approved):
Seems to work now.

PR #680, opened 2020-12-02
Add `pairwise` predicate and some missing lemmas

### LINKED ISSUE #639: Missing lemma `cycle_all2rel`
which roughly says that `transitive leT -> cycle leT =i all2rel leT`. Its `_in` counterpart is also missing. https://github.com/math-comp/math-comp/pull/632#discussion_r520532829

### LINKED ISSUE #676: `pairwise` predicate
> On second thoughts, I prefer to reserve `pairwise` for
> ```coq
> Fixpoint pairwise {T} (r : rel T) (xs : seq T) :=
>   if xs is x :: xs then all (r x) xs && pairwise r xs else true.
> ```
> That might come up in mathcomp 1.13 and that will be equivalent to `all2rel` under `reflexive r` and `symmetric r` only...

_Originally posted by @CohenCyril in https://github.com/math-comp/math-comp/pull/665#discussion_r527610388_

> BTW, `pairwise` is also equivalent to `sorted` under transitivity. Reformulating some lemmas about `sorted` using `pairwise` may allow us to drop the transitivity assumption.

_Originally posted by @pi8027 in https://github.com/math-comp/math-comp/pull/665#discussion_r530370885_

pi8027:

This PR introduces a higher-order predicate `pairwise r xs` which asserts that the relation `r` holds for any i-th and j-th element of `xs` such that i < j.

Closes #639
Closes #676

##### Things done/to do

### DIFF
--- a/CHANGELOG_UNRELEASED.md
+++ b/CHANGELOG_UNRELEASED.md
@@ -10,9 +10,25 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Added
 
+- in `seq.v`,
+  + new higher-order predicate `pairwise r xs` which asserts that the relation
+    `r` holds for any i-th and j-th element of `xs` such that i < j.
+  + new lemmas `allrel_mask(l|r)`, `allrel_filter(l|r)`, `sub(_in)_allrel`,
+    `pairwise_cons`, `pairwise_cat`, `pairwise_rcons`, `pairwise2`,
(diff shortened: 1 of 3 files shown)


pi8027:
Shouldn't `(sub|eq)_(path|cycle|sorted)_in` rather be renamed to `(sub|eq)_in_(path|cycle|sorted)`? (Sorry, my bad.)

CohenCyril (approved):
great!

PR #685, opened 2020-12-11
Change the interpretation scope of some nullary notations from `ring_scope` to `fun_scope`

pi8027:

so that we do not have to write `%R%R`.

##### Things done/to do


CohenCyril (approved):
Indeed `+%R%R` did not make sense... and the empty scope should not be used...

PR #687, opened 2020-12-15
nicer declarations of mixins

gares:
This PR trying to use Coq-Elpi's 1.8.x support for `let (...)` but apperently something goes wrong.


CohenCyril:
in master ? :laughing:

PR #696, opened 2021-01-18
Adding lemma sumnB

CohenCyril:

Missing lemma for sums of subtractions of natural numbers.
Cf https://stackoverflow.com/questions/61556710.

##### Things done/to do

### DIFF
--- a/CHANGELOG_UNRELEASED.md
+++ b/CHANGELOG_UNRELEASED.md
@@ -10,7 +10,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Added
 
-- in `ssrnat.v`, added lemma `sumnB`.
+- in `bigop.v`, added lemma `sumnB`.
 
 - in `seq.v`,
   + new higher-order predicate `pairwise r xs` which asserts that the relation
--- a/CHANGELOG_UNRELEASED.md
+++ b/CHANGELOG_UNRELEASED.md
@@ -10,6 +10,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Added
 
+- in `ssrnat.v`, added lemma `sumnB`.
+
 - in `seq.v`,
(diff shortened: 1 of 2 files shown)


ybertot:
There is clever use of  view sequencing in the one-liner proof.  This could deserve some elaborated explanation in some educational material, but this file is not the right place for such a documentation.  So, as far as I am concerned, **this is good to merge**.

ybertot (approved):

CHANGELOG_UNRELEASED.md
```diff
@@ -10,6 +10,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Added
 
+- in `ssrnat.v`, added lemma `sumnB`.
```
anton-trunov, line 13: Should this be as follows?
```suggestion
- in `bigop.v`, added lemma `sumnB`.
```

ybertot: I did clone the repository and tried to animate the proof.  I went first to ssrnat (because of the CHANGELOG information) and then thought "I am stupid, this is obviously a bigop operation).  I spent so much time studying the code that I did not realize the mishap with not finding sumnB in ssrnat was really a bug in the changelog.  I agree with Anton, this needs to be changed.

ybertot (changes requested):
Change "CHANGELOG" according to Anton's remark: ssrnat should be replace with bigop

CohenCyril:
> There is clever use of view sequencing in the one-liner proof. This could deserve some elaborated explanation in some educational material, but this file is not the right place for such a documentation. So, as far as I am concerned, **this is good to merge**.

FTR the explanation is here https://stackoverflow.com/questions/61556710/distributing-subtraction-over-bigop#answer-65779039

ybertot (approved):
Thanks to all

CohenCyril:
@ybertot is something blocking?

ybertot:
Thanks @CohenCyril, @anton-trunov

PR #703, opened 2021-03-04
Adding big block matrices

CohenCyril:

This is part of #207.  We add the analoguous of  `\matrix_(i, j) _` `\row_i _`, `\col_i _` and `diag_mx` but for block matrices instead of single elements.

##### Things done/to do

### DIFF
--- a/CHANGELOG_UNRELEASED.md
+++ b/CHANGELOG_UNRELEASED.md
@@ -52,6 +52,63 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
     `enum_val_bij`, `le_enum_val`, `le_enum_rank_in`, and
     `le_enum_rank`.  They are all accessible via module `Order` if one
     chooses not to import `Order.EnumVal`.
+  + a new module `tagnat` which provides a monotonous bijection
+    between the finite ordered types `{i : 'I_n & 'I_(p_ i)}`
+    (canonically ordered lexicographically), and `'I_(\sum_i p_ i)`,
+    via the functions `sig` and `rank`. We provide direct access to
+    the components of the former type via the functions `sig1`, `sig2`
+    and `Rank`. We provide the following lemmas on these definitions:
+    `card`, `sigK`, `sig_inj`, `rankK`, `rank_inj`, `sigE12`,
+    `rankE`, `sig2K`, `Rank1K`, `Rank2K`, `rank_bij`, `sig_bij `,
+    `rank_bij_on`, `sig_bij_on`, `le_sig`, `le_sig1`, `le_rank`,
+    `le_Rank`, `lt_sig`, `lt_rank`, `lt_Rank`, `eq_Rank`, `rankEsum`,
+    `RankEsum`, `rect`, and `eqRank`.
+
+- in `matrix.v`, seven new definitions:
+  + `mxblock`, `mxcol`, `mxrow` and `mxdiag` with notations
+    `\mxblock_(i < m, j < n) B_ i j`, `\mxcol_(i < m) B_ i`,
+    `\mxrow_(j < n) B_ j`, and `\mxdiag_(i < n) B_ i` (and variants
+    without explicit `< n`), to form big matrices described by blocks.
+  + `submxblock`, `submxcol` and `submxrow` to extract blocks from the
+    former constructions. There is no analogous for `mxdiag` because
+    one can simply apply `submxblock A i i`.
+  + We provide the following lemmas on these definitions:
+    `mxblockEh`, `mxblockEv`, `submxblockEh`, `submxblockEv`,
+    `mxblockK`, `mxrowK`, `mxcolK`, `submxrow_matrix`,
+    `submxcol_matrix`, `submxblockK`, `submxrowK`, `submxcolK`,
+    `mxblockP`, `mxrowP`, `mxcolP`, `eq_mxblockP`, `eq_mxblock`,
+    `eq_mxrowP`, `eq_mxrow`, `eq_mxcolP`, `eq_mxcol`, `row_mxrow`,
+    `col_mxrow`, `row_mxcol`, `col_mxcol`, `row_mxblock`,
+    `col_mxblock`, `tr_mxblock`, `tr_mxrow`, `tr_mxcol`,
+    `tr_submxblock`, `tr_submxrow`, `tr_submxcol`, `mxsize_recl`,
+    `mxrow_recl`, `mxcol_recu`, `mxblock_recl`, `mxblock_recu`,
+    `mxblock_recul`, `mxrowEblock`, `mxcolEblock`, `mxEmxrow`,
+    `mxEmxcol`, `mxEmxblock`, `mxrowD`, `mxrowN`, `mxrowB`, `mxrow0`,
+    `mxrow_const`, `mxrow_sum`, `submxrowD`, `submxrowN`, `submxrowB`,
+    `submxrow0`, `submxrow_sum`, `mul_mxrow`, `mul_submxrow`,
+    `mxcolD`, `mxcolN`, `mxcolB`, `mxcol0`, `mxcol_const`,
+    `mxcol_sum`, `submxcolD`, `submxcolN`, `submxcolB`, `submxcol0`,
+    `submxcol_sum`, `mxcol_mul`, `submxcol_mul`, `mxblockD`,
+    `mxblockN`, `mxblockB`, `mxblock0`, `mxblock_const`,
+    `mxblock_sum`, `submxblockD`, `submxblockN`, `submxblockB`,
+    `submxblock0`, `submxblock_sum`, `mul_mxrow_mxcol`,
+    `mul_mxcol_mxrow`, `mul_mxrow_mxblock`, `mul_mxblock_mxrow`,
+    `mul_mxblock`, `is_trig_mxblockP`, `is_trig_mxblock`,
+    `is_diag_mxblockP`, `is_diag_mxblock`, `submxblock_diag`,
+    `eq_mxdiagP`, `eq_mxdiag`, `mxdiagD`, `mxdiagN`, `mxdiagB`,
+    `mxdiag0`, `mxdiag_sum`, `tr_mxdiag`, `row_mxdiag`, `col_mxdiag`,
+    `mxdiag_recl`, `mxtrace_mxblock`, `mxdiagZ`, `diag_mxrow`,
+    `mxtrace_mxdiag`, `mul_mxdiag_mxcol`, `mul_mxrow_mxdiag`,
+    `mul_mxblock_mxdiag`, and `mul_mxdiag_mxblock`.
+   + adding missing lemmas `trmx_conform` and `eq_castmx`.
+
+- in `mxalgegra.v`,
+  + Lemmas about rank of block matrices with `0`s inside
+    `rank_col_mx0`, `rank_col_0mx`, `rank_row_mx0`, `rank_row_0mx`,
+    `rank_diag_block_mx`, and `rank_mxdiag`.
+  + we provide an equality of spaces `eqmx_col` between `\mxcol_i V_
+    i` and the sum of spaces `\sum_i <<V_ i>>)%MS`.
+
 ### Changed
 
 - In `ssralg.v` and `ssrint.v`, the nullary ring notations `-%R`, `+%R`, `*%R`,
--- a/mathcomp/algebra/matrix.v
+++ b/mathcomp/algebra/matrix.v
@@ -1,7 +1,7 @@
 (* (c) Copyright 2006-2016 Microsoft Corporation and Inria.                  *)
 (* Distributed under the terms of CeCILL-B.                                  *)
 From mathcomp Require Import ssreflect ssrbool ssrfun eqtype ssrnat seq choice.
-From mathcomp Require Import fintype finfun bigop finset fingroup perm.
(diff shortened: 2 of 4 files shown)


CHANGELOG_UNRELEASED.md
```diff
@@ -26,6 +26,52 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 - in `interval.v`, new lemmas: `ge_pinfty`, `le_ninfty`, `gt_pinfty`, `lt_ninfty`.
 
+- in `order.v`
```
thery, line 29: Could this be a separate commit?

mathcomp/algebra/matrix.v
```diff
@@ -49,6 +49,21 @@ From mathcomp Require Import div prime binomial ssralg finalg zmodp countalg.
 (*                   same width).            \ Ad /                           *)
 (* block_mx Aul Aur Adl Adr == the block matrix / Aul Aur \                   *)
 (*                                              \ Adl Adr /                   *)
+(* \mxrow_(i < n) B i ==  the block m * (\sum_i p_i) matrix                   *)
```
thery, line 52: Are they places where this new constructor could be used in the current library?

mathcomp/ssreflect/order.v
```diff
@@ -399,6 +403,20 @@ From mathcomp Require Import path fintype tuple bigop finset div prime.
...
+(* whenever the type is porderType, and their monotonicity is provided if     *)
+(* this order is total. The theory is in the module Order (Order.enum_valK,   *)
+(* Order.enum_rank_inK, etc) but Order.Enum can be imported to shorten these. *)
+(******************************************************************************)
+(* We provide an opaque monotonous bijection tagnat.val / tagnat.rank between *)
+(* the finite types {i : 'I_n & 'I_(p_ i)} and 'I_(\sum_i p_ i):              *)
+(*  tagnat.val   : 'I_(\sum_i p_ i) -> {i : 'I_n & 'I_(p_ i)}                 *)
+(*  tagnat.rank  : {i : 'I_n & 'I_(p_ i)} -> 'I_(\sum_i p_ i)                 *)
+(*  tagnat.val1  : 'I_(\sum_i p_ i) -> 'I_n                                   *)
+(*  tagnat.val2  : forall p : 'I_(\sum_i p_ i), 'I_(p_ (tagnat.val1 p))       *)
```
thery, line 417: I was expected `tagnat.tag` for` val1`  and `tagnat.nat` for `val2` but maybe it is not a good idea.

mathcomp/algebra/matrix.v
```diff
@@ -3686,3 +3716,67 @@ End mxOver.
...
+  \matrix_(j, k)
+   if tagnat.val1 j =P tagnat.val1 k isn't ReflectT e then 0
+   else B_ (tagnat.val1 k)
+     (cast_ord (congr1 p_ e) (tagnat.val2 j)) (tagnat.val2 k).
+
+Notation "\mxcol_ i E" := (mxcol (fun i => E)) : ring_scope.
+Notation "\mxrow_ i E" := (mxrow (fun i => E)) : ring_scope.
+Notation "\mxdiag_ i E" := (mxdiag (fun i => E)) : ring_scope.
+
+Lemma mulmx_row_col m p
```
thery, line 3739: A pity these new constructions come with very little theorems associated with
them.
I was thinking for example about the trace of `mxdiag` or the transpose of `mxcol`

CHANGELOG_UNRELEASED.md
CohenCyril: How do you suggest to split? (They are related in cascade, each idem depends on the previous one)

mathcomp/algebra/matrix.v
CohenCyril: Not yet, but that's all part of a bigger plan: https://github.com/math-comp/math-comp/pull/207/files#diff-f7a28e39a52f7013e677ad2bab9930222f5af8c0df4acb10e978bf2a1a2fa37fR291-R318

mathcomp/ssreflect/order.v
CohenCyril: I'm really open to any name here. Yours are funny, it also made me think `tagnat.val1` might not have to be a natural number... so that there could be a further generalisation... as long as it is an ordered finite type...

mathcomp/algebra/matrix.v
CohenCyril: Sure. I did not do the closure, I wrote the bare minimum for https://github.com/math-comp/math-comp/pull/207/files#diff-f7a28e39a52f7013e677ad2bab9930222f5af8c0df4acb10e978bf2a1a2fa37fR291-R318 to go through...
So what should I add?
- the trace of mxdiag
- the transpose of all of them
- I guess `mulmx_row_diag`
- maybe `mulmx_col_row` if it makes sense...
other ideas?

CohenCyril: Maybe stuff with `0`s and `1`s and `a%:M`

mathcomp/ssreflect/order.v
CohenCyril: Actually for now (because of the restriction on finite ordered types) its preferable to stick to the first component being a nat. It's a bit silly, but it could be generalized later. Also I prefer to keep `val1` and `val2` string-related to `val`, because it's easier for me to remember...

mathcomp/algebra/matrix.v
thery: I guess there are lots in fact, for example the rank of `\mx_diag` also is there an equivalent of `submx`

CHANGELOG_UNRELEASED.md
thery: I was just thinking that the first 3 items were mostly independent of block matrices. 
Tagnat instead seems more closely related to matrices.

mathcomp/algebra/matrix.v
thery: I was thinking at places where one has been using \matrix_ where one of your constructor could be more natural but a rapid grep showed me nothing

mathcomp/ssreflect/order.v
thery: Ok just it made explicit the encoding of tagnat but that'fine

CohenCyril:
@thery  I did the split -> #710

CohenCyril:
I also dramatically enhanced the block matrix theory... It will be here soon.

CohenCyril:
@thery I ended up adding a couple lemmas, feels rather complete now.

mathcomp/algebra/matrix.v
CohenCyril: Thanks for checking

CohenCyril:
@thery FYI I added a commit renaming `val` to `sig`: it causes less undesirable name-clashes, especially if one wants to import `tagnat`, as in [matrix.v](https://github.com/math-comp/math-comp/pull/703/files#diff-f457cb0bf11677ea475ce7d07aa9d8ce79d060ef642432808d92ceb55d6a3fc5R3743).

thery:
@cyril looks good. Will I have a closer look on monday.

CHANGELOG_UNRELEASED.md
```diff
@@ -52,6 +52,62 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
...
+    the components of the former type via the functions `sig1`, `sig2`
+    and `Rank`. We provide the following lemmas on these definitions:
+    `card`, `sigK`, `sig_inj`, `rankK`, `rank_inj`, `sigE12`,
+    `rankE`, `sig2K`, `Rank1K`, `Rank2K`, `rank_bij`, `sig_bij `,
+    `rank_bij_on`, `sig_bij_on`, `le_sig`, `le_sig1`, `le_rank`,
+    `le_Rank`, `lt_sig`, `lt_rank`, `lt_Rank`, `eq_Rank`, `rankEsum`,
+    `RankEsum`, `rect`, and `eqRank`.
+
+- in `matrix.v`, seven new definitions:
+  + `mxblock` `mxcol`, `mxrow` and `mxdiag` with notations
```
thery, line 68: mxblock,

CHANGELOG_UNRELEASED.md
```diff
@@ -52,6 +52,62 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
...
+    `\mxblock_(i < m, j < n) B_ i j`, `\mxcol_(i < m) B_ i`,
+    `\mxrow_(j < n) B_ j`, and `\mxdiag_(i < n) B_ i` (and variants
+    without explicit `< n`), to form big matrices described by blocks.
+  + `submxblock`, `submxcol` and `submxrow` to extract blocks from the
```
thery, line 72: why not submxdiag?

CHANGELOG_UNRELEASED.md
```diff
@@ -52,6 +52,62 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
...
+    former constructions.
+  + We provides the following lemmas on these definitions:
```
thery, line 74: provide

thery (commented):
impressive!

CHANGELOG_UNRELEASED.md
CohenCyril: This would be redundant with `submxblock A i i`

CohenCyril: I could add it as a notation, but I'm not convinced it is very useful

CHANGELOG_UNRELEASED.md
CohenCyril:
```suggestion
  + `mxblock`, `mxcol`, `mxrow` and `mxdiag` with notations
```

CHANGELOG_UNRELEASED.md
CohenCyril:
```suggestion
  + We provide the following lemmas on these definitions:
```

CHANGELOG_UNRELEASED.md
CohenCyril: cf https://github.com/math-comp/math-comp/pull/703/files#diff-f457cb0bf11677ea475ce7d07aa9d8ce79d060ef642432808d92ceb55d6a3fc5R4371-R4373

thery: maybe just adding a comment

CohenCyril: Where and what comment would you think is appropriate?
Something like ``there is no analogous for `mxdiag` because one can simply apply `submxblock A i i` ``?

CHANGELOG_UNRELEASED.md
CohenCyril: @thery if you resolve the issue by hand I cannot commit the "suggested change" any more. Please leave this kind of comment open until github automatically resolves them when I push the change. :smile_cat:

CHANGELOG_UNRELEASED.md
CohenCyril: done

CHANGELOG_UNRELEASED.md
CohenCyril: done

CHANGELOG_UNRELEASED.md
```diff
@@ -52,6 +52,62 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
...
+    `le_Rank`, `lt_sig`, `lt_rank`, `lt_Rank`, `eq_Rank`, `rankEsum`,
+    `RankEsum`, `rect`, and `eqRank`.
+
+- in `matrix.v`, seven new definitions:
+  + `mxblock`, `mxcol`, `mxrow` and `mxdiag` with notations
+    `\mxblock_(i < m, j < n) B_ i j`, `\mxcol_(i < m) B_ i`,
+    `\mxrow_(j < n) B_ j`, and `\mxdiag_(i < n) B_ i` (and variants
+    without explicit `< n`), to form big matrices described by blocks.
+  + `submxblock`, `submxcol` and `submxrow` to extract blocks from the
+    former constructions.
```
CohenCyril, line 73:
```suggestion
    former constructions. There is no analogous for `mxdiag` because
    one can simply apply `submxblock A i i`.
```

CHANGELOG_UNRELEASED.md
thery: yes in the documention of `matrix.v`.
This makes me realised that there is no doc for `subblock` and the others.

CohenCyril: See the new revision

thery: this should go in the header of `matrix.v` with the documentation of `subblock` and al

CohenCyril: I added a comment about the absence of `submxdiag` juste below the line you pointed out.
For the other subblocks I am tempted to open an issue instead of fixing it here.

thery: I am not sure I undertand. 
You have introduced new constructors `submxblock` ... so you should add their documentation in the header of `matrix`. I think the comment about the absense of `submxdiag` should go there too.

CohenCyril: Ah you are right, sorry I got confused between the changelog and the header

CohenCyril:
@thery I addressed all your comments.

thery (approved):

CohenCyril:
@thery Thanks!

PR #708, opened 2021-03-04
Silencing Hint Locality warning

### LINKED ISSUE #705: Attend warning `deprecated-hint-without-locality`
This message shows up regularly with Coq 8.13:
```
Warning: The default value for hint locality is currently "local" in a
section and "global" otherwise, but is scheduled to change in a future
release. For the time being, adding hints outside of sections without
specifying an explicit locality is therefore deprecated. It is recommended to
use "export" whenever possible. [deprecated-hint-without-locality,deprecated]
```

For now PR #708 just silences the warning, but a longer term solution must be used once Coq 8.11 and Coq 8.12 support is dropped.

CohenCyril:

For now, the `#[global] Hint` attribute is only partially implemented
in Coq 8.13 and unimplemented in Coq 8.11 and Coq 8.12 so we have no
other choice than silencing it for the time being.

Does not really solve #705 in a convincing manner.

A cleanup issue must be open subsequent to the merge of this PR and
targetted to a milestone several versions (TBD) in the future in order to
reactivate the warning (and possibly turn it into a bug) with a proper fix.

##### Things done/to do

### DIFF
--- a/mathcomp/Make
+++ b/mathcomp/Make
@@ -100,3 +100,4 @@ ssreflect/tuple.v
 -arg -w -arg -ambiguous-paths
 -arg -w -arg +non-primitive-record
 -arg -w -arg +undeclared-scope
+-arg -w -arg -deprecated-hint-without-locality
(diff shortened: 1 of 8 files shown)


CohenCyril:
@gares I think you followed this on the Coq side. Is that the right solution for now?

gares:
I'm afraid so, be we should put this problem in some milestone and signal we can't yet put the attribute because of backward compat

PR #727, opened 2021-03-19
Refactor some lemmas about `path.sort`

### LINKED ISSUE #648: Remove the deprecation aliases for `eq_sorted` and attribute them to new lemmas
This is a cleanup issue. Following the merging of #646, it remains to:
- remove the following deprecation aliases https://github.com/math-comp/math-comp/blob/2f906481423530b03f21ff605aaee3f281131608/mathcomp/ssreflect/path.v#L1465-L1472 https://github.com/math-comp/math-comp/blob/2f906481423530b03f21ff605aaee3f281131608/mathcomp/ssreflect/order.v#L3425-L3430
- add the following lemmas in `path.v`
  ```coq
  eq_in_sorted
       : forall (T : Type) (P : {pred T}) (e e' : rel T), {in P &, e =2 e'} ->
         forall s : seq T, all P s -> sorted e s = sorted e' s

  eq_sorted
       : forall (T : Type) (e e' : rel T), e =2 e' -> sorted e =1 sorted e'
  ```

pi8027:

- in `seq.v`, add new lemmas `cat_nilp`, `rev_nilp`, `allrelT`, `allrel_relI`, and `pairwise_relI`.
- in `path.v`:
  + add new lemmas: `pairwise_sorted`, `path_relI`, `cycle_relI`, `sorted_relI`, `eq(_in)_sorted`, `mergeA`, `all_merge`, and `sort_map_homo(_in)`,
  + remove the deprecation aliases `eq(_in)_sorted`, and
  + lemmas `merge_path`, `merge_sorted`, and `sort_sorted(_in)` are now derived from their stable variants.
- in `order.v`, remove the deprecation aliases `eq_sorted_(le|lt)`.

Closes #648

##### Things done/to do

### DIFF
--- a/mathcomp/ssreflect/path.v
+++ b/mathcomp/ssreflect/path.v
@@ -149,11 +149,14 @@ Proof. by case: s => //= y s /andP[]. Qed.
 Lemma path_min_sorted x s : all (e x) s -> path x s = sorted s.
 Proof. by case: s => //= y s /andP [->]. Qed.
 
+Lemma pairwise_sorted s : pairwise e s -> sorted s.
+Proof. by elim: s => //= x s IHs /andP[/path_min_sorted -> /IHs]. Qed.
+
 End Path.
 
-Section RevPath.
+Section PathEq.
 
-Variables (e : rel T).
+Variables (e e' : rel T).
 
 Lemma rev_path x p :
   path e (last x p) (rev (belast x p)) = path (fun z => e^~ z) x p.
@@ -171,7 +174,19 @@ Qed.
 Lemma rev_sorted p : sorted e (rev p) = sorted (fun z => e^~ z) p.
 Proof. by case: p => //= x p; rewrite -rev_path lastI rev_rcons. Qed.
 
-End RevPath.
+Lemma path_relI x s :
+  path [rel x y | e x y && e' x y] x s = path e x s && path e' x s.
+Proof. by elim: s x => //= y s IHs x; rewrite andbACA IHs. Qed.
+
+Lemma cycle_relI s :
+  cycle [rel x y | e x y && e' x y] s = cycle e s && cycle e' s.
+Proof. by case: s => [|? ?]; last apply: path_relI. Qed.
+
+Lemma sorted_relI s :
+  sorted [rel x y | e x y && e' x y] s = sorted e s && sorted e' s.
+Proof. by case: s; last apply: path_relI. Qed.
+
+End PathEq.
 
 Section SubPath_in.
 
@@ -977,6 +992,19 @@ Proof. by rewrite sorted_pairwise // pairwise_cat => /and3P[/allrel_merge]. Qed.
 Lemma sorted_sort s : sorted leT s -> sort s = s.
 Proof. by rewrite sorted_pairwise //; apply/pairwise_sort. Qed.
 
+Lemma mergeA : associative merge.
+Proof.
+elim=> // x xs IHxs; elim=> // y ys IHys; elim=> [|z zs IHzs] /=.
+  by case: ifP.
+case: ifP; case: ifP => /= lexy leyz.
+- by rewrite lexy (leT_tr lexy leyz) -IHxs /= leyz.
+- by rewrite lexy leyz -IHys.
+- case: ifP => lexz; first by rewrite -IHxs //= leyz.
+  by rewrite -!/(merge (_ :: _)) IHzs /= lexy.
+- suff->: leT x z = false by rewrite leyz // -!/(merge (_ :: _)) IHzs /= lexy.
+  by apply/contraFF/leT_tr: leyz; have := leT_total x y; rewrite lexy.
+Qed.
+
 End SortSeq.
 
 Arguments merge {T} relT !s1 !s2 : rename.
@@ -987,6 +1015,7 @@ Arguments merge_path {T leT} leT_total {x s1 s2}.
 Arguments merge_sorted {T leT} leT_total {s1 s2}.
 Arguments sorted_merge {T leT} leT_tr {s t}.
 Arguments sorted_sort {T leT} leT_tr {s}.
+Arguments mergeA {T leT} leT_total leT_tr.
 
 Section SortMap.
 Variables (T T' : Type) (f : T' -> T).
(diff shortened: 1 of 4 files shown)


pi8027:
I will port some more lemmas from [coq-stablesort](https://github.com/pi8027/stablesort) which is another library dedicated to stability proofs of sort and highly optimized sorting algorithms. (It will take some more time I guess.)

CohenCyril (approved):
Very nice

mathcomp/ssreflect/path.v
```diff
@@ -168,7 +171,19 @@ Qed.
...
+  path e x s && path e' x s = path [rel x y | e x y && e' x y] x s.
+Proof. by elim: s x => //= y s IHs x; rewrite andbACA IHs. Qed.
+
+Lemma cycle_relI s :
+  cycle e s && cycle e' s = cycle [rel x y | e x y && e' x y] s.
+Proof. by case: s => [|? ?]; last apply: path_relI. Qed.
+
+Lemma sorted_relI s :
+  sorted e s && sorted e' s = sorted [rel x y | e x y && e' x y] s.
+Proof. by case: s; last apply: path_relI. Qed.
```
chdoc, line 184: In line with the lemmas above, I think these should be named `relI_*` (rather than `*_relI`) and stated in the converse direction. Should we introduce some notation for relI, (e.g., `[relI e1 & e2]`)?

mathcomp/ssreflect/seq.v
```diff
@@ -302,6 +302,9 @@ Proof. by elim: s1 => //= x s1 ->. Qed.
 Lemma size_cat s1 s2 : size (s1 ++ s2) = size s1 + size s2.
 Proof. by elim: s1 => //= x s1 ->. Qed.
 
+Lemma nilp_cat s1 s2 : nilp (s1 ++ s2) = nilp s1 && nilp s2.
+Proof. by case: s1. Qed.
```
chdoc, line 306: I think this should be `cat_nilp`, as `nilp` is a unary predicate. The same holds for `nilp_rev` below. (I'm not a fan of the special convention for unary predicates, but that's a different discussion).

chdoc (commented):
Apart from the naming issues, this looks good to me.

mathcomp/ssreflect/path.v
pi8027: > In line with the lemmas above, I think these should be named `relI_*` (rather than `*_relI`)

I don't think so: https://github.com/math-comp/math-comp/blob/57af89f67dc3ace1a35fdfdbf32be309e20f0108/mathcomp/ssreflect/seq.v#L647-L660

> and stated in the converse direction.

Sure.

> Should we introduce some notation for relI, (e.g., `[relI e1 & e2]`)?

Probably, yes.

chdoc: I don't understand why you point to `all` here, which is a binary predicate. `CONTRIBUTING.md` states:
- `naryPredicate_mainSymbol+`
- `mainSymbol_unaryPredicate`
Summoning @CohenCyril

CohenCyril: Well, one can argue whether `sorted` and `path` are really unary and `all` binary 😹. We tend to consider them as following the unary convention in the main argument (the list), but the binary relation plays the same role as the unary predicate in `all`.

I agree the direction should be the converse one indeed (as for `all`).

mathcomp/ssreflect/seq.v
```diff
@@ -3618,6 +3624,14 @@ Lemma allrelP {T S : eqType} {r : T -> S -> bool} {xs ys} :
   reflect {in xs & ys, forall x y, r x y} (allrel r xs ys).
 Proof. by rewrite allrel_allpairsE; exact: all_allpairsP. Qed.
 
+Lemma allrelT {T S : Type} (xs : seq T) (ys : seq S) :
+  allrel (fun _ _ => true) xs ys = true.
+Proof. by elim: xs => //= ? ?; rewrite allrel_consl all_predT. Qed.
+
+Lemma allrel_relI {T S : Type} (r r' : T -> S -> bool) xs ys :
+  allrel r xs ys && allrel r' xs ys = allrel (fun x y => r x y && r' x y) xs ys.
```
CohenCyril, line 3632: Why don't you use `[rel xy | ...]`? I wonder whether we should add `[relI r & r' ]`...

chdoc: Isn't `[rel x y | ...]` is only for homogeneous relations?

mathcomp/ssreflect/path.v
chdoc: Well, while I am not exactly a fan of the special convention for unary predicates, but I would prefer if new names were in line with preexisting names. So that means defaulting to `<main_symbol>_path`, as with the other path lemmas. Indeed, I tend to think of path as unary, where the relation should be inferable from the context and the sequence is nonempty. Even more so for sorted. But with "for all" neither the domain nor the predicate is a natural candidate for being implicit or inferable from the context.

CohenCyril: I'm not 100% fan of the special convention for unary predicates either especially when the predicate is only unary through certain lenses... However, the convention is definitely pervasive and I'm afraid it would affect even more loc than #733 itself, not to mention power users neurons... :smile:

In other `path` lemmas' name  `<main_symbol>_path`, `<main_symbol>` refers to the path sequence argument, not the relation argument. Note that before this PR, AFAICT only `merge_path` and `map_path` do follow this convention, while `path_mask` and `path_filter` are breaking it... we should fix them (in another PR).

What I meant but maybe failed to convey is that, for me, the fact that the `<main_symbol>` is on the relation argument is a good enough reason to switch to the n+2-ary convention in the current case.

mathcomp/ssreflect/seq.v
CohenCyril: Ah yes you are right, ... such a shame though... maybe we can add notations `{rel T}` for unary and `{rel T & T'}` for binary and generalize constructions... probably not very important though...

CohenCyril: The rule should be flipped

pi8027: Done. Shall we open a new issue for `rel`?

mathcomp/ssreflect/path.v
pi8027: I have no strong opinion on this point but agree with the last paragraph by Cyril. Since these lemmas reason about two different graphs `e` and `e'`, `path` can be seen as a binary predicate in this particular case.

chdoc: Indeed, the naming of the path lemmas is not as consistent as one would hope, but that's another issue. I agree that it's hard to justify using the "unary predicate" convention for the cases where the main symbol is not within the "main" argument of a pseudo-unary predicate. So changing the direction is probably sufficient here.

PR #731, opened 2021-03-26
lemmas on bigcup, trivIset, and partition

chdoc:

Lemmas on partitions and related concepts needed when using partitions as colorings for graphs.

##### Things done/to do

### DIFF
--- a/mathcomp/ssreflect/finset.v
+++ b/mathcomp/ssreflect/finset.v
@@ -1647,12 +1647,26 @@ apply: (iffP idP) => [sFU i Pi| sFU].
 by apply/subsetP=> x /bigcupP[i Pi]; apply: (subsetP (sFU i Pi)).
 Qed.
 
-Lemma bigcup_disjoint U P F :
-  (forall i, P i -> [disjoint U & F i]) -> [disjoint U & \bigcup_(i | P i) F i].
+Lemma bigcup0P P F :
+  reflect (forall i, P i -> F i = set0) (\bigcup_(i | P i) F i == set0).
 Proof.
-move=> dUF; rewrite disjoint_sym disjoint_subset.
+rewrite -subset0; apply: (iffP (bigcupsP _ _ _)) => sub0 i /sub0; last by move->.
+by rewrite subset0 => /eqP.
+Qed.
+
+Lemma bigcup_disjointP U P F  :
+  reflect (forall i : I, P i -> [disjoint U & F i])
+          [disjoint U & \bigcup_(i | P i) F i].
+Proof.
+apply: (iffP idP) => [dUF i Pp|dUF].
+  by apply: disjointWr dUF; apply: bigcup_sup.
+rewrite disjoint_sym disjoint_subset.
 by apply/bigcupsP=> i /dUF; rewrite disjoint_sym disjoint_subset.
 Qed.
+
+Lemma bigcup_disjoint U P F :
+  (forall i, P i -> [disjoint U & F i]) -> [disjoint U & \bigcup_(i | P i) F i].
+Proof. by move/bigcup_disjointP. Qed.
 
 Lemma bigcup_setU A B F :
   \bigcup_(i in A :|: B) F i =
@@ -1719,6 +1733,7 @@ Arguments bigcup_sup [T I] j [P F].
 Arguments bigcup_max [T I] j [U P F].
 Arguments bigcupP {T I x P F}.
 Arguments bigcupsP {T I U P F}.
+Arguments bigcup_disjointP {T I U P F}.
 Arguments bigcap_inf [T I] j [P F].
 Arguments bigcap_min [T I] j [U P F].
 Arguments bigcapP {T I x P F}.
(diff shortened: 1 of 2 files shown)


affeldt-aist (approved):
This looks ok to me.
One may maybe wondering whether trivIsetD is useful since it is one two lemmas but that maybe does not hurt.
I didn't see why not using the \notin notation in partition0, it actually looks more natural in the context of this PR.
You may want to revert some of my suggestions.

chdoc:
I just pushed some more simplifications (mainly the proof of `partition_pigenhole`); made use of the brand new `/[! _ ]` feature. I also chose to deprecate `bigcup_disjoint`, because it is merely one direction of the new `bigcup_disjointP` view lemma. Are such deprecations added to the CHANGELOG? If so, where?

mathcomp/ssreflect/finset.v
```diff
@@ -1647,13 +1647,28 @@ apply: (iffP idP) => [sFU i Pi| sFU].
...
 Proof.
-move=> dUF; rewrite disjoint_sym disjoint_subset.
+apply: (iffP idP) => [dUF i Pp|dUF].
+  by apply: disjointWr dUF; apply: bigcup_sup.
+rewrite disjoint_sym disjoint_subset.
 by apply/bigcupsP=> i /dUF; rewrite disjoint_sym disjoint_subset.
 Qed.
 
+#[deprecated(since="mathcomp 1.13.0", note="Use bigcup_disjointP instead.")]
+Lemma bigcup_disjoint U P F :
```
CohenCyril, line 1668:
```suggestion
Lemma bigcup_disjoint U P F :
```
I'd keep it, because it allows for `rewrite bigcup_disjoint`, which has no equivalent with `bigcup_disjointP`, unless you had a 3-step renaming in mind?

mathcomp/ssreflect/finset.v
```diff
@@ -2128,6 +2235,60 @@ rewrite -(def_pblock tiQ Q_E E_A) -(def_pblock tiP _ Ax) ?(sQP E) //.
 by rewrite -(def_pblock tiQ Q_F F_B) -(def_pblock tiP _ Bx) ?(sQP F).
 Qed.
 
+Lemma indexed_partition (I T : finType) (J : {pred I}) (B : I -> {set T}) :
+  let P := [set B i | i in J] in
+  {in J &, forall i j : I, j != i -> [disjoint B i & B j]} ->
+  (forall i : I, J i -> B i != set0) -> partition P (cover P) /\ {in J&, injective B}.
```
CohenCyril, line 2241:
```suggestion
  (forall i : I, J i -> B i != set0) -> partition P (cover P) /\ {in J &, injective B}.
```

mathcomp/ssreflect/finset.v
```diff
@@ -2128,6 +2235,60 @@ rewrite -(def_pblock tiQ Q_E E_A) -(def_pblock tiP _ Ax) ?(sQP E) //.
...
+Proof.
+move=> P disjB inhB.
+have s0NP : set0 \notin P.
```
CohenCyril, line 2244:
```suggestion
move=> P disjB inhB; have s0NP : set0 \notin P.
```

mathcomp/ssreflect/finset.v
chdoc: I'm not sure I agree with this. If one adds the "customary implicits" (`Arguments bigcup_disjointP {T I U P F}.`) then `rewrite bigcup_disjoint` can be replaced by `rewrite (introT bigcup_disjointP)`. Without it one needs some parentheses and `_`. So in the unlikely case that this is really desirable, it can be done. And doesn't this argument apply to almost all reflection lemmas?

chdoc:
Fixed, and rebased/squashed.

chdoc:
@affeldt-aist , @CohenCyril, is there something left to be done, or can someone press the merge button?

mathcomp/ssreflect/finset.v
```diff
@@ -2128,6 +2235,59 @@ rewrite -(def_pblock tiQ Q_E E_A) -(def_pblock tiP _ Ax) ?(sQP E) //.
...
+
+Lemma imset_disjoint (A B : {pred T}) :
+  [disjoint f @: A & f @: B] = [disjoint A & B].
+Proof.
+apply/pred0Pn/pred0Pn => /= [[? /andP[/imsetP[x xA ->]] xB]|].
+  by exists x; rewrite xA -(mem_imset (mem B) x inj_f).
+by move => [x /andP[xA xB]]; exists (f x); rewrite !mem_imset ?xA.
+Qed.
+
+Lemma imset_trivIset : trivIset P = trivIset fP.
```
CohenCyril, line 2267: This equations looks to me like it should be the other way around

mathcomp/ssreflect/finset.v
```diff
@@ -2128,6 +2235,59 @@ rewrite -(def_pblock tiQ Q_E E_A) -(def_pblock tiP _ Ax) ?(sQP E) //.
...
+  by apply: trivP; rewrite imset_f.
+Qed.
+
+Lemma imset0mem : (set0 \in fP) = (set0 \in P).
+Proof.
+apply/imsetP/idP => [[A AP /esym/eqP]|P0]; last by exists set0; rewrite ?imset0.
+by rewrite imset_eq0 => /eqP<-.
+Qed.
+
+Lemma imset_partition : partition P D = partition fP (f @: D).
```
CohenCyril, line 2282: Same (other way around?)

chdoc: Hmm, I agree, but @affeldt-aist already merged this. :flushed:

affeldt-aist: My bad. Maybe a new, small fix-PR?

PR #732, opened 2021-03-29
lemmas about abs in ssrint

affeldt-aist:

I find myself wanting the following lemmas on several occasions:
- lemma `lez_abs` is on the model of `ler_norm`
~~- `{l,g}{e,t}z0_norm` are like `{l,g}{e,t}z0_abs` but easier to rewrite in a context with a `numDomainType`~~
- lemma `normrzE` to be used with lemmas `{l,g}{e,t}r0_norm`

##### Things done/to do

### DIFF
--- a/mathcomp/algebra/ssrint.v
+++ b/mathcomp/algebra/ssrint.v
@@ -1572,6 +1572,9 @@ Proof. by case: (intP m). Qed.
 Lemma ltz0_abs m : (m < 0)%R -> `|m| = - m :> int.
 Proof. by case: (intP m). Qed.
 
+Lemma lez_abs m : m <= `|m|%N :> int.
+Proof. by case: (intP m). Qed.
+
 Lemma absz_sign s : `|(-1) ^+ s| = 1.
 Proof. by rewrite abszX exp1n. Qed.
 
@@ -1597,6 +1600,18 @@ Lemma abszEsg m : (`|m|%:Z = sgz m * m)%R.
 Proof. by rewrite -sgrz -normrEsg. Qed.
 
 End Absz.
+
+Section MoreAbsz.
+Variable R : numDomainType.
+Implicit Type i : int.
+
+Lemma mulr_absz (x : R) i : x *+ `|i| = x *~ `|i|.
+Proof. by rewrite -abszE. Qed.
+
+Lemma natr_absz i : `|i|%:R = `|i|%:~R :> R.
+Proof. by rewrite -abszE. Qed.
+
+End MoreAbsz.
 
 Module Export IntDist.
 
--- a/CHANGELOG_UNRELEASED.md
+++ b/CHANGELOG_UNRELEASED.md
@@ -133,6 +133,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 - in `bigop.v`:
   + Lemmas `big_nat_widenl`, `big_geq_mkord`
+- in `ssrint.v`,
+  + Lemmas: `mulr_absz`, `natr_absz`, `lez_abs`
 
 - In `ssralg.v`
   + new lemma `fmorph_eq`


mathcomp/algebra/ssrint.v
```diff
@@ -1598,6 +1604,18 @@ Proof. by rewrite -sgrz -normrEsg. Qed.
...
+Proof. by move/gez0_abs => <-. Qed.
+
+Lemma gtz0_norm (R : numDomainType) (i : int) : 0 < i -> `|i|%:R = i%:~R :> R.
+Proof. by move/gtz0_abs => <-. Qed.
+
+Lemma ltz0_norm (R : numDomainType) (i : int) : i < 0 -> `|i|%:R = - i%:~R :> R.
+Proof. by move/ltz0_abs; rewrite -mulrNz => <-. Qed.
+
+Lemma lez0_norm (R : numDomainType) (i : int) : i <= 0 -> `|i|%:R = - i%:~R :> R.
+Proof. by move/lez0_abs; rewrite -mulrNz => <-. Qed.
```
pi8027, line 1617: I think that the following lemma could substitute the use of these lemmas:
```coq
Goal forall (R : numDomainType) (i : int), `|i|%:R = `|i|%:~R :> R.
Proof. by move=> R i; rewrite -abszE. Qed.
```

affeldt-aist: > I think that the following lemma could substitute the use of the above lemmas:

I have told that we could have them all. @CohenCyril 
`xyz_norm` lemmas have easy-to-search names, whereas the generic ones look like low-level manipulation of notations.

pi8027: Another issue is that `` `|_| `` in these statements are not `GRing.norm` but `absz`. I think we could copy almost any lemma involving `GRing.norm` that applies to `int` in this way, but I'm not sure if that's what we want.

affeldt-aist: > Another issue is that `` `|_| `` in these statements are not `GRing.norm` but `absz` and
> thus their current names seem wrong.

Indeed, but the name `gez0_abs` is already taken :-/. To me, the "_norm" prefix looked appropriate
for a lemma intended to be used with a `numDomainType`.

> I think we could copy almost any lemma involving `GRing.norm` that applies to `int` in this way,
> but I'm not sure if that's what we want.

Indeed. Rewriting with `gez0_abs` in a context with `numDomainType` still looks cumbersome to me though.

pi8027: If you point me out where you need these lemmas, I will try to propose a fix for that.

@CohenCyril What do you think?

affeldt-aist: These lemmas are used in https://github.com/math-comp/analysis/blob/lebesgue_measure/theories/measure_wip.v.

NB: they are part of code under refactoring.

pi8027:
To make sure these lemmas are not misnamed, let's summon @CohenCyril anyway.

mathcomp/algebra/ssrint.v
```diff
@@ -1598,6 +1601,18 @@ Proof. by rewrite -sgrz -normrEsg. Qed.
...
 
+Section Normz.
+Variable R : numDomainType.
+Implicit Type i : int.
+
+Lemma pmulrnorm (x : R) i : x *+ `|i| = x *~ `|i|.
+Proof. by rewrite -abszE. Qed.
+
+Lemma normrzE i : `|i|%:R = `|i|%:~R :> R.
+Proof. by rewrite -abszE. Qed.
```
CohenCyril, line 1612: In both these lemmas, the LHS is about `abs` not `norm`, moreover the name for the operation `*+` is `natmul` and the shape `mulr*n` or `mulrn*` is wildely use for lemmas about this (e.g. `mulrIn` and for `%:R` it is `natr` (as in `natrD` etc), so I'd pick the following names: `mulr_absz` and `natr_absz`.

CohenCyril (approved):

pi8027 (approved):

PR #740, opened 2021-05-11
Fix some inheritance issues

### LINKED ISSUE #689: Issues with the inheritance from decFieldType to finFieldType
- It is done by constructing a decision procedure and thus non-forgetful. Shouldn't we reimplement `DecidableFieldMixin` as a factory? https://github.com/math-comp/math-comp/blob/ee410eb85f36f5e366f807207a7dee8f3e1bdf3a/mathcomp/algebra/finalg.v#L1066-L1082
- `finFieldType` should also inherit from `countDecFieldType` but it does not. (`hierarchy.ml` did not recognize the above inheritance path since its implementation is irregular. I have a fix for this issue.)

pi8027:

- The way `hierachy.ml` recognizes inheritance has been changed: `S1` inherits  from `S2` when there is a coercion path from `S1.sort` to `S2.sort` and there  is a canonical structure instance that unifies `S1.sort` and `S2.sort`,  regardless of where (which module) these constants are declared.  As a result, it recognizes non-forgetful inheritance and checks the uniqueness  of join and exhaustiveness of canonical declarations involving it.
- `finFieldType` now inherits from `countDecFieldType`. This should probably be done with forgetful inheritance and a factory, but it remains non-forgetful inheritance for now.
- Get rid of a workaround in `ssrnum` thanks to the improvement of `hierarchy.ml`.

Closes #689.

##### Things done/to do

### DIFF
--- a/CHANGELOG_UNRELEASED.md
+++ b/CHANGELOG_UNRELEASED.md
@@ -186,6 +186,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - across the library, the deprecation mechanism to use has been changed from the
   `deprecate` notation to the `deprecated` attribute (cf. Removed section).
 
+- in `finalg.v`, `finFieldType` now inherits from `countDecFieldType`.
+
 ### Renamed
 
 - in `path.v`,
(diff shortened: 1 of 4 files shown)


CHANGELOG_UNRELEASED.md
```diff
@@ -145,4 +147,6 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Infrastructure
 
+- `hierarchy.ml` now recognizes non-forgetful inheritance.
```
CohenCyril, line 150: I'm not sure I understand. It "recognizes" non-forgetful inheritance, but what does it do with it? Warn? Error? Tolerate given some special instructions?

CohenCyril (commented):
LGTM

PR #753, opened 2021-06-12
intdiv.v : Fixed the doc + removed unused import

hivert:
There were typos and incomplete sentences in the doc of intdiv.v. Also finalg was imported but not needed. I fix both.


CohenCyril (approved):
Great!! Thanks

PR #754, opened 2021-06-13
Canonical additive, linear and rmorphism for fst and snd

hivert:
There is zmod, ring, l-module and various Canonical structure on pairs but the projection fst and snd are missing algebraic structures. I'm adding those.

- [x ] added corresponding entries in `CHANGELOG_UNRELEASED.md` (do not edit former entries)

### DIFF
--- a/mathcomp/field/separable.v
+++ b/mathcomp/field/separable.v
@@ -228,7 +228,7 @@ have co_p0_q0: coprimep p0 q0.
     by rewrite mulrC mulrBr mul_polyC addrAC -addrA -opprB -rmorphM -rmorphB.
   have ->: q0 = q2 ^ at_t \Po ('X - y%:P) by rewrite polyCK ?comp_polyXaddC_K.
   apply/coprimep_comp_poly/Bezout_coprimepP; exists (u ^ at_t, v ^ at_t).
-  by rewrite -!rmorphM -rmorphD Dr /= map_polyC polyC_eqp1.
+  by rewrite /= -!rmorphM -rmorphD Dr /= map_polyC polyC_eqp1.
 have{co_p0_q0}: gcdp p0 (q ^ iota) %= 'X - y%:P.
   rewrite /eqp Dq (eqp_dvdl _ (Gauss_gcdpr _ _)) // dvdp_gcdr dvdp_gcd.
   rewrite dvdp_mull // -root_factor_theorem rootE horner_comp !hornerE.
--- a/CHANGELOG_UNRELEASED.md
+++ b/CHANGELOG_UNRELEASED.md
@@ -143,6 +143,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 - In `ssralg.v`
   + new lemma `fmorph_eq`
+  + Canonical additive, linear and rmorphism for `fst` and `snd`
 
 - In `rat.v`
   + new lemmas `minr_rat`, `maxr_rat`
--- a/mathcomp/algebra/ssralg.v
+++ b/mathcomp/algebra/ssralg.v
@@ -6383,6 +6383,13 @@ Proof. by move=> x; congr (_, _); apply: addNr. Qed.
 Definition pair_zmodMixin := ZmodMixin pair_addA pair_addC pair_add0 pair_addN.
 Canonical pair_zmodType := Eval hnf in ZmodType (M1 * M2) pair_zmodMixin.
 
+Fact fst_is_additive : additive fst.
+Proof. by []. Qed.
+Canonical fst_additive := Additive fst_is_additive.
+Fact snd_is_additive : additive snd.
+Proof. by []. Qed.
+Canonical snd_additive := Additive snd_is_additive.
+
 End PairZmod.
 
 Section PairRing.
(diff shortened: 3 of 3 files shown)


pi8027 (changes requested):
They should be re-exported from here. https://github.com/math-comp/math-comp/blob/8a5f000f9da3acc497a0393558f53f79cb20b30f/mathcomp/algebra/ssralg.v#L6165-L6196

pi8027:
Also, this is a missing piece to support product rings in [algebra-tactics](https://github.com/math-comp/algebra-tactics). Thanks!

hivert:
> They should be re-exported from here.

Correct me if I'm wrong, but all those canonicals have to be re-exported because they were defined in a the Theory module. The new ones are defined outside...

hivert:
> Also, this is a missing piece to support product rings in [algebra-tactics](https://github.com/math-comp/algebra-tactics). Thanks!

I don't understand what you mean here (and know very little about tactic programming)...

pi8027:
> > They should be re-exported from here.
> 
> Correct me if I'm wrong, but all those canonicals have to be re-exported because they were defined in a the Theory module. The new ones are defined outside...

Sorry, my bad. Indeed, instances for products are defined after the `GRing` module.

Could you investigate the breakage in separable.v? I guess you have to add some explicit patterns to specify subterms to rewrite.

> > Also, this is a missing piece to support product rings in [algebra-tactics](https://github.com/math-comp/algebra-tactics). Thanks!
> 
> I don't understand what you mean here (and know very little about tactic programming)...

Nevermind. I just wanted to show my support for this PR. But here are some details in the case you are interested in it: the algebra-tactics library provides ring and field tactics (as in Coq) for MathComp with a work-in-progress feature to support morphisms, e.g., it can prove `f ((x * y) ^+ 2) = f x ^+ 2 + f y ^+ 2 + f x * f y *+ 2` automatically, where `f` is a ring morphism to a `comRingType`. By relying on this feature and the canonical instances added by this PR, it should also be able to prove equations on product rings involving `fst`, `snd`, and `pair`.

hivert:
> Could you investigate the breakage in separable.v? I guess you have to add some explicit patterns to specify subterms to rewrite.

I'm doing it right now, but I must confess it seems more complicated than that I expected that replacing the failing

    rewrite -!rmorphM

by

      rewrite -[X in X + _]rmorphM -[X in _ + X]rmorphM.

should work but it doesn't. What is even strange it that 

       set mor := [rmorphism of map_poly at_t].
       rewrite -!(rmorphM mor).

work as expected ! However I expect that you won't accept such a crude fix.
The canonical is here and for some reason Coq fail to find the correct matching. I'm trying some more things, but It may be beyond my skills...


> Nevermind. I just wanted to show my support for this PR. 

Thanks !

> But here are some details in the case you are interested in it: the algebra-tactics library provides ring and field tactics (as in Coq) for MathComp with a work-in-progress feature to support morphisms, e.g., it can prove `f ((x * y) ^+ 2) = f x ^+ 2 + f y ^+ 2 + f x * f y *+ 2` automatically, where `f` is a ring morphism to a `comRingType`. By relying on this feature and the canonical instances added by this PR, it should also be able to prove equations on product rings involving `fst`, `snd`, and `pair`.

That is indeed very cool !

hivert:
I just pushed a reasonable fix. However I've no idea why it used to work and it is failing now. I don't know enough about Coq's pattern matching/canonical machinery to investigate further. If someone is able to understand and give me an explanation, please do !

hivert:
The failure seems to be unrelated... I don't know how to relaunch the test to be sure.

mathcomp/field/separable.v
```diff
@@ -228,7 +228,7 @@ have co_p0_q0: coprimep p0 q0.
     by rewrite mulrC mulrBr mul_polyC addrAC -addrA -opprB -rmorphM -rmorphB.
   have ->: q0 = q2 ^ at_t \Po ('X - y%:P) by rewrite polyCK ?comp_polyXaddC_K.
   apply/coprimep_comp_poly/Bezout_coprimepP; exists (u ^ at_t, v ^ at_t).
-  by rewrite -!rmorphM -rmorphD Dr /= map_polyC polyC_eqp1.
+  by rewrite -![map_poly _ _ * _]rmorphM -rmorphD Dr /= map_polyC polyC_eqp1.
```
pi8027, line 231: After all, I investigated this by myself. The source of the failure here was that it unifies `(u ^ at_t, v ^ at_t).1` with a morphism application and then triggers a canonical structure resolution. Without the morphism instances for `fst`, it unfolds `fst` to obtain `u ^ at_t` and finds the fact that `map_poly` is a morphism, but this behavior changes in presence of the morphism instances for `fst`. So `simpl` before `rewrite -!rmorphM` would be enough here:
```suggestion
  by rewrite /= -!rmorphM -rmorphD Dr /= map_polyC polyC_eqp1.
```

pi8027 (approved):
Except for a small remark, it looks good to me.

mathcomp/field/separable.v
hivert: > After all, I investigated this by myself. The source of the failure here was that it unifies `(u ^ at_t, v ^ at_t).1` with a morphism application and then triggers a canonical structure resolution. Without the morphism instances for `fst`, it unfolds `fst` to obtain `u ^ at_t` and finds the fact that `map_poly` is a morphism, but this behavior changes in presence of the morphism instances for `fst`. So `simpl` before `rewrite -!rmorphM` would be enough here:

Thanks for the investigation and the explanation. I just applied your change (squashing the fix in a single commit) and (force) pushed. That should do the job.

hivert:
Good ! May I take the occasion to kindly ask the mathcomp project if they could put my name somewhere as an external contributor ? I've now probably more than 15 commit in the project, so that I think I deserve it ;-)

pi8027:
We have a list of contributors for each release in its announcement e-mail. But, I personally think it makes a lot of sense to increase its visibility. I added it to the [topics](https://github.com/math-comp/math-comp/wiki/TopicsNextMeeting) for the next MathComp meeting (10 AM CEST tomorrow).

Issue #620, opened 2020-10-29
Missing lemmas of membership equality


pi8027:
```coq
Lemma map_mem_eq  {T A : eqType} (f : T -> A) (s1 s2 : seq T) :
  s1 =i s2 -> [seq f a | a <- s1 ] =i [seq f a | a <- s2]. 

Lemma filter_mem_eq {T : eqType} p (s1 s2 : seq T) :
  s1 =i s2 -> [seq a <- s1 | p a] =i [seq a <- s2 | p a].
```

_Originally posted by @MrSet in https://github.com/math-comp/math-comp/issues/505#issue-617258774_

CohenCyril:
I'm in favor, but I'd name them `eq_mem_map` and `eq_mem_filter`.

chdoc:
There is also some equality lemmas missing, such as
```coq
Lemma eq_in_pmap (aT : eqType) rT (f1 f2 : aT -> option rT) (s : seq aT) : 
  {in s, f1 =1 f2} -> pmap f1 s = pmap f2 s.
```
and possibly the variant without `in`. Moreover, the proofs of these lemmas and (and any other lemmas with `{in s, _}` assumptions that are proved by induction on `s` might be simplified by adding the following helper lemma:
```coq
Lemma forall_cons (T : eqType) (P : T -> Prop) a (s : seq T) : 
  {in a::s, forall x, P x} <-> P a /\ {in s, forall x, P x}.
Proof.
split => [A|[A B]]; last by move => x /predU1P [-> //|]; apply: B.
by split => [|b Hb]; apply: A; rewrite !inE ?eqxx ?Hb. 
Qed.
```
Any other lemmas in the same vein that are missing?

PR #757, opened 2021-06-21
forall/exists_cons + some easy lemmas

chdoc:

* Introduce lemmas `forall_cons` and `exists_cons` to simplify inductive proofs involving `{in s, _}` 
  (and `{subset s1 <= s2}`)
* Some new lemmas suggested in #620 (except `filter_mem_eq`, which is trivial to inline)

closes: #620 

Remark, I'm not sure `forall_cons` and `exists_cons` are the best names, so I'm open to suggestions.

##### Things done/to do

### DIFF
--- a/CHANGELOG_UNRELEASED.md
+++ b/CHANGELOG_UNRELEASED.md
@@ -10,6 +10,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Added
 
+- in `ssrbool.v`, added lemmas `andPP` and `orPP`.
+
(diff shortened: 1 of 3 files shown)


chdoc:
I can take care of this this afternoon.

clayrat:
This PR is breaking https://github.com/imdea-software/fcsl-pcm. The reason seems to be that `T2` in the new version of `eq_in_map` is constrained to `eqType` whereas it was an arbitrary `Type` before. The new proof doesn't seem to actually require this constraint, so I assume it can be removed without consequences?

PR #759, opened 2021-06-30
[bigop] Add lemma for boolean enumeration in bigops

ejgallego:
In some cases, such as in the proof in [1], when we have a bigop
over the boolean enumeration, we need to open the implementation
internals which is not optimal and moreover fragile.

We instead provide a helper lemma that expands a bigop over the
booleans with a concrete enumeration, and which is more
straightforward to use than `big_enumP`.

Note we require commutativity of the operator, this is to safeguard
against changes on the low-level `bool_enumP` enumeration.

Co-authored-by: Amel Kebbouche <akebbouc@outlook.fr>

cc: @Amel-Kebbouche

[1] https://github.com/math-comp/analysis/blob/7b53e30b53b96e5eb3cabe476b0191353191b52c/theories/altreals/distr.v#L336

##### Things done/to do

### DIFF
--- a/CHANGELOG_UNRELEASED.md
+++ b/CHANGELOG_UNRELEASED.md
@@ -178,6 +178,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   `diagonalizableP`, `diagonalizable_conj_diag`, and
   `codiagonalizableP`.
 
+- in `bigop.v`, added lemma `big_bool`.
+
 ### Changed
 
 - In `ssralg.v` and `ssrint.v`, the nullary ring notations `-%R`, `+%R`, `*%R`,
--- a/mathcomp/ssreflect/bigop.v
+++ b/mathcomp/ssreflect/bigop.v
@@ -1865,6 +1865,10 @@ Lemma big_all_cond :
   \big[andb/true]_(i <- r | P i) B i = all [pred i | P i ==> B i] r.
 Proof. by rewrite big_mkcond unlock. Qed.
 
+Lemma big_bool R (idx : R) (op : Monoid.com_law idx) (F : bool -> R):
+  \big[op/idx]_(i : bool) F i = op (F true) (F false).
+Proof. by rewrite /index_enum !unlock /= Monoid.mulm1. Qed.
+
 End Seq.
 
 Section FinType.


mathcomp/ssreflect/bigop.v
```diff
@@ -1865,6 +1865,10 @@ Lemma big_all_cond :
   \big[andb/true]_(i <- r | P i) B i = all [pred i | P i ==> B i] r.
 Proof. by rewrite big_mkcond unlock. Qed.
 
+Lemma big_boolE R (idx : R) (op : R -> R -> R) (F : bool -> R):
+  \big[op/idx]_(i : bool) F i = op (F true) (op (F false) idx).
+Proof. by rewrite /index_enum !unlock. Qed.
```
CohenCyril, line 1870: @ejgallego would that be a problem for you to require the extra constraint that `op` a commutative monoid?

ejgallego: No problem at all I think, PR and commit message updated, thanks!

Now that we have added the commutativity condition on the operator, we could do the proof of this with `big_enumP`, but IMHO it is not worth, but let me know if you would like us to do so.

CohenCyril: BTW, there is no need for the `E`, I think `big_bool` is a good enough name.

ejgallego: Indeed, that was the original name @Amel-Kebbouche picked :) Will amend tomorrow.

CHANGELOG_UNRELEASED.md
```diff
@@ -178,6 +178,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   `diagonalizableP`, `diagonalizable_conj_diag`, and
   `codiagonalizableP`.
 
+- in `bigop.v`, added lemma `big_boolE`.
```
CohenCyril, line 181:
```suggestion
- in `bigop.v`, added lemma `big_bool`.
```

mathcomp/ssreflect/bigop.v
```diff
@@ -1865,6 +1865,10 @@ Lemma big_all_cond :
   \big[andb/true]_(i <- r | P i) B i = all [pred i | P i ==> B i] r.
 Proof. by rewrite big_mkcond unlock. Qed.
 
+Lemma big_boolE R (idx : R) (op : Monoid.com_law idx) (F : bool -> R):
```
CohenCyril, line 1868:
```suggestion
Lemma big_bool R (idx : R) (op : Monoid.com_law idx) (F : bool -> R):
```

mathcomp/ssreflect/bigop.v
```diff
@@ -1865,6 +1865,10 @@ Lemma big_all_cond :
...
+  \big[op/idx]_(i : bool) F i = op (F true) (op (F false) idx).
```
CohenCyril, line 1869: BTW since `op` is now a monoid law this can be:
```suggestion
  \big[op/idx]_(i : bool) F i = op (F true) (F false).
```

mathcomp/ssreflect/bigop.v
```diff
@@ -1865,6 +1865,10 @@ Lemma big_all_cond :
...
+Proof. by rewrite /index_enum !unlock. Qed.
```
CohenCyril, line 1870:
```suggestion
Proof. by rewrite /index_enum !unlock mulm1. Qed.
```

ejgallego:
Thanks for the comments, good catch about the monoid identity, indeed it should be this way.

Issue #762, opened 2021-07-04
Mixin for lattice


hivert:
Currently the mixin for lattice contains a lot of obligations (this is the algebraic point of view on lattices):
```
Record mixin_of (T0 : Type) (b : POrder.class_of T0)
                (T := POrder.Pack tt b) := Mixin {
  meet : T -> T -> T;
  join : T -> T -> T;
  _ : commutative meet;
  _ : commutative join;
  _ : associative meet;
  _ : associative join;
  _ : forall y x, meet x (join x y) = x;
  _ : forall y x, join x (meet x y) = x;
  _ : forall x y, (x <= y) = (meet x y == x);
}.
```
We only need to require that meet is idempotent (ie: meet x x = x), to prove that `<=` is a partial order (see below). So I feel that this mixin is too much to ask.

If we already know that `<=` is a partial order, then a simpler mixin could be (this is the order theory point of view):
```
Record of_ := Build {
  meet : T -> T -> T;
  join : T -> T -> T;
  meetP : forall x y z, (x <= meet y z) = (x <= y) && (x <= z);
  joinP : forall x y z, (join x y <= z) = (x <= z) && (y <= z);
}.
```

I'm not sure if we want to use the order-theory mixin only at creation or to change the internal structure of the mixin to the above one, using coercion to allow creation from different mixin for backward compatibility. 

If we go to the input only solution, I've code at https://github.com/hivert/Coq-Combi/blob/master/theories/SSRcomplements/lattice.v which create a latticeMixin from a mixin above. I'm ok to submit a PR.

#### Here is a proof of my claim:
```
Variables (T : eqType) (meet : T -> T -> T).
Hypothesis (meetI : forall x, meet x x = x)
           (meetC : commutative meet)
           (meetA : associative meet).

Definition le x y := (meet x y == x).

Lemma le_refl x : le x x.
Proof. by rewrite /le meetI. Qed.
Lemma le_trans : transitive le.
Proof.
move=> x y z; rewrite /le => /eqP myx /eqP mxz.
by rewrite -myx -meetA mxz.
Qed.
Lemma le_anti : antisymmetric le.
Proof.
move=> x y; rewrite /le => /andP [/eqP mxy].
by rewrite meetC mxy => /eqP.
Qed.
```

CohenCyril:
[...]

PR #763, opened 2021-07-05
meetJoinLeMixin to construct latticeType

hivert:

Currently the mixin for lattice contains a lot of obligations (see discussion at https://github.com/math-comp/math-comp/issues/762). If we already know that <= is a partial order, then the following simpler mixin is sufficient:
```
Record of_ := Build {
  meet : T -> T -> T;
  join : T -> T -> T;
  meetP : forall x y z, (x <= meet y z) = (x <= y) && (x <= z);
  joinP : forall x y z, (join x y <= z) = (x <= z) && (y <= z);
}.
```
I'm adding de necessary code to be able to construct a lattice from such a mixin. In a latter PR we might want to change the internal mixin of a lattice while keeping the external theory as discussed in the issue.

##### Things done/to do

### DIFF
--- a/mathcomp/ssreflect/order.v
+++ b/mathcomp/ssreflect/order.v
@@ -303,6 +303,11 @@ From mathcomp Require Import path fintype tuple bigop finset div prime finfun.
 (*                    idempotence of meet.                                    *)
 (*                    (can build:  porderType, latticeType, distrLatticeType) *)
 (*                                                                            *)
+(* - meetJoinLeMixin == on a porderType, takes meet, join, and a proof that   *)
+(*                    those are respectvely the greatest lower bound and the  *)
+(*                    least upper bound.                                      *)
+(*                    (can build: latticeType)                                *)
+(*                                                                            *)
 (* - leOrderMixin == on a choiceType, takes le, lt, meet, join,               *)
 (*                   antisymmetry, transitivity and totality of le.           *)
 (*                   (can build:  porderType, latticeType, distrLatticeType,  *)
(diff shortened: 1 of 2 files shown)


mathcomp/ssreflect/order.v
```diff
@@ -5267,6 +5272,85 @@ End Exports.
...
+Qed.
+Next Obligation.
+apply/le_anti/andP; split; first exact: meet_leL.
+by rewrite meetP le_refl join_leL.
+Qed.
+Next Obligation.
+apply/le_anti/andP; split; last exact: join_leL.
+by rewrite joinP le_refl meet_leL.
+Qed.
+Next Obligation. by rewrite eq_le meetP meet_leL le_refl. Qed.
```
CohenCyril, line 5341: Great contribution.
Please do not use `Program`

strub: > Please do not use `Program`

May I ask why? (This is a genuine question, as i find useful to use `Program` in that particular case)

CohenCyril: Sure, there is a tactic which is automatically used after `Next Obligation` which introduces some variables in the context with an automatic name, independently of the variables before the`:` in the statement. This is unstable.
One can also write `Obligation Tactic := idtac.` but then, no introduction at all, even for the variables before `:` in the statement, which you must then reintroduce for every single obligation.

CohenCyril: Also `Program Definition` may fill some obligations automatically (I do not know whether this is dependent or not from the `Obligation Tactic`) and I think it could use an axiom doing so...

CohenCyril: I would love an deterministic alternative to `Program`, but never took the time to complain (I try to focus my complaining on other things :laughing: )

hivert: If you haven't see, I put a not about that in the Contributing section of the wiki. I was not exactly sure where to put it though.

CohenCyril (approved):
LGTM

mathcomp/ssreflect/order.v
CohenCyril: @hivert yes the contributing section is the right place. I do not see your note however...

PR #770, opened 2021-07-21
[ignore]

gares:
This one evaluates the impact on HB.pack, see https://github.com/math-comp/hierarchy-builder/pull/252

### DIFF
--- a/.nix/config.nix
+++ b/.nix/config.nix
@@ -6,6 +6,7 @@
   default-bundle = "coq-8.13";
   bundles = let common = {
     hierarchy-builder.override.version = "master";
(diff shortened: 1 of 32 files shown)


.nix/config.nix
```diff
@@ -15,6 +16,7 @@
       mathcomp-single-planB-src.job = true;
       mathcomp-single-planB.job = true;
     } // common;
+    "coq-8.13".ocamlPackages.elpi.override.version = "v1.13.6";
```
CohenCyril, line 19: `elpi` is "manually" overriden in the `coq-elpi` derivation, do the override in `.nix/coq-overlays/coq-elpi/default.nix`

PR #772, opened 2021-07-26
add multi-rules raddfE rmorphE and linearE

ybertot:

In a zulip conversation, there was a request for an automatic tactic to dispose of goals using the algebraic properties of linear functions.  iterated rewriting with a multi-rule gathering `linearD`, `linearMn`, `linear0`, etc, appears to be a solution to this problem.  The good name for this multi-rule should be `linearE`.

It was also suggested that similar multi-rules `rmorphE` and `raddfE` should also be included.

##### Things done/to do


mathcomp/algebra/ssralg.v
```diff
@@ -6044,6 +6044,14 @@ Definition can2_lrmorphism := can2_lrmorphism.
 Definition bij_lrmorphism := bij_lrmorphism.
 Definition imaginary_exists := imaginary_exists.
 
+Definition raddf := (raddf0, raddfN, raddfD, raddfMn).
+
+Definition rmorphE :=
+  (rmorphD, rmorph0, rmorphB, rmorphN, rmorphMNn, rmorphMn).
```
CohenCyril, line 6050: What about `rmorph1` and `rmorphX`?

CohenCyril:
Thanks for the PR! Additionally could you try to spot a few places in the current code of mathcomp where it would be beneficial to use these multirules and try them out?

ybertot:
Following up on an oral conversation, we should also add more instances of `linearE` in files where more rules can be added.

siraben:
Looks good! I think this would be useful in helping shape the goal into some normal form when dealing with linear functions.

mathcomp/algebra/ssralg.v
gares: @ybertot can you push this change? Using the multirule here and there would be also nice, but not necessary for merging this PR strictly speaking. And we would like to merge this PR before the next release, since it is not a breaking change and looks very useful.

PR #839, opened 2022-01-13
Adding missing lemma sortedP

CohenCyril:

A characterization of `sorted` using `nth`.

##### Things done/to do


mathcomp/ssreflect/path.v
```diff
@@ -143,6 +143,10 @@ Proof. by rewrite -rot_cycle rotrK. Qed.
 
 Definition sorted s := if s is x :: s' then path x s' else true.
 
+Lemma sortedP s x :
+  reflect (forall i, i.+1 < size s -> e (nth x s i) (nth x s i.+1)) (sorted s).
```
proux01, line 147: Maybe
```suggestion
Lemma sortedP x s :
  reflect (forall i, i.+1 < size s -> e (nth x s i) (nth x s i.+1)) (sorted s).
```
(although it is less convenient for the `Arguments` command below)

mathcomp/ssreflect/path.v
```diff
@@ -377,6 +381,7 @@ End Transitive.
 End Paths.
 
 Arguments pathP {T e x p}.
+Arguments sortedP {T e s}.
```
proux01, line 384:
```suggestion
Arguments pathP {T e x p}.
Arguments sortedP {T e} x {s}.
```
if following previous suggestion

proux01 (approved):
Looks straightforward. I'd like to merge by the end of the week if there is no further comment.

mathcomp/ssreflect/path.v
CohenCyril: I mimicked the behaviour of `pathP` (`x0` there plays the role of `x` for `sortedP`) and I'd rather stick to it.

ejgallego:
I cc @jouvelot as he has a similar lemma in our codebase and may want to have a look.

jouvelot:
Thanks, Emilio: indeed sortedP seems quite natural. I'm not sure this is what you had in mind, but the other lemma I was looking for was the existence of a finite list of (improperly ordered) transpositions that, composed together, would yield a sorted list from an unsorted one (think bubble sort).

proux01:
@jouvelot I don't think we have that currently in MathComp, so feel free to pen a separate PR.

jouvelot:
Pierre, 
I wrote a specification and proof for the following BubbleSort-like lemma:
```
bs_spec
     : forall s : seq nat,
       uniq s ->
       exists ts : seq transposition,
         let '(all_bubbles, s') := swap aperm s ts in 
         all_bubbles /\ up_sorted s'
```
which fulfills what I was looking for. Since it's a big complicated, I'm not sure this would be acceptable for a mathcomp PR. Would you be willing to have a quick look at the file `bubblesort.v` in the public repo [https://github.com/jouvelot/mech.v](url) to give me some advice about this?
Thanks.

proux01:
Hi @jouvelot , sorry for the late answer. I finally got a look. IINM you want to prove that every permutation of a finite set is the composition of finitely many transpositions. I don't know that part of MathComp very well but it seems to me that this result is already proved as `prod_tpermP` in `perm.v`. Why don't you use that?

jouvelot:
No problem, @proux01. One limitation of `prod_termP` I see comes from the fact that the permutations in `perm.v` for a sequence `s` deal with the values in `s`, while I need to deal with both the values and their indices. The theorem I suggest sees transpositions as pairs of indices. The motivation for this is that one can then limit oneself, to sort upward `s`, to applying transpositions that all have the property of being non-sorted bubbles (i.e., if `i1 < i2`, then `nth s i2 < nth s i1`), which is what bubble sort uses. In addition of providing a specification for a kind of bubble sort algorithm, this theorem is a key ingredient to prove some lemmas in `mech.v`, which is a framework under development to specify auction mechanisms (in the Mechanism Design sense).

proux01:
Interesting. I wonder whether you could only go with indices (as values can be retrieved with `nth`). If it were me, I would probably have a detailed look at the proof of `prod_tpermP` and see if I can mimick it for my (stronger) property.
Anyway, to answer the initial question, "duplicating" the definition of transposition and redoing on nat things that were (for a small part) already there on ordinals seems a bit wasteful and I wouldn't consider this ready for a pull request (at least in its current state).

jouvelot:
I agree with your assessment. 

- I thought using `nat` instead of ordinals would be more useful, since it would have more impact than limiting it to finite types (which is ironic, since I use mostly ordinals and tuples in `mech.v`; in fact, I had to add a layer to map the `nat`-based version of my lemma to ordinals, so that I could use it in `mech.v`). 
- I also thought about using "mathcomp indices" instead of explicit ones, but they can sometimes be a bit difficult to use when duplicates exist in sequences (I use it nonetheless in the proof, since this issue is not a factor there). 
- The duplications you mention are indeed unfortunate, although most of the `max`-based lemmas are not present in `bigop.v` (and may not be applicable to `nat`, anyhow, if they were there). Some of them could in fact be added to `bigop.v`, if deemed useful.
- For now, I'll have, as you suggest, a detailed look at `prod_termP`. 

I managed to get my "bubble lemma" to work as expected in my project. In my mind, the question remains of whether getting a nicer-looking version of it would be of value to the mathcomp community, or whether I should limit myself to a finite-type-only version ... or just drop it.

PR #840, opened 2022-01-13
Added theorem pairwise_trans

CohenCyril:

If a list `s` is pairwise for an antisymmetric relation, then this
relation is transitive in `s`.

(If the relation is not antisymmetric, the relation is not necessarily
transitive, e.g. pose `<=` such that `pairwise <= [:: x; y; z]`,
`z <= y <= x`, and not `z <= x`).

##### Things done/to do


proux01 (approved):
I'm not an absolute specialist but this looks perfectly reasonable to me.
So I'll merge by the end of the week if no further comments.

CohenCyril:
I think it's ok to merge right away ;)

PR #869, opened 2022-04-05
pattern for rewriting Boolean (in)equalities

affeldt-aist:

These patterns were introduced by @CohenCyril in mathcomp-analysis and proved useful @thery .
They do not seem to be that useful in mathcomp itself but they could benefit to other users.

##### Things done/to do

### DIFF
--- a/mathcomp/ssreflect/eqtype.v
+++ b/mathcomp/ssreflect/eqtype.v
@@ -38,6 +38,8 @@ From mathcomp Require Import ssreflect ssrfun ssrbool.
 (*         x != y :> T <=> x and y compare unequal at type T.                 *)
 (*             x =P y  :: a proof of reflect (x = y) (x == y); x =P y coerces *)
 (*                     to x == y -> x = y.                                    *)
+(*               eqLHS := (X in (X == _))%pattern (for rewriting)             *)
+(*               eqRHS := (X in (_ == X))%pattern (for rewriting)             *)
 (*               eq_op == the boolean relation behind the == notation.        *)
 (*             pred1 a == the singleton predicate [pred x | x == a].          *)
 (* pred2, pred3, pred4 == pair, triple, quad predicates.                      *)
@@ -188,6 +190,9 @@ Notation "x =P y" := (eqP : reflect (x = y) (x == y))
   (at level 70, no associativity) : eq_scope.
 Notation "x =P y :> T" := (eqP : reflect (x = y :> T) (x == y :> T))
   (at level 70, y at next level, no associativity) : eq_scope.
+
+Notation eqLHS := (X in (X == _))%pattern.
+Notation eqRHS := (X in (_ == X))%pattern.
 
 Prenex Implicits eq_op eqP.
 
--- a/mathcomp/ssreflect/eqtype.v
+++ b/mathcomp/ssreflect/eqtype.v
@@ -38,8 +38,8 @@ From mathcomp Require Import ssreflect ssrfun ssrbool.
 (*         x != y :> T <=> x and y compare unequal at type T.                 *)
 (*             x =P y  :: a proof of reflect (x = y) (x == y); x =P y coerces *)
 (*                     to x == y -> x = y.                                    *)
-(*               eqLHS := (X in (X == _))%pattern (for rewriting)             *)
-(*               eqRHS := (X in (_ == X))%pattern (for rewriting)             *)
+(*              eqbLHS := (X in (X == _))%pattern (for rewriting)             *)
+(*              eqbRHS := (X in (_ == X))%pattern (for rewriting)             *)
 (*               eq_op == the boolean relation behind the == notation.        *)
 (*             pred1 a == the singleton predicate [pred x | x == a].          *)
 (* pred2, pred3, pred4 == pair, triple, quad predicates.                      *)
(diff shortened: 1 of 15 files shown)


ybertot:
LGTM, but

- It feels that users will rarely use these, because it takes too much remembering that these pattern exist.
- This suggests that the "RHS" notation should be usable to denote the right-hand side of equalities and inequalities alike.  How would that be possible?  It seems that one would need to be able to define "multi-patterns" (disjunctions of patterns), in the same way that we already define multi-rules.  On the surface, it seems an easy thing to define, but digging more deeply, it feels that providing this together with extensibility (as in extending multi-rules) is quite hard actually.

ybertot:
Another issue I have with this PR is that it corrects some spacing problems which are independent from the subject in the title.  It is too bad that these corrections will be forgotten if this PR is not incorporated.

ybertot (approved):

CohenCyril:
> * This suggests that the "RHS" notation should be usable to denote the right-hand side of equalities and inequalities alike.  How would that be possible?

I tried at some point and failed: this was too ambiguous, e.g. for which `r` would `RHS` match `y` in `r x y`? 
Even if we narrow `r` down to a few symbols (e.g. through some for of table, e.g. canonical projections).
Many goals containing `_ <= _` also contain `_ = _`, if we are not more precise, how do we get it.

Like many things in ssr most people will not use it until they really feel the need, and that's when they must be able find them through documentation. Maybe we should rewrite a mathcomp local addentum to https://coq.inria.fr/refman/proof-engine/ssreflect-proof-language.html for that purpose?

affeldt-aist:
The topic was discussed during the last mathcomp-dev meeting, see https://github.com/math-comp/math-comp/wiki/Minutes-April-06-2022.
After the meeting @gares figured out a way based on @CohenCyril 's idea:
```
Notation LEFT symb := (X in symb X _)%pattern.
```
with the following comment: "one still needs to put in %pattern scope some notations for the usual infix symbols."

Should it come as a replacement or an addition?

CohenCyril:
> Should it come as a replacement or an addition?

I would vote for "as an addition"

gares:
I guess one needs to make the scope explicit around symb as well. And test it a little, since I did not do it very carefully.

thery:
a pity we can't write with `[LEFT <=]` and `[LEFT =]`

mathcomp/ssreflect/eqtype.v
```diff
@@ -38,6 +38,8 @@ From mathcomp Require Import ssreflect ssrfun ssrbool.
 (*         x != y :> T <=> x and y compare unequal at type T.                 *)
 (*             x =P y  :: a proof of reflect (x = y) (x == y); x =P y coerces *)
 (*                     to x == y -> x = y.                                    *)
+(*               eqLHS := (X in (X == _))%pattern (for rewriting)             *)
```
thery, line 41: should it not be `eqbLHS`?

gares:
I think you can if you add these notations in the pattern scope (not as infix, just as atoms).

thery:
I am pretty bad at notation I tried to add
`Notation "LEFT <=" := (X in X <= _)%pattern (at level 10).`
and it broke everything :see_no_evil:

gares:
I'm not on my pc
```
Notation "<=" := leq : symb_scope.
Notation "LEFT f" := (X in f%symb X _)%pattern.
```

mathcomp/ssreflect/eqtype.v
affeldt-aist: for the sake of consistency indeed, but `eqLHS` is free and shorter; if we enforce consistency to have `eqbLHS` then we could think of using `eqLHS` instead of `LHS` and that does not seem a good idea given than it is used very often

affeldt-aist:
> I would vote for "as an addition"

Then we can maybe merge this one for now and issue for the more general pattern (it might require some testing).

thery:
I have tried
```coq
Notation "<=" := leq : symb_scope.
Delimit Scope symb_scope with symb.
Notation "'LEFT' f" := (X in f%symb X _)%pattern (at level 10).
Goal 1 <= 2 .```
and I get a
`Unknown interpretation for notation "<=".`

File: mathcomp/ssreflect/eqtype.v (hunk 1 above)
Comment (thery) [reply mathcomp/ssreflect/eqtype.v:41]: my poor brain prefers consistency ;-)

Comment (thery) [reply mathcomp/ssreflect/eqtype.v:41]: I would prefer `eqLHS` to be a synonymous of `LHS` and is there a need for `leqLHS` and `ltnLHS` or is this captured by `leLHS` and `ltLHS`?

**Comment by CohenCyril** [comment]
I think we need to create a new nonterminal (a.k.a. custom entry) for that otherwise we will screw up the main parser.

**Comment by thery** [comment]
```coq
Delimit Scope symb_scope with symb.
Declare Custom Entry symb.
Notation "<=" := leq (in custom symb at level 0) : symb_scope.
Notation "'LEFT' f" := (X in f%symb X _)%pattern (at level 10).
Goal 1 <= 2 .
rewrite -[LEFT <=](addn1).
```
now it breaks at the `<=` of the rewrite with
`Syntax error: [term] expected after '<=' (in [term]).`

mathcomp/ssreflect/eqtype.v
affeldt-aist: The last push force added leqLHS, etc. They seem quite useful. As for adding eqLHS as a synonymous of LHS, it looks like it better be an addition to Coq's ssrmatching.v. Should we PR this to Coq?

mathcomp/ssreflect/ssrnat.v
```diff
@@ -53,6 +53,12 @@ Require Export Ring.
 (*   the size of the library down. All the inequalities refer to the same     *)
 (*   constant "leq"; in particular m < n is identical to m.+1 <= n.           *)
 (*                                                                            *)
+(* -> patterns for contextual rewriting:                                      *)
+(*      leqLHS := (X in (X <= _)%N)%pattern                                   *)
+(*      leqRHS := (X in (_ <= RHS)%N)%pattern                                 *)
```
clayrat, line 58: Should this by `(X in (_ <= X)%N)`?

affeldt-aist: Indeed, thank you for noticing!

gares (approved):

PR #901, opened 2022-06-22
Extra theorems about half, uphalf and double

thery:

An extra theorem about half : ` (m <= n./2) = (m.*2 <= n)`  and  uphalf : `(uphalf m <= n) = (m <= n.*2)`

##### Things done/to do


proux01 (approved):
LGTM, will merge next week if no further comment.
Maybe add a similar lemma for hal (something like `m <= n./2 = m.*2 <= n`).

thery:
> Maybe add a similar lemma for half (something like `m <= n./2 = m.*2 <= n`).

done, thanks

PR #903, opened 2022-06-30
Lemmas relating preffix/infix/suffix and path/sorted

mgbedmar:

The predicates `prefix/infix/suffix` were just added to `seq.v` and I found myself in need of theory regarding `path/sorted`. If there are more lemmas that should be added, let me know.

##### Things done/to do


proux01 (approved):
Could you please update CHANGELOG_UNRELEASED?
Otherwise, LGTM, will merge by the end of the week (July 8th) if there is no further comment.

PR #909, opened 2022-07-25
generalize take_take and add take_takel and take_taker

thery:

More generic theorem `take_take`

##### Things done/to do

### DIFF
--- a/CHANGELOG_UNRELEASED.md
+++ b/CHANGELOG_UNRELEASED.md
@@ -11,10 +11,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Added
 
 - in `seq.v`
-  + lemma `subset_mapP`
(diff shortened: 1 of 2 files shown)


CohenCyril:
Great! But even if the CI does not seem to show a single failure, this change is not backward compatible. I'm a little bit hesitant to approve. It could be make backward compatible at the cost of naming the new (generalized) lemma slightly differently (e.g. `take_take_min`, or just `take_min` swapping the RHS and LHS of the equality).
@thery what do you think?

pi8027:
Related discussion: https://github.com/math-comp/math-comp/pull/381#discussion_r320603902

thery:
@pi8027 oops I've missed that discussion. I know it is a breaking change but don't you think it is more reasonable like this?

CohenCyril (approved):
Let's rebase and merge

PR #912, opened 2022-08-08
Add root_expn and root_deg1 to poly.v

### LINKED ISSUE #886: Suggestion of new lemmas for polynomial root computation in poly.v
Among the lemmas with the root predicate, the following is in my opinion lacking:
```
Section MissingRoot.

Variable R: idomainType.

Lemma root_expn (p: { poly R }) (n: nat) a: comm_poly p a -> (0 < n)%N -> root (p ^+ n) a = root p a.
Proof.
move=> Hp Hn.
by rewrite !rootE horner_exp_comm // expf_eq0 Hn //=.
Qed.

End MissingRoot.
```

SnarkBoojum:

Those simple lemmas should be useful.
Fixes #886 

##### Things done/to do

Fixes #886 

### DIFF
--- a/CHANGELOG_UNRELEASED.md
+++ b/CHANGELOG_UNRELEASED.md
@@ -14,7 +14,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   + lemmas `prefix_path`, `prefix_sorted`, `infix_sorted`, `suffix_sorted` 
 
 - in `poly.v`
-  + lemmas `comm_polyD`, `comm_polyM`, `comm_poly_exp`, `root_expn`, `root_deg1`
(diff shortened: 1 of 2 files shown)


proux01 (approved):
@SnarkBoojum thanks for the contribution. I simplified proofs a bit (we try to prefer patterns to index selectors `{2}` for instance as they are often more robust, but this is mostly nitpicking).
This LGTM, I'll squash and merge by the end of the week if there is no further comment.

BTW : please name your branch everything but `master` when opening a PR, this is more convenient to distinguish it from the `math-comp:master` branch for instance or if you have multiple PRs opened at the same time.

SnarkBoojum:
I like your proofs better, indeed. I'll try to name my branches in the future.

PR #936, opened 2022-10-16
Avoid reusing the name of `take_take`

CohenCyril:

Fixing [a problem declared on Zulip](https://coq.zulipchat.com/#narrow/stream/237665-math-comp-devs/topic/breaking.20change.20of.20take_take/near/304290639) introduced by #909

CC @thery @palmskog

##### Things done/to do

### DIFF
--- a/.github/workflows/nix-action-coq-8.16.yml
+++ b/.github/workflows/nix-action-coq-8.16.yml
@@ -1259,6 +1259,52 @@ jobs:
       name: Building/fetching current CI target
       run: NIXPKGS_ALLOW_UNFREE=1 nix-build --no-out-link --argstr bundle "coq-8.16"
         --argstr job "odd-order"
+  reglang:
+    needs:
+    - coq
+    - mathcomp-ssreflect
+    runs-on: ubuntu-latest
+    steps:
+    - name: Determine which commit to test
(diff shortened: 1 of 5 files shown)


CohenCyril:
Actually `take_min` might be a better choice than `take_take`... what do you think @thery?
Should we add commit 10f5a1cbb5f77020bfa3d0b3afff8387828c7116 or not?

thery:
do you want to deprecate take_take and only lieave take_taker and take_takel?

thery:
ok this is what you did :+1:

CohenCyril:
> do you want to deprecate take_take and only lieave take_taker and take_takel?

Actually I was considering deprecating nothing and keeping `take_min`, `take_take` (as it was in 1.15) and `take_taker`, since actually `take_min` seems at least as natural to me than the "new" `take_take` (from the current `master` branch)...
What do you think @thery?

thery:
I find a bit strange to have `take_taker` and not `take_takel` but  there are other places in the library where it is like this, for example  `comp_polyXr`

thery:
maybe it makes sense to have `take_takel` since there is no reason to priviledge `i<=j` to `j<=i`

thery (commented):
Fine with me

PR #964, opened 2023-02-07
change sval from notation to definition

affeldt-aist:

I noticed that the `sval` notation prints as `@sval`, exhibiting its contents.
It therefore clutters the local context and in practice users are tempted to use `projT1` instead.
Maybe it is not the original intent. `sval` may be more useful as a definition (like `s2val`).
Hence this PR.

##### Things done/to do


affeldt-aist:
NB: the `sval` theory is duplicated in `ssrfun.v`.

Issue #579, opened 2020-09-07
sugestion/discussion for a `contra:` tactic


chdoc:
Being a frequent user of `contra_` lemmas I wonder whether there would be a way to avoid having a separate lemma for every combination of (decidable) predicates (cf. #578), and instead have a tactic `contra:` such that `contra: H` looks at the goal and applies canonical/registered negations to both premise and conclusion of the resulting implication. 

Unfortunately, some predicates can be negated in several ways, e.g. `n <= m` can be negated as `m < n`, `~~ (n <= m)` or even `~(n <= m)` by applying different contra lemmas. So this alone may not solve the problem completely.

CohenCyril:
my suggestion is to make contra a generic lemma using canonical structures / typeclass attaching a negation to each construct. So that one can use `apply: contra H` in any case.

PR #1119, opened 2023-11-06
Contrib bigop

pPomCo:

Add some new lemmas and generalize big_set0 to any op/idx

##### Things done/to do

      the `mathcomp-1` branch **or** I already opened an issue or PR (please cross reference).~

### DIFF
--- a/mathcomp/ssreflect/bigop.v
+++ b/mathcomp/ssreflect/bigop.v
@@ -2000,11 +2000,8 @@ Lemma bigUI (I : finType) (P Q : pred I) F :
   (\big[*%M/1]_(i in [predU P & Q]) F i) * \big[*%M/1]_(i in [predI P & Q]) F i
   = (\big[*%M/1]_(i in P) F i) * (\big[*%M/1]_(i in Q) F i).
 Proof.
-rewrite [E in E * _ = _](bigID (fun i => i \in P)).
-rewrite [E in  (E * _) * _ = _](bigID (fun i => i \in Q)).
-rewrite -Monoid.mulmA.
-rewrite [E in _ = _ * E](bigID (fun i => i \in P)).
-rewrite [E in _ = E * _](bigID (fun i => i \in Q)).
+rewrite (bigID (mem P)) (bigID (mem Q)) -Monoid.mulmA.
+rewrite [E in _ = _ * E](bigID (mem P)) [E in _ = E * _](bigID (mem Q)).
 rewrite [E in _ = _ * E]Monoid.mulmC.
 do 2 congr (_*_); apply: eq_bigl=>i; rewrite !inE;
   by case (i \in P); case (i \in Q).
--- a/mathcomp/ssreflect/bigop.v
+++ b/mathcomp/ssreflect/bigop.v
@@ -2105,12 +2105,12 @@ Local Notation "x * y" := (op x y).
 Lemma big_eq1F (I : finType) r (P : pred I) F :
   \big[op/idx]_(i <- r | P i) F i != idx -> exists i, [&& i \in r, P i & F i != idx].
 Proof.
-move=>H; apply/existsP/negbNE/existsPn=>Hcontra.
+move=>Hneq1; apply/existsP/negbNE/existsPn=>Hcontra.
 have Hcontra2 :  forall i : I, P i && (i \in r) -> F i = idx.
 move=>i /andP[Hi1 Hi2] ; move: (Hcontra i).
 by rewrite !negb_and=>/orP// ; case=>[|/orP] ; last case ;
   rewrite ?Hi2 ?Hi1=>///negPn/eqP->.
-by rewrite (big1_seq _ Hcontra2) eqxx in H.
+by rewrite (big1_seq _ Hcontra2) eqxx in Hneq1.
 Qed.
 End Plain.
 End MonoidPropertiesWithDecEq.
--- a/mathcomp/ssreflect/bigop.v
+++ b/mathcomp/ssreflect/bigop.v
@@ -1996,6 +1996,20 @@ Lemma bigU (I : finType) (A B : pred I) F :
     (\big[*%M/1]_(i in A) F i) * (\big[*%M/1]_(i in B) F i).
 Proof. exact/bigU_idem/mul1m. Qed.
 
+Lemma bigUI (I : finType) (P Q : pred I) F :
+  (\big[*%M/1]_(i in [predU P & Q]) F i) * \big[*%M/1]_(i in [predI P & Q]) F i
+  = (\big[*%M/1]_(i in P) F i) * (\big[*%M/1]_(i in Q) F i).
+Proof.
+rewrite [E in E * _ = _](bigID (fun i => i \in P)).
+rewrite [E in  (E * _) * _ = _](bigID (fun i => i \in Q)).
+rewrite -Monoid.mulmA.
+rewrite [E in _ = _ * E](bigID (fun i => i \in P)).
+rewrite [E in _ = E * _](bigID (fun i => i \in Q)).
+rewrite [E in _ = _ * E]Monoid.mulmC.
+do 2 congr (_*_); apply: eq_bigl=>i; rewrite !inE;
+  by case (i \in P); case (i \in Q).
+Qed.
+
 Lemma partition_big I (s : seq I)
       (J : finType) (P : pred I) (p : I -> J) (Q : pred J) F :
   (forall i, P i -> Q (p i)) ->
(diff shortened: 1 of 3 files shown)


Tragicus:
Hello pPomCo. Thank you for the contribution. I will do a first review. Apart from a few specific comments (to come), the only issues I see are the spacing conventions and the statements of equalities. For the former, I think the general rule is that there are spaces around symbols like `:`, `=>`, `/=` and `//`, except when they follow immediately a tactic, as in `move=>` (also, `case` and `apply` always come with a colon (except for `apply/H`)). For the latter, if the equality is too long for one line, it seems that the rule is to put the `=` at the end of the first line. As for whether the second line should be indented, the library does not seem consistent. so I guess you can choose.

mathcomp/ssreflect/bigop.v
```diff
@@ -2399,6 +2437,20 @@ Lemma bigA_distr_bigA (I J : finType) F :
     = \big[+%M/0]_(f : {ffun I -> J}) \big[*%M/1]_i F i (f i).
 Proof. by rewrite bigA_distr_big; apply: eq_bigl => ?; apply/familyP. Qed.
 
+Lemma partition_big_distrl (I J : finType) (P : pred J)
+                           (h : I -> J) (f : I -> R) (g : J -> R) :
+    \big[+%M/0]_(A | P (h A)) (f A * g (h A))
+    = \big[+%M/0]_(B | P B) (\big[+%M/0]_(A | h A == B) f A * g B).
```
Tragicus, line 2443:
```suggestion
Lemma partition_big_distrl I (s : seq I)
      (J : finType) (P : pred I) (p : I -> J) (Q : pred J) F G :
  (forall i, P i -> Q (p i)) ->
  \big[+%M/0]_(i <- s | P i) (F i * G (p i)) =
  \big[+%M/0]_(j : J | Q j) (\big[+%M/0]_(i <- s | (P i) && (p i == j)) F i * G j).
```
This looks like a refinement of `partition_big`, so it should have a similar statement (and it is more general this way). You can also add a `partition_big_distrr`.

mathcomp/ssreflect/bigop.v
```diff
@@ -2399,6 +2437,20 @@ Lemma bigA_distr_bigA (I J : finType) F :
...
+Proof.
+under [RHS]eq_bigr do rewrite big_distrl /=.
+rewrite [LHS](partition_big h P) => //.
+apply: eq_bigr => B HB ; apply: eq_big => [A | A /andP [_ HA2]].
+- case (boolP (h A == B)) => H; last by rewrite andbF.
+  by rewrite (eqP H) HB.
+- by rewrite (eqP HA2).
```
Tragicus, line 2450:
```suggestion
under [RHS]eq_bigr do rewrite big_distrl /=.
rewrite (partition_big p Q) //.
by apply: eq_bigr => b Qb; apply: eq_big => a // /andP [_] /eqP ->.
```
With the new statement of `partition_big_distrl`,  the proof could look like this. In any case, some variables are not informative enough (e.g. H), and base objects (i.e. those that are not types and to some extent functions) should use small letters.

mathcomp/ssreflect/finset.v
```diff
@@ -1369,6 +1369,29 @@ Arguments imsetP {aT rT f D y}.
 Arguments imset2P {aT aT2 rT f2 D1 D2 y}.
 Arguments imset_disjoint {aT rT f A B}.
 
+Section BigOpsAnyOp. (* Any operator op : R -> R -> R *)
+Variables (R : Type) (x : R).
+Variables (op : R -> R -> R).
+Variables I J : finType.
```
Tragicus, line 1375: I and J could have `Implicit Type`s since they are not used in all lemmas (as you did with the following constants).

mathcomp/ssreflect/finset.v
```diff
@@ -1369,6 +1369,29 @@ Arguments imsetP {aT rT f D y}.
...
+
+Lemma big_set0 F : \big[op/x]_(i in set0) F i = x.
+Proof. by apply big_pred0=>i ; rewrite inE. Qed.
+
+Lemma big_set1E j F : \big[op/x]_(i in [set j]) F i = op (F j) x.
+Proof. by rewrite -big_pred1_eq_id; apply: eq_bigl=>i; exact: in_set1. Qed.
+
+Lemma big_condT A F :
+  \big[op/x]_(i in A | true) F i = \big[op/x]_(i in A) F i.
+Proof. by apply: eq_bigl=>i ; by rewrite andbT. Qed.
```
Tragicus, line 1389: This is already in bigop.v, so it should not need to be here too.

mathcomp/ssreflect/finset.v
```diff
@@ -1456,6 +1520,82 @@ Lemma partition_big_imset h (A : {pred I}) F :
      \big[aop/idx]_(j in h @: A) \big[aop/idx]_(i in A | h i == j) F i.
 Proof. by apply: partition_big => i Ai; apply/imsetP; exists i. Qed.
 
+Lemma big_subsetI A B (F : {set I} -> R) :
+  aop (\big[aop/idx]_(C : {set I} | (C \subset A) && ~~(C \subset B)) F C)
+      (\big[aop/idx]_(C : {set I} | C \subset A :&: B) F C)
+  = \big[aop/idx]_(C : {set I} | C \subset A) F C.
+Proof.
+rewrite [E in _=E](bigID (fun C : {set I} => C \subset B)) [E in _=E]Monoid.mulmC.
```
Tragicus, line 1528:
```suggestion
rewrite [RHS](bigID (fun C : {set I} => C \subset B)) [RHS]Monoid.mulmC.
```
golfing

mathcomp/ssreflect/finset.v
```diff
@@ -1456,6 +1520,82 @@ Lemma partition_big_imset h (A : {pred I}) F :
...
+by congr (aop _ _); apply: eq_bigl=>/=C; exact: subsetI.
+Qed.
+
+Lemma big_partitionS (f : {set I} -> I -> R) :
+  \big[aop/idx]_(A : {set I}) (\big[aop/idx]_(i in A) f A i)
+  = \big[aop/idx]_(i : I) \big[aop/idx]_(A : {set I} | i \in A) f A i.
```
Tragicus, line 1534: The name does not fit, there is no partition involved.
Besides, this should be generalized to permuting any two bigops (I am actually surprised that it is not already in the library, @ Very Informed Person?).

mathcomp/ssreflect/finset.v
```diff
@@ -1456,6 +1520,82 @@ Lemma partition_big_imset h (A : {pred I}) F :
...
+      by rewrite -Hcontra HB in HA.
+  + rewrite /f1/f2 => /= /andP [Hx Hinv].
+    move: Hinv ; rewrite /f2inv.
+    case (boolP (x\in A)) => H ; last by rewrite Hx in H.
+    by move/eqP ; case => ->.
+- rewrite /omap/obind/oapp/f2inv => x Hx.
+  by case (boolP (x \in A)) => // ; rewrite Hx.
+rewrite proof2 proof1.
+rewrite -[LHS](partition_big f1 (P:=predT) predT)=> //.
+by rewrite -[RHS](partition_big f2 (P:=predT) predT)=> //.
```
Tragicus, line 1596:
```suggestion
under eq_bigr do rewrite big_mkcond.
under [LHS]eq_bigr do rewrite big_mkcond.
do 2 rewrite pair_bigA_idem ?Monoid.op1m //.
rewrite (reindex (fun x => (x.2, x.1))) //=.
by exists (fun x => (x.2, x.1)) => -[]. 
```
Now I feel bad, because you probably spent a lot of time on this one and had some very good ideas, but it so happens that there is a trick that trivialises it.

CohenCyril: I will probably make you feel worse, but this 
`exchange_big_dep predT`

mathcomp/ssreflect/bigop.v
```diff
@@ -2077,6 +2091,30 @@ End Abelian.
...
+Variable idx : R.
+Local Notation "1" := idx.
+
+Section Plain.
+Variable op : Monoid.law 1.
+Local Notation "*%M" := op (at level 0).
+Local Notation "x * y" := (op x y).
+
+Lemma big_eq1F (I : finType) r (P : pred I) F :
+  \big[op/idx]_(i <- r | P i) F i != idx -> exists i, [&& i \in r, P i & F i != idx].
```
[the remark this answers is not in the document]
CohenCyril: Indeed, but it's a contrapositive invoving "~ forall -> exists", which is only valid constructively when the indexes range in a fintype, which is exactly the case here.
Maybe this calls for a lemma in fintype `contra_Nexists` etc. More generally, this calls for a modularization of `contar` as advocated in https://github.com/math-comp/math-comp/issues/579

mathcomp/ssreflect/finset.v
Tragicus: According to the discussion on the next comment, `big_partitionS` is probably not useful, as it is covered by `big_exchange_dep`.

mathcomp/ssreflect/finset.v
CohenCyril: Here I think it makes sense to keep them as variables (otherwise the following implicit type declaration will fail). Moreover section discharge will take care of removing unused variables.

CohenCyril:
Hi @pPomCo do you intend to follow up on this? If not that's ok, we will finish this.

pPomCo:
Hi,

I was very busy last months, that's why the PR was asleep.
I will have some time available now so I will follow this PR and I hope I can commit on it next month.
Anyway, if you want to improve it, feel free to commit any change you like.

pPomCo

CohenCyril:
@pPomCo should we take over?

CohenCyril:
@erikmd @pPomCo should we take over?

pPomCo:
Hi,
Thanks for asking! Of course, feel free to take over — I haven’t had a chance to work on it at all.

PR #1149, opened 2024-01-12
Fix typo

ju-sh:
Fix a repetition that looked like a typo.


CohenCyril (approved):
Thanks for fixing typos!

PR #1267, opened 2024-09-09
Adding gtn_sorted_uniq_geq and reflexive closure

jouvelot:

This PR intends to add `gtn_sorted_uniq_geq` to `path.v`for consistency, since there is currently only a version for `ltn`, namely `ltn_sorted_uniq_leq`. It also generalizes these lemmas to any reflexively closed relation (in fact, the section `ReflClosure` should probably be put somewhere else, e.g., `eqtype.v`, since it introduces the notion of reflexive closure and some general lemmas in fact not needed here).

If this generalization is deemed too much, I can replace this PR with a simpler one that would only prove the `gtn_sorted_uniq_geq` lemma as follows:
```
Lemma gtn_sorted_uniq_geq s : sorted gtn s = uniq s && sorted geq s.
Proof. by rewrite -rev_sorted -[in RHS]rev_sorted -rev_uniq ltn_sorted_uniq_leq. Qed.
```
Any suggestions?

##### Minimal TODO list

### DIFF
--- a/mathcomp/ssreflect/path.v
+++ b/mathcomp/ssreflect/path.v
@@ -1170,10 +1170,57 @@ Qed.
 Lemma size_sort (T : Type) (leT : rel T) s : size (sort leT s) = size s.
 Proof. exact: (count_sort _ predT). Qed.
 
+Section ReflClosure.
+
+Variable (T : eqType) (r : rel T).
+
+Definition refl_closure := [rel x y : T | (x == y) || r x y].
+
+Lemma refl_closure_reflexive : reflexive refl_closure.
+Proof. by move=> ?/=; rewrite eq_refl orTb. Qed.
+
+Lemma refl_closure_symmetric : symmetric r -> symmetric refl_closure.
+Proof.
+move=> sr x y/=.
+have [/eqP ->|] := boolP (x == y); first by rewrite eq_refl.
+by rewrite sr eq_sym => /negPf ->.
+Qed.
+
+Lemma refl_closure_trans : transitive r -> transitive refl_closure.
+Proof. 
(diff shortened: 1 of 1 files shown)


pi8027:
I think that `lt_sorted_uniq_le` in `order.v` is an equivalent of `refl_closure_sorted_uniq`.

jouvelot:
> I think that `lt_sorted_uniq_le` in `order.v` is an equivalent of `refl_closure_sorted_uniq`.

Thanks for the pointer to `order.v`. This is indeed quite similar, even though the notion of reflexive closure is not limited to order relations and is not really "first-class" there. But I guess most uses of this notion will deal with order types.

So, I suppose the other, one-line proof I mentioned above is good enough to ensure the consistency of `path.v`. I'll send a new PR for this one, then.

pi8027:
> even though the notion of reflexive closure is not limited to order relations

If you assume `r : rel T` to be transitive and irreflexive, that's a (strict) partial order. `lt_sorted_uniq_le` applies to any partially ordered type. The only difference is that the latter requires the carrier type to be a `choiceType`.

> So, I suppose the other, one-line proof I mentioned above is good enough to ensure the consistency of `path.v`.

Yes, I agree.

jouvelot:
> > ....  the notion of reflexive closure is not limited to order relations 

> If you assume `r : rel T` to be transitive and irreflexive, that's a (strict) partial order. `lt_sorted_uniq_le` applies to any partially ordered type. 

Indeed. What I meant is that the notion of reflexive closure itself (as in the `Section ReflClosure` only) works for any kind of relation.

> > So, I suppose the other, one-line proof I mentioned above is good enough to ensure the consistency of `path.v`.
> Yes, I agree.

I'll do that, then. Thanks for your help.

PR #1300, opened 2024-11-27
Generalize eqfun and eqrel to dependent functions

Tragicus:

Generalizes eqfun and eqrel to dependent functions (which will be useful when we will want and be able to put instances on dependent function types). This changes the implicit arguments of `_ =2 _` since now, in `(f =2 g) x y`, `x` appears in the type of `y`.

##### Minimal TODO list

### DIFF
--- a/CHANGELOG_UNRELEASED.md
+++ b/CHANGELOG_UNRELEASED.md
@@ -359,6 +359,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 	 `Zp_mul_addl`, `Zp_inv_out`
 	 generalized from `'I_p.+1` to `'I_p`.
 
+- in `ssrfun.v`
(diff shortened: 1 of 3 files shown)


CohenCyril:
You could add an `Arguments` directive to preserve the former implicits. I think in this particular case I'd rather put one `_` to omit a redundant argument in the dependent case rather than use `@` and put many `_` in the non-dependent case...

Tragicus:
I can not access `x` to make it explicit without changing the type of `eqrel _ _`, which would break things very hard. https://github.com/coq/coq/issues/13435 seems to imply that this issue is by design (see also [this Zulip thread](https://coq.zulipchat.com/#narrow/channel/237977-Coq-users/topic/.60Set.20Implicit.20Arguments.60.20overrides.20hidden.20arguments/near/484912705))

PR #1404, opened 2025-04-12
Realign conjC and conj_op

### LINKED ISSUE #1401: Complex conjugate is defined twice
It appears as a part of the mixin of `numClosedFieldType`:
https://github.com/math-comp/math-comp/blob/affaff8dd8275cad0518fc02e31612ce894e9ad6/algebra/ssrnum.v#L343-L352
and then, `sesquilinear.v` defines a copy of this constant:
https://github.com/math-comp/math-comp/blob/affaff8dd8275cad0518fc02e31612ce894e9ad6/algebra/sesquilinear.v#L156

CohenCyril:

Fixes #1401 

##### Overlay (to be merged before the current PR)

- https://github.com/math-comp/odd-order/pull/75

##### Minimal TODO list

### DIFF
--- a/algebra/num_theory/numfield.v
+++ b/algebra/num_theory/numfield.v
@@ -6,15 +6,11 @@ From mathcomp Require Import ssrAC div fintype path bigop order finset fingroup.
 From mathcomp Require Import ssralg poly orderedzmod numdomain.
 
 (******************************************************************************)
-(*                            Number structures                               *)
+(*                      Number structures (numfield.v)                        *)
 (*                                                                            *)
 (* NB: See CONTRIBUTING.md for an introduction to HB concepts and commands.   *)
 (*                                                                            *)
-(* This file defines some classes to manipulate number structures, i.e,       *)
-(* structures with an order and a norm. To use this file, insert              *)
-(* "Import Num.Theory." before your scripts. You can also "Import Num.Def."   *)
-(* to enjoy shorter notations (e.g., minr instead of Num.min, lerif instead   *)
-(* of Num.leif, etc.).                                                        *)
+(* NB: The header of ssrnum.v explains how to use the files in this directory.*)
 (*                                                                            *)
 (* This file defines the following number structures:                         *)
 (*                                                                            *)
@@ -31,10 +27,11 @@ From mathcomp Require Import ssralg poly orderedzmod numdomain.
 (*       Num.sqrt x == in a real-closed field, a positive square root of x if *)
 (*                     x >= 0, or 0 otherwise                                 *)
 (* For numeric algebraically closed fields we provide the generic definitions *)
-(*         'i == the imaginary number (:= sqrtC (-1))                         *)
+(*         'i == the imaginary number                                         *)
 (*      'Re z == the real component of z                                      *)
 (*      'Im z == the imaginary component of z                                 *)
-(*        z^* == the complex conjugate of z (:= conjC z)                      *)
+(*        z^* == the complex conjugate of z                                   *)
+(*            := Num.conj z                                                   *)
 (*    sqrtC z == a nonnegative square root of z, i.e., 0 <= sqrt x if 0 <= x  *)
 (*  n.-root z == more generally, for n > 0, an nth root of z, chosen with a   *)
 (*               minimal non-negative argument for n > 1 (i.e., with a        *)
@@ -58,6 +55,11 @@ From mathcomp Require Import ssralg poly orderedzmod numdomain.
 (* Pdeg2.Real : theory of the degree 2 polynomials on RealField and rcfType.  *)
 (* Pdeg2.RealMonic : theory of Pdeg2.Real specialized to monic polynomials.   *)
 (******************************************************************************)
+
+Reserved Notation "n .-root" (format "n .-root").
+Reserved Notation "'i".
+Reserved Notation "'Re z" (at level 10, z at level 8).
+Reserved Notation "'Im z" (at level 10, z at level 8).
 
 Set Implicit Arguments.
 Unset Strict Implicit.
--- a/algebra/num_theory/numfield.v
+++ b/algebra/num_theory/numfield.v
@@ -86,14 +86,19 @@ HB.export NumFieldExports.
 
 HB.mixin Record NumField_isImaginary R of NumField R := {
   imaginary : R;
-  conj_op : {rmorphism R -> R};
+  conj_subdef : {rmorphism R -> R};
   sqrCi : imaginary ^+ 2 = - 1;
-  normCK : forall x, `|x| ^+ 2 = x * conj_op x;
+  normCK_subdef : forall x, `|x| ^+ 2 = x * conj_subdef x;
 }.
 
 #[short(type="numClosedFieldType")]
 HB.structure Definition ClosedField :=
   { R of NumField_isImaginary R & GRing.ClosedField R & NumField R }.
+
+Definition conj {C : numClosedFieldType} : C -> C := @conj_subdef C.
+#[export] HB.instance Definition _ C := GRing.RMorphism.on (@conj C).
+#[deprecated(since="mathcomp 2.5.0",note="Use conj instead.")]
+Notation conj_op := conj (only parsing).
 
 Module ClosedFieldExports.
 Bind Scope ring_scope with ClosedField.sort.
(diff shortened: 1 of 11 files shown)


algebra/sesquilinear.v
```diff
@@ -99,7 +99,7 @@ From mathcomp Require Import zmodp poly order ssrnum matrix mxalgebra vector.
 (*                                                                            *)
 (*            symmetricmx := hermitianmx _ false idfun                        *)
 (*                 skewmx := hermitianmx _ true idfun                         *)
-(*              hermsymmx := hermitianmx _ false conjC                        *)
+(*              hermsymmx := hermitianmx _ false conj_op                      *)
```
CohenCyril, line 102:
```suggestion
(*              hermsymmx := hermitianmx _ false conjC                        *)
```

algebra/ssrnum.v
```diff
@@ -342,15 +342,18 @@ HB.export NumFieldExports.
 
 HB.mixin Record NumField_isImaginary R of NumField R := {
   imaginary : R;
-  conj_op : {rmorphism R -> R};
+  conj_op_subdef : {rmorphism R -> R};
   sqrCi : imaginary ^+ 2 = - 1;
-  normCK : forall x, `|x| ^+ 2 = x * conj_op x;
+  normCK : forall x, `|x| ^+ 2 = x * conj_op_subdef x;
```
pi8027, line 347:
```suggestion
  normCK_subdef : forall x, `|x| ^+ 2 = x * conj_op_subdef x;
```
and re-state it using `conj_op` below rather than using `conj_op_subdef`.

algebra/ssrnum.v
```diff
@@ -342,15 +342,18 @@ HB.export NumFieldExports.
...
 }.
 
 #[short(type="numClosedFieldType")]
 HB.structure Definition ClosedField :=
   { R of NumField_isImaginary R & GRing.ClosedField R & NumField R }.
 
+Definition conj_op {C : numClosedFieldType} : C -> C := @conj_op_subdef C.
+#[export] HB.instance Definition _ C := GRing.RMorphism.on (@conj_op C).
```
pi8027, line 355: Why isn't it `Num.conj` but `Num.conj_op`? (not from this PR)

algebra/ssrnum.v
```diff
@@ -3640,6 +3643,7 @@ Qed.
 
 End RealClosedFieldTheory.
 
+Notation conjC := conj_op.
 Notation "z ^*" := (conj_op z) : ring_scope.
 Notation "'i" := imaginary (at level 0) : ring_scope.
```
pi8027, line 3648: Isn't it better to define the abbreviation `conjC` in `Num.Def` and the two notations in `Num.Syntax`? (To do so, we need to move the definitions of some structures up, but it looks doable.)

pi8027:
@CohenCyril Do you need any help with this PR? For example, I can implement the changes I suggested.

CohenCyril:
> @CohenCyril Do you need any help with this PR? For example, I can implement the changes I suggested.

Sure, please do.

algebra/num_theory/numfield.v
```diff
@@ -144,12 +149,20 @@ End RealClosed.
...
 Notation sqrt := sqrtr.
 
+Module Import Syntax.
+
+Notation "z ^*" := (conj z) : ring_scope.
+Notation "'i" := imaginary : ring_scope.
+
+End Syntax.
+
 Module Export Theory.
```
pi8027, line 166: I found it a bit weird that this `Theory` module is exported since we should not import the `Num` module. Similarly, `Def`, `Syntax`, and `Theory` modules in `numdomain.v` are exported.

CohenCyril: I guess it does not hurt either...

pi8027:
I added the changelog entry.

algebra/num_theory/numfield.v
```diff
@@ -144,12 +149,20 @@ End RealClosed.
 
 Module Import Def.
 
+Notation conjC := conj.
```
affeldt-aist, line 152: This notation does not seem to be used in MathComp. Should it be deprecated/removed?

pi8027: No, it shouldn't.

affeldt-aist: Ok, give a bit of time to recollect why.

pi8027: Well, `conjC` is the shorthand of `Num.conj` that becomes available only when you import `Num.Def`. It's the same story as `Num.le` and `ler`, for example.

affeldt-aist: Oh right, I was not thinking straight. Thanks for the explanation. May I suggest a commit that makes it even more obvious in the documentation?

affeldt-aist: Is it also ok if I try to use this shorter notation in files such as `sesquilinear.v`? Many occurrences of `Num.conj` seem to replaceable.

algebra/num_theory/numfield.v
```diff
@@ -13,8 +13,10 @@ From mathcomp Require Import ssralg poly orderedzmod numdomain.
 (* This file defines some classes to manipulate number structures, i.e,       *)
 (* structures with an order and a norm. To use this file, insert              *)
 (* "Import Num.Theory." before your scripts. You can also "Import Num.Def."   *)
-(* to enjoy shorter notations (e.g., minr instead of Num.min, lerif instead   *)
-(* of Num.leif, etc.).                                                        *)
+(* to enjoy shorter notations, e.g.:                                          *)
+(* - minr instead of Num.min,                                                 *)
+(* - lerif instead of Num.leif,                                               *)
+(* - conjC instead of Num.conj, etc.                                          *)
```
pi8027: This part of the header doc is repeated in `numdomain.v`, `numfield.v`, and `orderedzmod.v`. Isn't it better to move it to `ssrnum.v` and refer the users to it? (I suggest doing so in a separate PR.)

algebra/num_theory/numfield.v
```diff
@@ -36,7 +38,8 @@ From mathcomp Require Import ssralg poly orderedzmod numdomain.
 (*         'i == the imaginary number (:= sqrtC (-1))                         *)
 (*      'Re z == the real component of z                                      *)
 (*      'Im z == the imaginary component of z                                 *)
-(*        z^* == the complex conjugate of z (:= conjC z)                      *)
+(*        z^* == the complex conjugate of z                                   *)
+(*            := conjC z (provided "Import Num.Def." has been executed)       *)
 (*    sqrtC z == a nonnegative square root of z, i.e., 0 <= sqrt x if 0 <= x  *)
```
pi8027, line 35: I prefer to mention longer names (`Num.conj` and `Num.sqrt`) here and document the `Num.Def` module separately. (Well, currently it's not the case.)

algebra/num_theory/numfield.v
affeldt-aist: You are right, this is too much duplication, the header part should move to `ssrnum.v`. Maybe pointers could look like this:
```
(*            := conjC z (requires "Import Num.Def.", see `ssrnumv.`)       *)
```
?

algebra/num_theory/numfield.v
affeldt-aist: I am afraid that a casual user might not understand easily that easier notations are available then.

character/classfun.v
```diff
@@ -372,7 +372,7 @@ Definition cfAut_closed (S : seq classfun) :=
 End Automorphism.
 
 (* FIX ME this has changed *)
```
affeldt-aist: This notation (and the FIX ME comment) can be removed if we import `Num.Def`.

character/classfun.v
```diff
@@ -432,7 +432,7 @@ Notation "''CF' ( G , A )" := (classfun_on G A) : ring_scope.
 Notation "1" := (@GRing.one (cfun_nzRingType _)) (only parsing) : cfun_scope.
 
 (* FIX ME this has changed *)
```
affeldt-aist: Same comment as for line 374.

algebra/num_theory/numfield.v
affeldt-aist: It will look something like that: https://github.com/affeldt-aist/math-comp/commit/c826c6429821d6cc6690bda4660f482ac9404cb7 (updated)
`sesquilinear.v` using already a lot of notations, that improves it a bit imho.

algebra/num_theory/numfield.v
pi8027: If you mention shorter names by default, you need to say `Import Num.Def` everywhere, including `sqrtC` below and a few lines in `orderedzmod.v`. IMO, that's too redundant.

affeldt-aist: Yes, I realize that, but since this `Import` invocations are FAQ I think it wouldn't hurt.
Anyway, in the interest of time and maintainability, let's go with your rationale!

affeldt-aist: (if you don't have time I can prepare a commit tomorrow)

affeldt-aist (approved):
For me, this can be merged but one comment indicates
that https://github.com/math-comp/odd-order/pull/75 should be merged first.

PR #1503, opened 2025-11-27
rename `GRing.exp` to `pown`

affeldt-aist:

as decided during the regular meeting

##### Minimal TODO list

- [x] added changelog entries ~~with `doc/changelog/make-entry.sh`~~

### DIFF
--- a/algebra/ssralg.v
+++ b/algebra/ssralg.v
@@ -6731,8 +6731,6 @@ Definition exprDn := pownDn.
 Definition pownBn := pownBn.
 #[deprecated(since="mathcomp 2.6.0", use=pownBn)]
 Definition exprBn := pownBn.
-#[deprecated(since="mathcomp 2.6.0")]
-Notation "'GRing.exprB'" := (pownBn) (only parsing).
 Definition subrXX := subrXX.
 Definition sqrrD := sqrrD.
 Definition sqrrB := sqrrB.
@@ -6800,6 +6798,8 @@ Definition exprVn := pownVn.
 Definition pownrB := pownrB.
 #[deprecated(since="mathcomp 2.6.0", use=pownrB)]
 Definition exprB := pownrB.
+#[deprecated(since="mathcomp 2.6.0")]
+Notation "'GRing.exprB'" := (pownrB) (only parsing).
 Definition invr_signM := invr_signM.
 Definition divr_signM := divr_signM.
 Definition rpred0D := @rpred0D.
--- a/doc/changelog/03-renamed/1503-expn_pown.md
+++ b/doc/changelog/03-renamed/1503-expn_pown.md
@@ -23,7 +23,7 @@
   + `exprMn` -> `pownMn`
   + `exprDn` -> `pownDn`
   + `exprDn_pchar` -> `pownDn_pchar`
-  + `exprB` -> `pownBn`
+  + `exprBn` -> `pownBn`
   + `exprZn` -> `pownZn`
   + `exprVn` -> `pownVn`
   + `exprB` -> `pownrB`


algebra/ssralg.v
```diff
@@ -911,28 +911,28 @@ Definition natr_sum := big_morph (natmul 1) natrD (mulr0n 1).
 Lemma natrM m n : (m * n)%:R = m%:R * n%:R :> R.
 Proof. by rewrite mulrnA mulr_natr. Qed.
 
-Lemma expr0 x : x ^+ 0 = 1. Proof. by []. Qed.
-Lemma expr1 x : x ^+ 1 = x. Proof. by []. Qed.
-Lemma expr2 x : x ^+ 2 = x * x. Proof. by []. Qed.
+Lemma pownr0 x : x ^+ 0 = 1. Proof. by []. Qed.
+Lemma pownr1 x : x ^+ 1 = x. Proof. by []. Qed.
+Lemma pownr2 x : x ^+ 2 = x * x. Proof. by []. Qed.
```
pi8027, line 916: I think that these lemmas should be called `powr0n`, `powr1n`, and `powr2n`, because these `0`, `1`, and `2` are `nat`. (See `mulr0n`, `mulr1n`, and `mulr2n` in `nmodule.v`.)

algebra/ssralg.v
```diff
@@ -911,28 +911,28 @@ Definition natr_sum := big_morph (natmul 1) natrD (mulr0n 1).
...
 
-Lemma exprS x n : x ^+ n.+1 = x * x ^+ n.
+Lemma pownrS x n : x ^+ n.+1 = x * x ^+ n.
```
pi8027, line 918: ~`powrSn`~ `powrS`, like `mulrS`

algebra/ssralg.v
```diff
@@ -911,28 +911,28 @@ Definition natr_sum := big_morph (natmul 1) natrD (mulr0n 1).
...
 Proof. by case: n => //; rewrite mulr1. Qed.
 
-Lemma expr0n n : 0 ^+ n = (n == 0%N)%:R :> R.
-Proof. by case: n => // n; rewrite exprS mul0r. Qed.
+Lemma pown0n n : 0 ^+ n = (n == 0%N)%:R :> R.
```
pi8027, line 921: `pow0rn`, like `mul0rn`

algebra/ssralg.v
```diff
@@ -911,28 +911,28 @@ Definition natr_sum := big_morph (natmul 1) natrD (mulr0n 1).
...
+Proof. by case: n => // n; rewrite pownrS mul0r. Qed.
 
-Lemma expr1n n : 1 ^+ n = 1 :> R.
-Proof. by elim: n => // n IHn; rewrite exprS mul1r. Qed.
+Lemma pown1n n : 1 ^+ n = 1 :> R.
```
pi8027, line 924: `pow1rn`

algebra/ssralg.v
```diff
@@ -911,28 +911,28 @@ Definition natr_sum := big_morph (natmul 1) natrD (mulr0n 1).
...
+Proof. by elim: n => // n IHn; rewrite pownrS mul1r. Qed.
 
-Lemma exprD x m n : x ^+ (m + n) = x ^+ m * x ^+ n.
-Proof. by elim: m => [|m IHm]; rewrite ?mul1r // !exprS -mulrA -IHm. Qed.
+Lemma pownrD x m n : x ^+ (m + n) = x ^+ m * x ^+ n.
```
pi8027, line 927: `powrDn`? But I'm not so sure. (The additive counterpart is `mulrnDr`, but it makes sense only because `natmul` is both left- and right-distributive.)

algebra/ssralg.v
```diff
@@ -911,28 +911,28 @@ Definition natr_sum := big_morph (natmul 1) natrD (mulr0n 1).
...
+Proof. by elim: m => [|m IHm]; rewrite ?mul1r // !pownrS -mulrA -IHm. Qed.
 
-Lemma exprSr x n : x ^+ n.+1 = x ^+ n * x.
-Proof. by rewrite -addn1 exprD expr1. Qed.
+Lemma pownrSr x n : x ^+ n.+1 = x ^+ n * x.
```
pi8027, line 930: `powrSr`, like `mulrSr`?

algebra/ssralg.v
```diff
@@ -971,43 +971,43 @@ Lemma commr_nat x n : comm x n%:R. Proof. exact/commrMn/commr1. Qed.
...
-Lemma exprMn_comm x y n : comm x y -> (x * y) ^+ n = x ^+ n * y ^+ n.
+Lemma pownMn_comm x y n : comm x y -> (x * y) ^+ n = x ^+ n * y ^+ n.
 Proof.
 move=> com_xy; elim: n => /= [|n IHn]; first by rewrite mulr1.
-by rewrite !exprS IHn !mulrA; congr (_ * _); rewrite -!mulrA -commrX.
+by rewrite !pownrS IHn !mulrA; congr (_ * _); rewrite -!mulrA -commrX.
 Qed.
 
-Lemma exprMn_n x m n : (x *+ m) ^+ n = x ^+ n *+ (m ^ n) :> R.
+Lemma pownMn_n x m n : (x *+ m) ^+ n = x ^+ n *+ (m ^ n) :> R.
```
pi8027, line 983: `powrMn_n`?

algebra/ssralg.v
```diff
@@ -971,43 +971,43 @@ Lemma commr_nat x n : comm x n%:R. Proof. exact/commrMn/commr1. Qed.
...
 Proof.
 elim: n => [|n IHn]; first by rewrite mulr1n.
-by rewrite exprS IHn mulrnAl mulrnAr -mulrnA exprS expnSr.
+by rewrite pownrS IHn mulrnAl mulrnAr -mulrnA pownrS expnSr.
 Qed.
 
-Lemma exprM x m n : x ^+ (m * n) = x ^+ m ^+ n.
+Lemma pownrM x m n : x ^+ (m * n) = x ^+ m ^+ n.
```
pi8027, line 989: `powrMn`?

pi8027 (changes requested):
While I'm not providing an exhaustive review, I think many lemmas are poorly named. `pownr` should be changed to `powrn` because of the order of the parameters of `pown`.

What about `expn` in `ssrnat.v` and `exprz` in `intmul.v`? (The latter is indeed not a power function, but do we really want `GRing.pown` to be a "power function" while we call `exprz` "exponentiation"?)

algebra/ssralg.v
pi8027: Also, the names `pown0n` and `pown1n` (below) look ambiguous because they do not contain `r`; that is, these names do not indicate that these lemmas are for ring-like structures.

affeldt-aist:
> What do you think?

I'll address all your comments but I am a bit skeptical about enlarging the scope of this PR to other naming issues because it could take a long time to complete.

pi8027:
@CohenCyril Right. I didn't see the issues before reading the patch.

@affeldt-aist Just in case, let me write down the naming convention I suggest here, so that we can discuss it further at some point. This is based on my understanding of the naming convention for the lemmas about `GRing.natmul` and `intmul`.

As a rule of thumb, a lemma whose conclusion has the form `?r ^+ ?n = _` should be called `exp<r><n>` (or `pow<r><n>` if we decide to rename `GRing.exp`), where:
- `<r>` is the abbreviation corresponding to `?r`. For example:
  + `<r>` is `r` if `?r` is a variable,
  + `<r>` is `1r` if `?r` is `1%R`,
  + `<r>` is `Mr` if `?r` is `(_ * _)%R`, and
  + `<r>` is `rMn` if `?r` is `(_ *+ _)%R`.
- `<n>` is the abbreviation corresponding to `?n`. For example:
  + `<n>` is `n` if `?n` is a variable,
  + `<n>` is `1n` if `?n` is `1%N`,
  + `<n>` is `S` if `?n` is `S _`, and
  + `<n>` is `Dn` if `?n` is `(_ + _)%N`.

Note that we shouldn't use very short abbreviations like `M` and `D` (instead of `Mr` and `Dn` above), because these `r` and `n` represent the types of `?r` and `?n` to disambiguate with `exp` of other types, e.g., `nat -> nat -> nat`, `R -> int -> R`, and `R -> R -> R`. For example, `expMn : ((_ * _) ^+ _)%R = _` is too ambiguous. When we need a separator between `<r>` and `<n>`, we put `_`, e.g., `exprMn_n`. On the other hand, when we have `GRing.exp` as a subterm of LHS like `?r` and `?n` above, its abbreviation is `rXn` or `Xn`.

As a result, the names of lemmas for `exp` (`_ ^ _`) of type `R -> R -> R` may need further disambiguation with `GRing.exp`. For example, a lemma whose LHS has the form `_ ^ (_ *+ _)` should probably be called `expr_rMn` rather than `exprMn` (which is already taken by `ssralg.v`).

Although it obviously doesn't cover everything, I hope it conveys my idea.

affeldt-aist:
Restarted as PR to https://github.com/math-comp/math-comp/pull/1544

PR #1504, opened 2025-11-28
Split ssralg.v

pi8027:
##### Motivation for this change

Following the [discussion](https://github.com/math-comp/math-comp/wiki/Minutes-2025-11-12#splitting-ssralgv-1275) during a MathComp meeting, I propose splitting `ssralg.v` into the following files:
- `alg_theory/algebra.v` contains the (semi)ring, (semi)module, and (semi)algebra structures below `unitRingType`.
- `alg_theory/divalg.v` contains the structures with division (`unitRingType` and above, but below `decFieldType`).
- `alg_theory/decfield.v` contains the reflection of the first-order theory of `unitRingType`, `decFieldType`, and `closedFieldType`.
- `alg_theory/ssralg.v` re-exports the contents of the above files, and contains most of the deprecation abbreviations from old `ssralg.v`. This file should act as the compatibility layer.

**I'm still looking for a better name for `alg_theory/algebra.v`,** but `ring.v` is currently taken by Algebra Tactics. I think it makes sense to rename `ring.v` of Algebra Tactics to, say, `ring_tactics.v`. However, it's going to take a while because we need to go through the following steps:
1. renaming `ring.v` to `ring_tactics.v` and deprecating the former in Algebra Tactics,
2. removing the deprecated `ring.v` in Algebra Tactics (after waiting for at least a few months), and
3. renaming `alg_theory/algebra.v` to `alg_theory/ring.v` in MathComp.

I think we should eventually move `ring_quotient.v`, `countalg.v`, and `finalg.v` to `alg_theory` too. The contents of `alg_theory/*.v` should eventually be re-exported from `alg_theory/all.v` (cf. #1505), and then, `alg_theory/ssralg.v` should probably be deprecated and removed.

This change will probably enable us to:
- add more structures with division: `unitSemiRingType`, Euclidean domains, etc., and
- reorder the declarations in the same way as `order.v`: all the structures first, theory, factories, then instances.

Closes #1275.

##### Dependency

- #1506

##### Minimal TODO list

- [x] added changelog entries with `doc/changelog/make-entry.sh`

- [x] added corresponding documentation in the headers
- [x] tried to abide by the [contribution guide](https://github.com/math-comp/math-comp/blob/master/CONTRIBUTING.md)
- [ ] this PR contains an optimum number of meaningful commits

See [this Checklist](https://github.com/math-comp/math-comp/wiki/Checklist-for-creating-and-review-PRs) for details.

##### Automatic note to reviewers

[...]

PR #1338, opened 2025-02-05
[Draft] Add extended real numbers

CohenCyril:
~Depends on https://github.com/math-comp/math-comp/pull/1520~ (merged)

##### Minimal TODO list

- [ ] added changelog entries with `doc/changelog/make-entry.sh`

- [ ] SemiPseudoNormed (c.f. last commit)
- [ ] adapt mathcomp analysis, note that `lee_dsum_nneg_subfset`, `lee_dsum_npos_subfset`, `lee_sum_nneg_subfset`, `lee_sum_npos_subfset` cannot be ported here.
- [ ] backport `numext` lemmas about Z-modules  to `zmodext`,
- [ ] added corresponding documentation in the headers
- [ ] tried to abide by the [contribution guide](https://github.com/math-comp/math-comp/blob/master/CONTRIBUTING.md)
- [ ] this PR contains an optimum number of meaningful commits

See [this Checklist](https://github.com/math-comp/math-comp/wiki/Checklist-for-creating-and-review-PRs) for details.

##### Automatic note to reviewers

Read [this Checklist](https://github.com/math-comp/math-comp/wiki/Checklist-for-creating-and-review-PRs).

PR #1506, opened 2025-11-30
Reduce and reorder `Require Import`s in `algebra/`

pi8027:

This change removes some unnecessary libraries from `Require Import`s in `algebra/`. Most notably, the compilation of `fraction.v` does not need to wait for the compilation of `countalg.v`, `poly.v`, and `polydiv.v`.

Similarly, I found that `orderedzmod.v` does not depend on `poly.v`, but I refrained from performing this change in this PR to avoid interfering with #1338. Furthermore, only a very small part of `ssrnum.v` depends on `poly.v`: `real_closed_axiom` and three lemmas. It would be interesting to see if we can compile `alg_theory/` (#1504) and `num_theory/` in parallel by eliminating (parallelizing) the dependency path `numdomain.v` -> `poly.v` -> `countalg.v` -> `decfield.v` -> `divalg.v` -> `algebra.v`. However, we probably won't benefit from this idea when compiling the entire MathComp with `make -j`, because compiling `order.v` takes a long time, unless we split it into smaller pieces (#1508).

##### Minimal TODO list

- ~[ ] added changelog entries with `doc/changelog/make-entry.sh`~

### DIFF
--- a/algebra/countalg.v
+++ b/algebra/countalg.v
@@ -2,7 +2,7 @@
 (* Distributed under the terms of CeCILL-B.                                  *)
 From HB Require Import structures.
 From mathcomp Require Import ssreflect ssrfun ssrbool eqtype ssrnat seq choice.
-From mathcomp Require Import fintype bigop ssralg.
+From mathcomp Require Import bigop ssralg.
 
 (*****************************************************************************)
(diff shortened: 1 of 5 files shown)


CohenCyril:
I think  we will merge #1504 direclty

PR #1514, opened 2025-12-28
Added lemma uniq_map_indexE and uniq_map_in_inj to seq

hivert:

Add some missing lemma about unicity and map in seq.v
Should be ready for review now.

##### Minimal TODO list

### DIFF
--- a/boot/seq.v
+++ b/boot/seq.v
@@ -2848,36 +2848,29 @@ case/subseqP=> m sz_m ->; apply/subseqP.
 by exists m; rewrite ?size_map ?map_mask.
 Qed.
 
-Lemma nth_index_map s x0 x :
-  {in s &, injective f} -> x \in s -> nth x0 s (index (f x) (map f s)) = x.
-Proof.
-elim: s => //= y s IHs inj_f s_x; rewrite (inj_in_eq inj_f) ?mem_head //.
-move: s_x; rewrite inE; have [-> // | _] := eqVneq; apply: IHs.
-by apply: sub_in2 inj_f => z; apply: predU1r.
-Qed.
-
 Lemma index_map_in s x :
   {in s &, injective f} -> x \in s -> index (f x) (map f s) = index x s.
 Proof.
-elim: s => // y s IHs inj_f s_x /=.
-have -> : (f y == f x) = (y == x).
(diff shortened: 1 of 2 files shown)


boot/seq.v
```diff
@@ -2856,6 +2856,20 @@ move: s_x; rewrite inE; have [-> // | _] := eqVneq; apply: IHs.
 by apply: sub_in2 inj_f => z; apply: predU1r.
 Qed.
 
+Lemma uniq_map_indexE s :
```
proux01: Maybe
```suggestion
Lemma index_map_uniq s :
```
since it's a variant of `index_map`? (also, it's not really Eliminating `index`)
That being said, I wonder if it wouldn't be better to first prove the stronger `index_map_in : {in s &, injective f} -> index (f x) (map f s) = index x s` and deduce the current lemma from it?

boot/seq.v
```diff
@@ -2856,6 +2856,20 @@ move: s_x; rewrite inE; have [-> // | _] := eqVneq; apply: IHs.
...
+  uniq (map f s) -> forall x, x \in s -> index (f x) (map f s) = index x s.
+Proof.
+move=> f_uniq x /[dup] xins /map_f fxinfs; apply/eqP.
+rewrite -(nth_uniq (f x) _ _ f_uniq) ?index_mem ?size_map ?index_mem //.
+by rewrite nth_index // (nth_map x) ?index_mem // nth_index.
+Qed.
+
+Lemma uniq_map_in_inj s : uniq (map f s) -> {in s &, injective f}.
```
proux01:
```suggestion
Lemma uniq_map_inj_in s : uniq (map f s) -> {in s &, injective f}.
```
(that convention seems a bit more common)

boot/seq.v
hivert: The problem is that bu doing so, I don't now anymore how to prove 
```
Lemma uniq_map_inj_in s : uniq (map f s) -> {in s &, injective f}.
```

hivert: I decided to reprove it anyway and have your two suggested lemmas.

hivert:
I just added two more missing lemmas `take_mkseq` and `drop_mkseq`.

PR #1515, opened 2026-01-01
lemmas from MathComp-Analysis' unstable.v

affeldt-aist:

three lemmas that have been for a while in MathComp-Analysis' `unstable.v`

##### Minimal TODO list

### DIFF
--- a/algebra/archimedean.v
+++ b/algebra/archimedean.v
@@ -856,13 +856,6 @@ Proof. exact: real_ceil_le0. Qed.
 Lemma ceil_floor x : ceil x = floor x + (x \isn't a int_num).
 Proof. exact: real_ceil_floor. Qed.
 
-Lemma abs_ceil_ge x : `|x| <= `|ceil x|.+1%:R.
-Proof.
-rewrite -natr1 natr_absz; have [x0|x0] := ltP 0 x.
-- by rewrite !gtr0_norm ?ceil_gt0// ler_wpDr// ceil_ge.
-- by rewrite !ler0_norm ?ceil_le0// ceilNfloor opprK intrD1 ltW// floorD1_gt.
-Qed.
-
 End ArchiRealDomainTheory.
 
 #[deprecated(since="mathcomp 2.4.0", note="Renamed to floor_le.")]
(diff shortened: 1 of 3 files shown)


algebra/archimedean.v
```diff
@@ -856,6 +856,13 @@ Proof. exact: real_ceil_le0. Qed.
 Lemma ceil_floor x : ceil x = floor x + (x \isn't a int_num).
 Proof. exact: real_ceil_floor. Qed.
 
+Lemma abs_ceil_ge x : `|x| <= `|ceil x|.+1%:R.
```
pi8027: You should first see if `Num.bound x` replaces `(absz (ceil x)).+1` in MCA or not. https://github.com/math-comp/math-comp/pull/1510#issuecomment-3611733949

pi8027: (They are not the same, but `Num.bound` is likely what you need when you need a sort of "upper bound" of type `nat`.)

pi8027: I demonstrated in math-comp/analysis#1818 that this lemma is unnecessary, although I didn't use `Num.bound` in the end.

algebra/ssrint.v
```diff
@@ -657,13 +657,17 @@ Lemma intrN n : (- n)%:~R = - n%:~R :> R. Proof. exact: mulrNz. Qed.
 
 Lemma intrD m n : (m + n)%:~R = m%:~R + n%:~R :> R. Proof. exact: mulrzDr. Qed.
 
+Lemma intrD1 m : m%:~R + 1 = (m + 1)%:~R :> R. Proof. by rewrite intrD. Qed.
+
+Lemma intr1D m : 1 + m%:~R = (1 + m)%:~R :> R. Proof. by rewrite intrD. Qed.
```
pi8027: To be consistent with `intrD` above, the LHS and RHS of these lemmas should be swapped IMO.

algebra/archimedean.v
CohenCyril: @pi8027 it is not because a proof can be shorter by not going through this lemma that it is unnecessary. Users expect to find it, so let's keep it.

pi8027: @CohenCyril For me, the statement of this lemma looks too specific. For example, the following statement works for any `archiNumDomainType` and also slightly tightens the bound:
```coq
forall x, `|x| <= (ceil `|x|)%:~R
```

pi8027 (approved):

pi8027:
We can still consider adding `abs_ceil_ge` or its modified version(s), but the discussion can be done in a MathComp meeting and/or another PR/issue.

PR #1513, opened 2025-12-26
define joins of POrderedZmodule and (Semi)NormedZmodule

zhou31416:
##### Motivation for this change

Define the joins between POrderedZmodule and SemiNormedZmodule/NormedZmodule (in algebra/num_theory/numdomain.v).

This allows for the subsequent development of mathematical theories related to vector orders.

##### Minimal TODO list

- [ ] added changelog entries with `doc/changelog/make-entry.sh`

- [ ] added corresponding documentation in the headers
- [ ] tried to abide by the [contribution guide](https://github.com/math-comp/math-comp/blob/master/CONTRIBUTING.md)
- [ ] this PR contains an optimum number of meaningful commits

See [this Checklist](https://github.com/math-comp/math-comp/wiki/Checklist-for-creating-and-review-PRs) for details.

##### Automatic note to reviewers

Read [this Checklist](https://github.com/math-comp/math-comp/wiki/Checklist-for-creating-and-review-PRs).

PR #1520, opened 2026-01-13
Add numZmodType

proux01:
~TODO: ensure https://github.com/math-comp/math-comp/pull/1513#issuecomment-3721752811 is addressed~ (done)

##### Minimal TODO list

### DIFF
--- a/algebra/num_theory/numdomain.v
+++ b/algebra/num_theory/numdomain.v
@@ -264,19 +264,6 @@ Qed.
 HB.instance Definition _ := GRing.isDivClosed.Build R nneg_num_pred
   nneg_divr_closed.
 
-Fact real_addr_closed : addr_closed (@real R).
-Proof.
-split=> [|x y Rx Ry]; first by rewrite realE lexx.
-without loss{Rx} x_ge0: x y Ry / 0 <= x.
-  case/orP: Rx => [? | x_le0]; first exact.
-  by rewrite -rpredN opprD; apply; rewrite ?rpredN ?oppr_ge0.
-case/orP: Ry => [y_ge0 | y_le0]; first by rewrite realE -nnegrE rpredD.
-by rewrite realE -[y]opprK orbC -oppr_ge0 opprB !subr_ge0 ger_leVge ?oppr_ge0.
-Qed.
-#[export]
-HB.instance Definition _ := GRing.isAddClosed.Build R real_num_pred
-  real_addr_closed.
-
 Fact real_divr_closed : divr_closed (@real R).
 Proof.
 split=> [|x y Rx Ry]; first by rewrite realE ler01.
@@ -396,10 +383,6 @@ rewrite sqrf_eq1 => /orP[/eqP //|]; rewrite -ger0_def le0r oppr_eq0 oner_eq0.
 by move/(addr_gt0 ltr01); rewrite subrr ltxx.
 Qed.
 
-Lemma sum_real I (P : pred I) (F : I -> R) (s : seq I) :
-  {in P, forall i, F i \is real} -> \sum_(i <- s | P i) F i \is real.
-Proof. by apply/big_real; [apply: rpredD | apply: rpred0]. Qed.
-
 Lemma prod_real I (P : pred I) (F : I -> R) (s : seq I) :
   {in P, forall i, F i \is real} -> \prod_(i <- s | P i) F i \is real.
 Proof. by apply/big_real; [apply: rpredM | apply: rpred1]. Qed.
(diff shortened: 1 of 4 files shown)


algebra/num_theory/numdomain.v
```diff
@@ -52,7 +52,16 @@ Import orderedzmod.Num.
...
 
-HB.mixin Record Zmodule_isSemiNormed (R : POrderedZmodule.type) M
+(* 
+HB.mixin Record Zmodule_isSemiPseudoNormed (R : POrderedZmodule.type) M
+         of GRing.Zmodule M := {
+  enorm : M -> \bar R;
+  ler_normD : forall x y, norm (x + y) <= norm x + norm y;
+  normrMn : forall x n, norm (x *+ n) = norm x *+ n;
+  normrN : forall x, norm (- x) = norm x;
+}. *)
```
proux01: This should be handled (or discarded / issue opened) before merging.

algebra/num_theory/orderedzmod.v
```diff
@@ -54,9 +65,66 @@ Fact ring_display : Order.disp_t. Proof. exact. Qed.
...
+HB.mixin Record Add_isHomo R of POrderNmodule R := {
+  ler_wD2l : forall x : R, {homo +%R x : y z / y <= z}
+}.
+
+(* TODO provide the positive cone definition of porderZmodType *)
+(* HB.factory Record ZmodulePositiveCone_isPOrdered of GRing.Zmodule R *)
+(*    & Order.isPOrder ring_display R := { *)
+(*   (* TODO: is posnum the right name? *) *)
+(*   posnum : {pred R}; *)
+(*   posnum0 :  *)
```
proux01: This should be handled (or discarded / issue opened) before merging.

algebra/num_theory/numdomain.v
proux01: Deferred to https://github.com/math-comp/math-comp/pull/1338

algebra/num_theory/orderedzmod.v
```diff
@@ -206,7 +274,818 @@ Export Order.PreOCoercions.
...
+Lemma realBC x y : (x - y \is Num.real) = (y - x \is Num.real).
+Proof. exact: rpredBC. Qed.
+
+Lemma realD : {in Num.real &, forall x y, x + y \is Num.real}.
+Proof. exact: rpredD. Qed.
+
+(* TODO: should work for semiring *)
+(* Lemma real1 : 1%R \is @real R. Proof. exact: rpred1. Qed. *)
+(* Lemma realn n : n%:R \is @real R. Proof. exact: rpred_nat. Qed. *)
+(* #[local] Hint Resolve real0 real1 : core. *)
```
proux01, line 880: Maybe we should open an issue about that to remember to handle it in a future PR?

algebra/num_theory/orderedzmod.v
```diff
@@ -54,9 +65,103 @@ Fact ring_display : Order.disp_t. Proof. exact. Qed.
...
+
+Section ZmodulePositiveCone.
+Variable R : zmodType.
+Variable posnum : {pred R}.
+Hypothesis posnum0 : posnum 0.
+Hypothesis posnumD : forall x y, posnum x -> posnum y -> posnum (x + y).
+Hypothesis posnum_definite : forall x, posnum x -> posnum (- x) -> x = 0.
+Local Notation type := (type posnum0 posnumD posnum_definite).
+
+Definition le x y := posnum (y - x).
```
proux01: @CohenCyril is it what you had in mind when you wrote
```coq
(* TODO provide the positive cone definition of porderZmodType *)
(* HB.factory Record ZmodulePositiveCone_isPOrdered of GRing.Zmodule R *)
(*    & Order.isPOrder ring_display R := { *)
(*   (* TODO: is posnum the right name? *) *)
(*   posnum : {pred R}; *)                                                                                                                                                                
(*   posnum0 :  *)
```
If yes, we should document it in the header and this is ready.
Otherwise we should probably postpone this point to a further PR.

CohenCyril: Yes! Except I screwed up the name since it should be called nonneg, I envisioned it as a factory since it is an equivalent presentation.

proux01: Apparently, the term "positive cone" also exists: https://en.wikipedia.org/wiki/Partially_ordered_group but I agree that nonneg would be better (at least to replace `posnum`).
I did not implement it as a factory because of non-forgetful inheritance (the <= being built out of posnum and subtraction). Maybe a better solution would be to add an axiom `le_sub_def : forall x y, (x <= y) = posnum (y - x)`or something like that.

CohenCyril: > Apparently, the term "positive cone" also exists: https://en.wikipedia.org/wiki/Partially_ordered_group but I agree that nonneg would be better (at least to replace `posnum`).

Yes. We should definitely keep the name "positive cone" and just change the name `posnum`.

> I did not implement it as a factory because of non-forgetful inheritance (the <= being built out of posnum and subtraction). Maybe a better solution would be to add an axiom `le_sub_def : forall x y, (x <= y) = posnum (y - x)` or something like that.

Where would be the non-forgetful inheritance? There is no instance involved here... That being said it would indeed duplicate the posnum/nonneg definition and the one from `interval_inference` is the one that is the most useful. Also an unspoken invariant of factories, written [here](https://github.com/math-comp/hierarchy-builder/issues/541) but not yet enforced, that makes having `posnum` part of a factory an issue.

proux01: My bad, I miswrote `HB.factory Record ZmodulePositiveCone (R : zmodType)` instead of `HB.factory Record ZmodulePositiveCone R of GRing.Zmodule R` and this lead to a spurious non forgetful inheritance warning (and other issues). Now fixed.

PR #1446, opened 2025-07-24
Tensor formalization

joshua-smart:
##### Motivation for this change

MathComp does not currently have a formalization for the tensor data structure. This would have significant application in the field of Neural Network verification, and an immediate use case within the [vehicle](https://github.com/vehicle-lang/vehicle) project (see https://github.com/vehicle-lang/vehicle/pull/900).

##### Minimal TODO list

- [ ] added changelog entries with `doc/changelog/make-entry.sh`

- [ ] added corresponding documentation in the headers
- [ ] tried to abide by the [contribution guide](https://github.com/math-comp/math-comp/blob/master/CONTRIBUTING.md)
- [ ] this PR contains an optimum number of meaningful commits

See [this Checklist](https://github.com/math-comp/math-comp/wiki/Checklist-for-creating-and-review-PRs) for details.

##### Automatic note to reviewers

Read [this Checklist](https://github.com/math-comp/math-comp/wiki/Checklist-for-creating-and-review-PRs).

PR #1535, opened 2026-02-12
 Tensor formalization

hoheinzollern:

This PR introduces a tensor formalization, with the changes discussed in PR #1446 (moving to finite functions instead of lists for the index types), renaming the multiplication to the Hadamard product and introducing proper tensor product with appropriate lemmas.

As it stands, the tensor product is not an operator of type `V -> V -> V`, as the types of the operands can differ and so the resulting output. Therefore, we are currently missing some instances (it's incompatible with standard algebra interfaces).

##### Minimal TODO list


algebra/tensor.v
```diff
@@ -0,0 +1,853 @@
...
+
+Context {R : pzSemiRingType}.
+
+Definition tensor1 := @const_t _ _ _ u_ d_ (GRing.one R).
+
+(* Hadamard product: element-wise multiplication *)
+Definition hmult (t u : 'T[R]_(u_, d_)) :=
+  @Tensor _ _ u_ d_ R (map2_mx *%R (\val t) (\val u)).
+
+Lemma hmultA : associative hmult.
```
affeldt-aist:
```suggestion
#[local] Fact hmultA : associative hmult.
```

CohenCyril (commented):
mult -> mul

algebra/tensor.v
```diff
@@ -0,0 +1,853 @@
...
+Lemma prod_card (m n : nat) : #|{:'I_m * 'I_n}| = (m * n)%N.
+Proof. by rewrite card_prod !card_ord. Qed.
+
+Definition prod_split (m n : nat) (i : 'I_(m * n)) : 'I_m * 'I_n :=
+  enum_val (cast_ord (esym (prod_card m n)) i).
+
+Definition prod_unsplit (m n : nat) (ij : 'I_m * 'I_n) : 'I_(m * n) :=
+  cast_ord (prod_card m n) (enum_rank ij).
+
+Definition tprod (t : 'T[R]_(u1_, d1_)) (u : 'T[R]_(u2_, d2_))
```
affeldt-aist: Rename to `mults`? (see the minutes of meeting of 2026-04-15)

algebra/tensor.v
```diff
@@ -0,0 +1,853 @@
...
+From mathcomp Require Import ssreflect seq matrix bigop ssrbool eqtype choice.
+From mathcomp Require Import fintype ssralg ssrnat ssrfun order finfun tuple.
+From mathcomp Require Import finset sesquilinear.
+From mathcomp Require Import interval interval_inference numdomain.
+From Corelib Require Import ssreflect.
+
+(******************************************************************************)
+(* This file defines tensors,                                                 *)
+(* For tensors we define:                                                     *)
+(*                 pseq == the type of sequences of strictly positive natural *)
```
affeldt-aist: Quentin: documented but not in the code!

algebra/tensor.v
```diff
@@ -0,0 +1,853 @@
...
+  const_t (a * b)%R.
+Proof.
+by apply/val_inj/matrixP => i j; rewrite /tensor_val /= !mxE.
+Qed.
+
+
+End TensorProduct.
+
+Notation "*t%R" := tprod : ring_scope.
+Notation "x *t y" := (tprod x y) (at level 40, left associativity) : ring_scope.
```
affeldt-aist: Use `Reserved Notation`?

algebra/tensor.v
```diff
@@ -0,0 +1,853 @@
...
+(* structures (including finType), as well as nmodType, zmodType,             *)
+(* lSemiModType, lModType.                                                    *)
+(*   Tensors also implement a pointwise partial ordering, as well as          *)
+(* ring instances where the underlying type satisfies that instance.          *)
+(* NOTE: Ring multiplication ( *%R ) is the Hadamard (element-wise) product.  *)
+(* Proper tensor product forms a bilinear structure.                          *)
+(*                                                                            *)
+(* Tensor operations:                                                         *)
+(* t *h u == Hadamard product of t and u (element-wise multiplication)        *)
+(* t *t u == proper tensor product: combines dimensions,                      *)
```
t6s, line 73: I have not examined the actual definition yet, but isn't this the Kronecker product of t and u? If so, I suggest renaming it Kronecker product (`kprod`), and other identifier accordingly. I think "tensor product" usually refers to an operation on vector spaces, not on vectors.

CohenCyril: "Tensor product" also applies to matrices and tensors (see eg. https://www.mathworks.com/help/matlab/ref/tensorprod.html). Indeed the map $\otimes : V \times W \to V \otimes W$ is also called tensor product, and tensors are essentially elements of a particular case of homogeneous tensor product $V^{\otimes m} \otimes {(V^*)}^{\otimes n}$...

t6s: Matlab doc is authoritative enough to change my mind.

algebra/tensor.v
hoheinzollern: That was a relic from the previous version.

algebra/tensor.v
```diff
@@ -0,0 +1,859 @@
...
+
+Context {R : pzSemiRingType}.
+
+Definition tensor1 := @const_t _ _ _ u_ d_ (GRing.one R).
+
+(* Hadamard product: element-wise multiplication *)
+Definition hmul (t u : 'T[R]_(u_, d_)) :=
+  @Tensor _ _ u_ d_ R (map2_mx *%R (\val t) (\val u)).
+
+#[local] Fact hmulA : associative hmul.
```
affeldt-aist:
```suggestion
#[local] Lemma hmulA : associative hmul.
```

algebra/tensor.v
```diff
@@ -0,0 +1,859 @@
...
+Proof. by move=> x y z; rewrite /hmul map2_mxA. Qed.
+Lemma hmul1t : left_id tensor1 hmul.
```
affeldt-aist:
```suggestion
#[local] Lemma hmul1t : left_id tensor1 hmul.
```

affeldt-aist: The same for all lemmas that are just used with `HB.instance` (we decided `Lemma` instead of `Fact` during the last meeting).

pi8027: However, note that we haven't started putting the `local` attribute to see its consequences (see #1563 and #1569). I can do that once I finish splitting order.v.

pi8027: > we decided `Lemma` instead of `Fact` during the last meeting

I don't recall the discussion, but the minutes seem to say the opposite: "Use `#[local] Fact`". https://github.com/math-comp/math-comp/wiki/Minutes-2026%E2%80%9004%E2%80%9015

affeldt-aist: Arg, I do remember that we chose `Lemma` because insisting on using `Fact` is just an additional burden with little gain. I also wrote the minutes... I think that it is a mistake in the minutes (the kind of mistake I easily do when doing several things at a time). Not 100% sure though.

proux01: I also remember `#[local] Fact` bu my memory may betray me. Sorry, I should have checked the minutes more carefully.

affeldt-aist: Ah but wait, this looks like a comment I did live during the meeting, so maybe we indeed ended up deciding `Fact`. If that is the case, that doesn't sound like a good idea to me: it is easy to remember you want to put a `#[local]` but easy to overlook that one needs the seldom-used `Fact`. (I already miss the `Let` feature... which by the way has the advantage of freeing the name once the section is closed, not the case of `Fact` and `Lemma` even when they are local.)

affeldt-aist: Whatever is decided should be put in `CONTRIBUTING.md`.

proux01: Let's rediscuss it during next meeting (I don't care that much, Lemma or Fact is just a comment anyway).

affeldt-aist: The choice between Lemma and Fact is less important than between Let and Lemma/Fact, a related discussion: https://github.com/rocq-prover/rocq/issues/10459#issuecomment-507574702

hoheinzollern: Ok, I'm happy to go either way, should I revert to Lemma?

pi8027: On `Lemma` vs `Fact`, I prefer the latter since we can easily find them by `grep`. So, I don't think it is a "little gain".

On `#[local]`, my understanding is that we decided (in #1563) to **try** putting it everywhere to see what happens, but we haven't decided yet to actually put it everywhere.

affeldt-aist: Can I clarify one thing? Since this conversation (https://github.com/rocq-prover/rocq/issues/10459#issuecomment-507574702) is marked as "not planned", why not stick to `Let`?
(During the meeting, I understood that we cannot rely on the semantics of `Let` but since
it is not planned to change, why not?)

proux01: > Ok, I'm happy to go either way, should I revert to Lemma?

Whatever you prefer. I wouldn't bother too much since it seems we don't have a definitive choice. So let's merge the PR with whatver you prefer and we'll potentially "fix" it later (and a nice thing with the current `#[local] Fact` is that whatever we decide, it should be relatively easy to "fix").

pi8027: I believe that the majority of internal lemmas in MathComp are currently marked as `Fact`. If you want to merge this PR before the release (and thus don't want to wait for the discussion and decision), going for `Fact` would make the uniformization process easier (again, because we can spot them by `grep`).

proux01: > I understood that we cannot rely on the semantics of `Let` but since it is not planned to change, why not?)

The semantics of `Let` + `Qed` is stable and what we want since https://github.com/rocq-prover/rocq/pull/17576 (8.19).

pi8027: Thanks @proux01, I was not aware of that. It sounds like we should use `Let` + `Qed` then.

I still prefer to first replace the existing `Fact` (ideally, all of them) with `Let` to see if that is really what we want.

hoheinzollern: OK thanks @proux01 I'll go with that.

hoheinzollern:
This PR should now be ready.

algebra/tensor.v
```diff
@@ -0,0 +1,879 @@
+From HB Require Import structures.
+From mathcomp Require Import ssreflect seq matrix bigop ssrbool eqtype choice.
+From mathcomp Require Import fintype ssralg ssrnat ssrfun order finfun tuple.
+From mathcomp Require Import finset sesquilinear.
+From mathcomp Require Import interval interval_inference numdomain.
+From Corelib Require Import ssreflect.
+
+(******************************************************************************)
+(* This file defines tensors,                                                 *)
+(* For tensors we define:                                                     *)
```
affeldt-aist: These two lines look redundant.

algebra/tensor.v
```diff
@@ -0,0 +1,879 @@
...
+Lemma otensor_eqP (d : {posnum nat}) (t v : 'oT[R]_(fcons d tensor_nil_f))
+  : t = v <-> forall i, t`_=i = v`_=i.
+Proof.
+split=> [->//|eq_i]; apply/otensorP=> i.
+by move: (eq_i i)=> /tensor_nil_eqP.
+Qed.
+
+(* Note: there seems to be some conflict between ring scope and the `_ notation *)
+
+Lemma nstackE {u : {posnum nat}} {k l} {u_ : {posnum nat} ^ k} {d_ : {posnum nat} ^ l} (f : 'I_u%:num%R -> 'T[R]_(u_, d_)) i : (nstack f)^^i = f i.
```
affeldt-aist: Maybe try to reformat to fit the 80 characters rule?

algebra/tensor.v
```diff
@@ -0,0 +1,879 @@
...
+Reserved Notation "''oT[' R ]_ ( d_ )" (at level 0, d_ at level 2).
+  (* only parsing *)
+Reserved Notation "''T_' ( u_ , d_ )"
+  (at level 0, u_ at level 2, d_ at level 2, format "''T_' ( u_ ,  d_ )").
+Reserved Notation "''T[' R ]_ ( u_ , d_ )"
+  (at level 0, u_ at level 2, d_ at level 2). (* only parsing*)
+
+Reserved Notation "t ^^ i"
+  (at level 3, i at level 2, left associativity, format "t ^^ i").
+Reserved Notation "t `_ i"
```
affeldt-aist, line 105: ``` `_ ``` looks like a notation that can be used elsewhere (I think it is already used to address tuples), so to decide conflicts maybe adding a scope to this file could be useful
as for non-conflicting notations, maybe they should be in the same scope as the notation for matrices

algebra/tensor.v
hoheinzollern: Fixed

algebra/tensor.v
hoheinzollern: Fixed

algebra/tensor.v
hoheinzollern: Addressed

algebra/tensor.v
pi8027: In case it was not clear enough, what I suggested was to use `Fact` for now, and later see whether replacing `Fact` with `Let` everywhere works as expected.

algebra/tensor.v
```diff
@@ -0,0 +1,912 @@
...
+
+End SubtypeInstances.
+
+
+Section TensorPOrder.
+
+Import Order.POrderTheory.
+Open Scope order_scope.
+
+Context (o : Order.disp_t) (R : porderType o).
```
pi8027: We usually use `d` or `disp` for variables of type `Order.disp_t`.

algebra/tensor.v
```diff
@@ -0,0 +1,912 @@
...
+Qed.
+
+Let le_t_trans : transitive (le_t).
+Proof.
+move=> x y z /forallP le_t_yx /forallP le_t_xz.
+apply/forallP=> ij; exact /le_trans.
+Qed.
+
+HB.instance Definition _ := Order.isPOrder.Build
+  o 'T[R]_(u_, d_) lt_t_def le_t_refl le_t_anti le_t_trans.
```
pi8027: This means that `'T[R]_(u_, d_)` has the same display as `R`, and thus, they share the same set of order notations. Is this really what you want? (It looks wrong to me.)

hoheinzollern: Actually, that made me realise that we probably shouldn't ship a default order for tensors, like matrices don't have one, this is application specific.

hoheinzollern: I removed the order theory from the tensor library.

algebra/tensor.v
hoheinzollern: I now added a tensor scope.

hoheinzollern:
After reviewing the code, we noticed two things:
- `fcat` and `prod_fcat` should live in `finfun.v` and `bigop.v` respectively (the latter generalized to `big_fcat`)
- we need similar machinery for reshaping tensors like in `castmx`, introduced here and called `castt` (please confirm that it works for your use @lstrsrmn)

algebra/tensor.v
```diff
@@ -0,0 +1,926 @@
...
+  := 'T[R]_( [tuple of u1%:posnat :: .. [:: uk%:posnat] ..] , [tuple] )
+  (only parsing).
+Notation "''oT[' R ]_ ( d_ )" := 'T[R]_( [tuple] , d_ ) (only parsing).
+Notation "''oT[' R ]_ [ d1 , .. , dl ]"
+  := 'T[R]_( [tuple] , [tuple of d1%:posnat :: .. [:: dl%:posnat] ..] )
+  (only parsing).
+Notation "''oT_' ( d_ )" := (tensor [tuple] d_ _).
+Notation "''nT_' ( u_ )" := (tensor u_ [tuple] _).
+
+Import Algebra GRing.
```
pi8027: We usually avoid importing `Algebra` and `GRing`.

algebra/tensor.v
```diff
@@ -0,0 +1,926 @@
...
+Let invt_out {R : unitRingType} : {in [predC @unitt R], invt =1 id}.
+Proof. by move=> t /negbTE t_not_unit; rewrite /invt t_not_unit. Qed.
+
+HB.instance Definition _ {R : unitRingType} :=
+  GRing.NzRing_hasMulInverse.Build 'T[R] mulVt divtt unittP invt_out.
+
+End TensorRing.
+
+(* Notations for Hadamard product *)
+Notation "*h%R" := hmul : ring_scope.
```
pi8027:
```suggestion
Notation "*h%R" := hmul : function_scope.
```

lstrsrmn:
In my opinion, a more powerful tensor indexing operator should be added. Example type signatures:
```coq
Definition tensor_nindex {R : Type} {k l : nat} {u_ : k.-tuple {posnum nat}} {d_ : l.-tuple {posnum nat}}
  (t : 'T[R]_(u_, d_)) (i : 'I_(\prod_(i <- u_) i%:posnum)) : R.
```
This is indexing fully, similarly, a partial indexing, but potentially more than 1 layer
```coq
Definition tensor_nindex_partial {R : Type} {k l : nat} {u_ : k.-tuple {posnum nat}} {d_ : l.-tuple {posnum nat}}
  (t : 'T[R]_(u_, d_)) (n : 'I_k) (i : 'I_(\prod_(i <- take n u_) i%:posnum)) : 'T[R]_(drop_tuple n u_, d_).
```
Or similar, as currently, the indexing in only one layer at a time which could be limiting.

lstrsrmn:
There are also still issues around moving between finite functions and tuples, specifically:
```coq
Definition test {k} (t : {posnum nat} ^k) : k.-tuple {posnum nat} := t.
```
I would expect this to coerce and work naturally, but it currently doesnt.

hoheinzollern:
> In my opinion, a more powerful tensor indexing operator should be added. Example type signatures:
> 
> ```coq
> Definition tensor_nindex {R : Type} {k l : nat} {u_ : k.-tuple {posnum nat}} {d_ : l.-tuple {posnum nat}}
>   (t : 'T[R]_(u_, d_)) (i : 'I_(\prod_(i <- u_) i%:posnum)) : R.
> ```
> 
> This is indexing fully, similarly, a partial indexing, but potentially more than 1 layer
> 
> ```coq
> Definition tensor_nindex_partial {R : Type} {k l : nat} {u_ : k.-tuple {posnum nat}} {d_ : l.-tuple {posnum nat}}
>   (t : 'T[R]_(u_, d_)) (n : 'I_k) (i : 'I_(\prod_(i <- take n u_) i%:posnum)) : 'T[R]_(drop_tuple n u_, d_).
> ```
> 
> Or similar, as currently, the indexing in only one layer at a time which could be limiting.

Through private conversation, we figured that these definitions are not actually helpful and break the expected interface.

algebra/tensor.v
```diff
@@ -0,0 +1,932 @@
...
+Let subsemimod_closed {m n} (R : pzSemiRingType)
+  : @subsemimod_closed R 'M[R]_(n, m) predT.
+Proof. by []. Qed.
+HB.instance Definition _ (R : pzSemiRingType) :=
+  SubNmodule_isSubLSemiModule.Build _ _ _ 'T[R] (subsemimod_closed R).
+
+Let submod_closed {m n} (R : pzRingType) : @submod_closed R 'M[R]_(n, m) predT.
+Proof. by []. Qed.
+HB.instance Definition _ (R : pzRingType) :=
+  GRing.SubChoice_isSubLmodule.Build _ _ _ 'T[R] (submod_closed R).
```
pi8027: The `SubChoice_isSubLmodule` factory generates a `subLSemiModType` instance and thus overlaps with the `HB.instance` declaration above.
```suggestion
HB.instance Definition _ (R : pzRingType) := GRing.SubLSemiModule.on 'T[R].
```

algebra/tensor.v
```diff
@@ -0,0 +1,932 @@
...
+  'T[R] hmulA hmul1t hmul1 hmulDl hmulDr hmul0t hmul0.
+
+End TensorSemiRing.
+
+Let hmulC {R : comPzSemiRingType} : @commutative 'T[R] _ hmul.
+Proof. by move=> x y; rewrite /hmul map2_mxC. Qed.
+
+HB.instance Definition _ {R : pzRingType} := GRing.Zmodule_isPzRing.Build
+  'T[R] (@mulrA 'T[R]) (@mul1r 'T[R]) (@mulr1 'T[R])
+    (@mulrDl 'T[R]) (@mulrDr 'T[R]).
```
pi8027: Again, this overlaps with other instances. I suggest removing it here and adding it below.
```suggestion
```

algebra/tensor.v
```diff
@@ -0,0 +1,932 @@
...
+
+HB.instance Definition _ {R : comPzRingType} :=
```
pi8027: This instance can be generalised to commutative semirings.
```suggestion
HB.instance Definition _ {R : comPzSemiRingType} := 
```

algebra/tensor.v
```diff
@@ -0,0 +1,932 @@
...
+  GRing.PzSemiRing_isNonZero.Build
+  'T[R] onet_neq0.
+
+HB.instance Definition _ {R : nzRingType} := GRing.Zmodule_isNzRing.Build
+  'T[R] (@mulrA 'T[R]) (@mul1r 'T[R]) (@mulr1 'T[R])
+    (@mulrDl 'T[R]) (@mulrDr 'T[R]) (@oner_neq0 'T[R]).
+
+HB.instance Definition _ {R : comNzRingType} := GRing.Zmodule_isComNzRing.Build
+  'T[R] (@mulrA 'T[R]) (@mulrC 'T[R]) (@mul1r 'T[R])
+    (@mulrDl 'T[R]) (@oner_neq0 'T[R]).
```
pi8027:
```suggestion
(* FIXME: HB.saturate *)
HB.instance Definition _ {R : pzRingType} := GRing.PzSemiRing.on 'T[R].
HB.instance Definition _ {R : nzRingType} := GRing.NzSemiRing.on 'T[R].
HB.instance Definition _ {R : comPzRingType} := GRing.PzRing.on 'T[R].
HB.instance Definition _ {R : comNzRingType} := GRing.NzRing.on 'T[R].
```
Unfortunately, `HB.saturate tensor` takes 50 seconds on my laptop.

algebra/tensor.v
```diff
@@ -0,0 +1,932 @@
...
+         (castt (eq_u, eq_d)).
+Proof. by move=> t; apply/val_inj; rewrite !val_castt castmxKV. Qed.
+
+End TensorCast.
+
+
+Section TensorRing.
+
+Local Open Scope ring_scope.
+Import GRing.Theory.
```
pi8027: I think there is no need to import `GRing.Theory` locally (unless you really need to do so to avoid name collisions of lemmas). It can be imported at the beginning of the file.

algebra/tensor.v
```diff
@@ -0,0 +1,932 @@
...
+  : (t + u).[::] = t.[::] + u.[::] :> R.
+Proof. by rewrite /tensor_nil mxE. Qed.
+
+Lemma tensor_nilN {R : zmodType} t
+  : (- t).[::] = - t.[::] :> R.
+Proof. by rewrite /tensor_nil mxE. Qed.
+
+Lemma tensor_nilM {R : pzSemiRingType} t u
+  : (t * u).[::] = t.[::] * u.[::] :> R.
+Proof. by rewrite /tensor_nil mxE. Qed.
```
pi8027: I think this means that `tensor_nil` is an N-module/semiring morphism.
```suggestion
Fact tensor_nil_is_nmod_morphism {U : nmodType} : nmod_morphism (@tensor_nil U).
Proof. by split=> [|? ?]; rewrite /tensor_nil mxE. Qed.

HB.instance Definition _ {U : nmodType} :=
  GRing.isNmodMorphism.Build 'T_([tuple], [tuple]) U (@tensor_nil U)
    tensor_nil_is_nmod_morphism.

Fact tensor_nil_is_monoid_morphism {R : pzSemiRingType} :
  monoid_morphism (@tensor_nil R).
Proof. by split=> [|? ?]; rewrite /tensor_nil mxE. Qed.

HB.instance Definition _ {R : pzSemiRingType} :=
  GRing.isMonoidMorphism.Build 'T_([tuple], [tuple]) R (@tensor_nil R)
    tensor_nil_is_monoid_morphism.
```

algebra/tensor.v
```diff
@@ -0,0 +1,932 @@
...
+Lemma const_tN {R : zmodType} x
+  : @const_t R _ _ u_ d_ (- x) = - const_t x.
+Proof. by apply/val_inj/matrixP=> i j; rewrite !mxE. Qed.
+
+Lemma const_tM {R : pzSemiRingType} x y
+  : @const_t R _ _ u_ d_ (x * y) = const_t x * const_t y.
+Proof. by apply/val_inj/matrixP=> i j; rewrite !mxE. Qed.
+
+Definition const_tr_spec {R : pzRingType} := 
+  (@const_tM R, @const_tN R, @const_tD R).
```
pi8027: Same here?

algebra/tensor.v
```diff
@@ -0,0 +1,932 @@
...
+apply/eqP; case; apply/matrixP; rewrite /const_mx/eqrel.
+case: (\prod_(i < k) (u_ i)%:num)
+    (@prodn_gt0 _ (index_enum 'I_k) predT (fun i=> (u_ i)%:num))=> [//|n0 _].
+  suff : (forall i : 'I_k, 0 < (u_ i)%:posnum) by move=> /[swap] /[apply].
+  by move => i; rewrite gtn0.
+case: (\prod_(i < l) (d_ i)%:posnum)
+    (@prodn_gt0 _ (index_enum 'I_l) predT (fun i=> (d_ i)%:posnum))=> [//|n1 _].
+  suff : (forall i : 'I_l, 0 < (d_ i)%:posnum) by move=> /[swap] /[apply].
+  by move => i; rewrite gtn0.
+by move=> /(_ ord0 ord0); rewrite !mxE; apply/eqP/oner_neq0.
```
pi8027:
```suggestion
apply/eqP; case=> /matrixP.
set i := \prod_(i < k) (u_ i)%:num; set j := \prod_(i < l) (d_ i)%:num.
have: 0 < i by apply/prodn_gt0 => ?; rewrite gtn0.
have: 0 < j by apply/prodn_gt0 => ?; rewrite gtn0.
case: i j => [|i] [|j]// _ _ /(_ ord0 ord0).
by apply/eqP; rewrite !mxE oner_neq0.
```

algebra/tensor.v
```diff
@@ -0,0 +1,932 @@
...
+
+Let nmod_closed {m n} (R : nmodType) : @nmod_closed 'M[R]_(n, m) predT.
+Proof. by []. Qed.
+HB.instance Definition _ (R : nmodType) := SubChoice_isSubNmodule.Build
+  _ _ 'T[R] (nmod_closed R).
+
+Let zmod_closed {m n} (R : zmodType) : @zmod_closed 'M[R]_(n, m) predT.
+Proof. by []. Qed.
+HB.instance Definition _ (R : zmodType) := SubChoice_isSubZmodule.Build
+  _ _ 'T[R] (zmod_closed R).
```
pi8027: `SubChoice_isSubZmodule` should not be used with `SubChoice_isSubNmodule` above (again, overlapping instances). I think the right solution is to add the `SubNmodule_isSubZmodule` factory to `nmodule.v`. I can open a pull request for this change.

Also, `zmod_closed` here can be replaced with `nmod_closed`:
```suggestion
HB.instance Definition _ (R : zmodType) := GRing.SubChoice_isSubZmodule.Build
  _ _ 'T[R] (nmod_closed R).
```
(This suggestion doesn't fix the issue, though.)

algebra/tensor.v
pi8027: I think we should actually consider adding the pointwise order for matrices, but I have no applications in mind. So, we can postpone it anyway.

algebra/tensor.v
hoheinzollern: Done.

algebra/tensor.v
hoheinzollern: Now done! Thanks @pi8027 for suggesting these changes!

algebra/tensor.v
hoheinzollern: Applied.

algebra/tensor.v
hoheinzollern: (was applied)

algebra/tensor.v
```diff
@@ -0,0 +1,918 @@
...
+  (at level 3, i at level 2, left associativity, format "t `_= i").
+
+Reserved Notation "\tensor ^^ i E"
+  (at level 34, E at level 39, i binder, format "\tensor ^^ i  E").
+Reserved Notation "\tensor `_ i E"
+  (at level 34, E at level 39, i binder, format "\tensor `_ i  E").
+Reserved Notation "\tensor ^^ ( i < u ) E"
+  (E at level 39, i, u at level 50). (* only parsing *)
+Reserved Notation "\tensor `_ ( i < d ) E"
+  (E at level 39, i, d at level 50). (* only parsing *)
```
proux01: To compile on Rocq 9.2:
```suggestion
Reserved Notation "\tensor ^^ ( i < u ) E"
  (at level 34, E at level 39, i, u at level 50). (* only parsing *)
Reserved Notation "\tensor `_ ( i < d ) E"
  (at level 34, E at level 39, i, d at level 50). (* only parsing *)
```

algebra/tensor.v
```diff
@@ -0,0 +1,918 @@
...
+Local Notation u_cons := [tuple of u :: u_].
+
+Lemma tensormx_cast : #|{:'I_u%:num * 'I_(\prod_(i < k) (u_ i)%:num)}|
+                        = \prod_(i < k.+1) (u_cons i)%:num.
+Proof.
+rewrite card_prod !card_ord big_ord_recl ffunE [tnth _ _]tnth0.
+by congr (_ * _); apply: eq_bigr => i _; rewrite !ffunE tnthS.
+Qed.
+
+Definition tensormx_index (ij : 'I_u%:num * 'I_\prod_(i < k) (u_ i)%:num)
```
proux01:
```suggestion
Definition tensormx_index (ij : 'I_u%:num * 'I_(\prod_(i < k) (u_ i)%:num))
```
(and so on) for >= 9.2

algebra/tensor.v
pi8027: The other `` `_ `` is defined in `nmodule.v` and used in `poly.v`: https://github.com/math-comp/math-comp/blob/60270fd2cf06e70860a194d4b0e800095811b20a/boot/nmodule.v#L1164 https://github.com/math-comp/math-comp/blob/9661ff7e9b8b663ef1cc4eabe7e4e3d7ac517482/algebra/poly.v#L29-L31

I think adding `tensor_scope` is not an ideal solution: `tensor.v` provides instances of module and ring-like structures for tensors, and thus, the notations for tensors should probably belong to `ring_scope`.

Perhaps, it's similar to the question of "sub scopes", e.g., `ring_scope` and `int_scope` in `ssrint.v`.

lstrsrmn:
There seems to be issues with the `[tensor^^= ...]` notation. It expects the list to be of type `(x%:posnum).-tuple R`, where it naturally becomes `x.-tuple R` (with x instantiated). Minimal example:

```coq
From mathcomp Require Import all_boot.
From mathcomp Require Import all_algebra.
From mathcomp Require Import interval_inference.
Local Open Scope ring_scope.
Local Open Scope order_scope.
Local Open Scope tensor_scope.

Definition meanScalingValues (R : realFieldType) : 'nT[R]_[2] := [tensor^^= (1 : R); (1 : R)].
```
This produces the error:
```coq
The term "[tuple 1 : R; 1 : R]" has type "2.-tuple R" while it is expected to have type "(?x%:posnum).-tuple ?R".
```

algebra/tensor.v
```diff
@@ -0,0 +1,918 @@
...
+
+Definition ostack_tuple (t : x%:num%R.-tuple 'T[R]_(u_, d_)) :=
+  \tensor`_i tnth t i.
+
+Lemma nstack_tupleE t i : (nstack_tuple t)^^i = tnth t i.
+Proof. exact: nstackE. Qed.
+
+Lemma ostack_tupleE t i : (ostack_tuple t)`_i = tnth t i.
+Proof. exact: ostackE. Qed.
+
```
lstrsrmn, line 745:
```suggestion
Lemma ntensor_of_tupleK : cancel ntensor_of_tuple tuple_of_ntensor.
Proof.
by move=> t; apply/eq_from_tnth=> i; rewrite tnth_mktuple nstack_eqE.
Qed.

Lemma tuple_of_ntensorK : cancel tuple_of_ntensor ntensor_of_tuple.
Proof.
by move=> t; apply/ntensor_eqP=> i; rewrite nstackE const_tK tnth_mktuple.
Qed.

Lemma otensor_of_tupleK : cancel otensor_of_tuple tuple_of_otensor.
Proof.
by move=> t; apply/eq_from_tnth=> i; rewrite tnth_mktuple ostack_eqE.
Qed.

Lemma tuple_of_otensorK : cancel tuple_of_otensor otensor_of_tuple.
Proof.
by move=> t; apply/otensor_eqP=> i; rewrite ostackE const_tK tnth_mktuple.
Qed.

```
Adding cancels to some operators

hoheinzollern:
Linking relevant Zulip conversation: https://rocq-prover.zulipchat.com/#narrow/channel/237664-math-comp-users/topic/Issue.20with.20coercing.20tuple.20sizes.20to.20.60.7Bposnum.20nat.7D.60

hoheinzollern:
The CI is still failing due to `mathcomp-abel` defining the same `^^` at a different level, so I'm attempting to unify the two notations (in favor of `mathcomp-abel` given `tensor.v` is the new kid on the block).

boot/finfun.v
```diff
@@ -16,6 +16,9 @@ From mathcomp Require Import fintype tuple.
 (*   define inductive types, e.g., Inductive tree := node n of tree ^ n (see  *)
 (*   mid-file for an expanded example).                                       *)
 (* --> More generally, {ffun fT} is always structurally positive.             *)
+(*    fcat f g == concatenation of f : T ^ n and g : T ^ m as a finfun        *)
+(*               of type T ^ (n + m), with lemmas fcat_lshift, fcat_rshift    *)
+(*      f +++ g == infix notation for fcat f g                                *)
```
proux01: Please align the `==` with the ones below (oar at least both of them)

boot/finfun.v
```diff
@@ -16,6 +16,9 @@ From mathcomp Require Import fintype tuple.
 (*   define inductive types, e.g., Inductive tree := node n of tree ^ n (see  *)
 (*   mid-file for an expanded example).                                       *)
 (* --> More generally, {ffun fT} is always structurally positive.             *)
+(*    fcat f g == concatenation of f : T ^ n and g : T ^ m as a finfun        *)
+(*               of type T ^ (n + m), with lemmas fcat_lshift, fcat_rshift    *)
```
proux01:
```suggestion
(*    fcat f g == concatenation of f : T ^ n and g : T ^ m as a finfun        *)
(*               of type T ^ (n + m)    *)
```
we don't document lemmas in the header doc, they are to be found through the Search command

boot/finfun.v
proux01:
```suggestion
(*      f +++ g := fcat f g                                *)
```

boot/finfun.v
```diff
@@ -290,6 +293,25 @@ Qed.
 
 End FinFunTuple.
 
+Reserved Notation "f +++ g" (at level 60, right associativity).
+
```
proux01: Please move that line at the beginning of the file

boot/ssrnat.v
```diff
@@ -399,6 +399,9 @@ Proof. by elim: n m p => [|i IHn] [|m] [|p] //; apply: IHn m p. Qed.
 
 Lemma leq_ltn_trans n m p : m <= n -> n < p -> m < p.
 Proof. by move=> Hmn; apply: leq_trans. Qed.
+ 
+Lemma ltn_leq_trans n m p : m < n -> n <= p -> m < p.
+Proof. by move=> Hmn; apply: leq_trans. Qed.
```
proux01:
```suggestion
Proof. exact: leq_trans. Qed.
```

proux01 (approved):
I didn't actually review the tensor.v file.

As a side note, I'm not sure using the dependent type `{posnum nat}` rather than encoding `n.+1` as `n` is really worth the effort here.

Another question: what would happen if we allowed zero size tensors? (just like we allow zero size matrices)

hoheinzollern:
> As a side note, I'm not sure using the dependent type `{posnum nat}` rather than encoding `n.+1` as `n` is really worth the effort here.

We have run into this issue and discussed it with @lstrsrmn for another formalization that depends on this. We will have to make the `{posnum nat}` interface more ergonomic in the future.

> Another question: what would happen if we allowed zero size tensors? (just like we allow zero size matrices)

Zero-size tensors work well as is, the `tensor_nil` theory covers that.

proux01:
> > Another question: what would happen if we allowed zero size tensors? (just like we allow zero size matrices)
> 
> Zero-size tensors work well as is, the `tensor_nil` theory covers that.

Then I'm lost, I don't understand why we require non zero sizes. But again, I didn't really delve into it.

hoheinzollern:
> > > Another question: what would happen if we allowed zero size tensors? (just like we allow zero size matrices)
> > 
> > 
> > Zero-size tensors work well as is, the `tensor_nil` theory covers that.
> 
> Then I'm lost, I don't understand why we require non zero sizes. But again, I didn't really delve into it.

In matrix you have a single size for rows/colums, while in tensor you have a list of sizes. The zero size is the empty list, but when there's an index we require that it comes from a strictly positive ordinal, otherwise there are multiple tensors with empty size but different index shapes. In principle, this could be a design choice, but having 0-sized indexes could be ugly. Worth keeping in mind if it bites back in the future, e.g., if we need some kind of split operation that returns a 0-sized index.

lstrsrmn:
I think that in practice having the ability to construct empty tensors is more inconvenient. With the current version, the problems with `{posnum nat}` only affect constructing new tensors, and this can be avoided by explicitly converting nats to posnums. However, dealing with a 0 somewhere in a long list of dimensions could be very hard and could create more problems down the road. As with most things, you deal with using tensors a lot more than constructing them, and I think for usability the posnum nats are better from my experience.

Issue #1505, opened 2025-11-28
Uniformize the name of the re-exportation files


pi8027:
Pointed out by @CohenCyril during the MathComp meeting or the sharing day: `all_*.v`, `ssrnum.v`, and `ssralg.v` (in #1504) re-export the contents of all the files in the same directory. These files should have the same name, e.g., `all.v`, so that the user can easily spot them.

To require/import one of these `all.v`, we need to give a prefix to disambiguate it, e.g., `From mathcomp Require Import algebra.all`.

affeldt-aist:
Can't we have, say, an `algebra.v` file in the directory `algebra` (or `algebra_theory`) so that a user just has to do `Require Import algebra`?

Issue #1508, opened 2025-12-01
Split order.v


pi8027:
Since `order.v` is 8k lines long (currently the largest file in MathComp), I propose splitting `order.v`, perhaps into the following files:
- `porder.v`: partial orders,
- `lattices.v`: semilattices, lattices, distributive lattices,
- `total.v`: total orders,
- `complemented.v`: complemented lattices, and
- `order.v` re-exports everything to provide the compatibility layer for the old `order.v`.

`order.v` should eventually be replaced by `all.v`, deprecated, and then removed. After that, we can consider renaming `total.v` to `order.v`.

In addition, we can probably get rid of the `Order.LTheory`, `Order.CTheory`, and `Order.TTheory` modules. All this stuff should be called `Order.Theory`, and we need to import only the files we need *in the dependency order* before importing `Order.Theory` (as in #1504). (The only problem is that we cannot deprecate modules...)

Furthermore, it will probably enable us to compile `num_theory` without waiting for `complemented.v` to compile.

pi8027:
(Also, these lexicographic and pointwise orders are probably not at all used in MathComp. Isolating them enables more parallel compilation.)

pi8027:
I'm unassigning myself since I probably don't have time to work on this very soon. Sorry about that. :(

pi8027:
FYI, I still plan to work on this, but probably not before April 20.

PR #1547, opened 2026-02-27
Rename alg/, num_theory/, and algebra.v

pi8027:

These two directories should follow the same naming convention.

~IMO, this should be consistent with `alg/`.~

~We can consider renaming `alg/` to `alg_theory/` instead, but I slightly prefer renaming `num_theory/` to `num/`.~

~It will also be related to the naming convention of the re-exportation files (#1505). If we decide to use short names for the directories (`alg/` and `num/`), their re-exportation files should probably be `all_alg.v` and `all_num.v`.~

~We discussed it in a meeting, but we probably didn't make a decision on whether we keep `_theory` or not: https://github.com/math-comp/math-comp/wiki/Minutes-2025-12-10 https://github.com/math-comp/math-comp/issues/1505#issuecomment-3636335448~

##### Minimal TODO list


CohenCyril (approved):
I think this is the better way to homogenize.

pi8027:
I added the changelog entry, but we can wait for others' opinions.

pi8027:
@CohenCyril @affeldt-aist  I updated the PR following: https://github.com/math-comp/math-comp/wiki/Minutes-2026%E2%80%9004%E2%80%9001 (I was not there for making the decision, so I may have misunderstood something.)

This PR performs only the changes related to `algebraic_hierarchy/` and `numeric_hierarchy/`.

And I have a few objections:

> `ssralg.v` -> `algebraic_hierarchy.v`

`ssralg.v` should not be renamed to `algebraic_hierarchy.v`. We should rather introduce the latter as a new file because:
- `ssralg.v` contains a lot of deprecation stuff which `algebraic_hierarchy.v` should not reexport, and
- `ssralg.v` should not re-export `countalg.v`, `finalg.v`, and `ring_quotient.v`, while `algebraic_hierarchy.v` should do so.

Perhaps the same applies to `ssrnum.v`.

> `order.v`

If we split `order.v` (see #1508), the issue should mostly disappear.

pi8027:
BTW, I strongly prefer to merge this before the release so that we can rename `algebra.v` without deprecation.

CohenCyril:
Hi, looks good to me, except for the missing "s" in "rings_modules_and_algebra*s*", unless that was a deliberate choice?

pi8027:
Right, I'm fixing it.

pi8027:
Done.

pi8027:
I suggest opening an issue for keeping track of other renamings to be done. Perhaps that should subsume #1505 and a half of #1557.

proux01 (approved):
LGTM
Should we also add a new `algebraic_hierarchy.v` that would be a copy of `ssralg.v` without the deprecations and deprecate `ssralg.v`, since we agreed on it? (and same for `ssrnum.v`)

pi8027:
I would rather wait until we can rename the existing `all_*.v` files to introduce `algebraic_hierarchy.v` and `numeric_hierarchy.v`.

pi8027:
I think we should also move `countalg.v`, `finalg.v`, and `ring_quotient.v` to `algebraic_hierarchy/` before introducing `algebraic_hierarchy.v`. I won't do that before the release.

PR #1554, opened 2026-03-05
[CI] Update Nix toolbox

proux01:
And update ocamlPackages.elpi version, which should fix CI on Rocq master


proux01:
Hopefully, it will remain green more than a couple hours this time

PR #1169, opened 2024-02-11
Preorder

Tragicus:
##### Motivation for this change

Defines preordered types and generalizes lemmas on `porderType` to `preorderType`.

#### Dependencies (Edit by Cyril)

- https://github.com/coq/coq/pull/19611

##### Things done/to do

- [ ] added corresponding entries in `CHANGELOG_UNRELEASED.md`

- [x] added corresponding documentation in the headers
- [x] tried to abide by the [contribution guide](https://github.com/math-comp/math-comp/blob/master/CONTRIBUTING.md)

##### Compatibility with MathComp 1.X

- [ ] ~I added the label `TODO: MC-1 port` to make sure someone ports this PR to
      the `mathcomp-1` branch **or** I already opened an issue or PR (please cross reference).~ (let's not backport that)

##### Automatic note to reviewers

Read [this Checklist](https://github.com/math-comp/math-comp/wiki/Checklist-for-creating-and-review-PRs) and put a milestone if possible.

#### Overlay (to be merged before the current PR)

* https://github.com/math-comp/analysis/pull/1577

PR #1578, opened 2026-04-16
Clean up the documentation of `order.v`

### LINKED ISSUE #1516: about the documentation of `preorder.v` and `order.v`
`order_morphism` appears as a `Definition` in `preorder.v`

https://github.com/math-comp/math-comp/blob/bfeeb3f3090c0c0a44d841f8d4d79d6eefc198db/order/preorder.v#L1940

but is not documented in the header while accessible afterwards at least via `Order.order_morphism`.

On the other hand, the documentation of `order.v` talks about `nondecreasing` but it looks like a duplicate of documentation already in `preorder.v`.

Maybe this can be clarified.

pi8027:

Most of these definitions now belong to `preorder.v` and are documented there.

~`\max_<range>` and `\min_<range>` appear to be exceptions: there are `Reserved Notation` declarations for them in `preorder.v`, but they are actually not defined.~

This PR also restores `\max_<range>` and `\min_<range>`, which were accidentally removed in #1169.

Fix #1516

##### Minimal TODO list


pi8027:
The documentation of `preorder.v` and `order.v` still contains some duplications: theory modules, displays, inference (parsing), etc. I will probably move them to `all_order.v` to ease the splitting of `order.v` (#1508).

proux01:
> `\max_<range>` and `\min_<range>` appear to be exceptions: there are `Reserved Notation` declarations for them in `preorder.v`, but they are actually not defined.

They used to be but were (accidentally?) removed in https://github.com/math-comp/math-comp/pull/1169 . I think the right fix would be to restore them rather than removing the doc. Cc @Tragicus for confirmation.

Tragicus:
Yes, that is definitely a mistake on my part. Would you like for me to open a separate PR to restore them?

proux01:
It could also just be a commit in the current PR, as you prefer.

proux01 (approved):
Let's merge when CI is happy

Tragicus:
There was the same issue with `\max^d` and `\min^d`.
Should I add something in the changelog?

proux01:
Indeed, a changelog entry would be welcome (since the accidental removal was released in 2.5.0)

pi8027:
@proux01 @Tragicus Thanks! It looks good to me.

pi8027:
@proux01 @Tragicus FYI, `\min^d_` and `\max^d_` were already defined in `order.v`. So we should have moved them from `order.v` to `preorder.v`. I'm trying to fix it in #1590.

PR #1581, opened 2026-04-21
Rename packages fingroup -> finite-group and character -> group-representation

proux01:
##### Motivation for this change

This will enable the `all_foo.v` -> `foo.v` renamings by clearing the `fingroup.v` and `character.v` conflicts.

#### Overlay (to be merged before the current PR)

* https://github.com/SSProve/ssprove/pull/117

##### Minimal TODO list

- [ ] added changelog entries with `doc/changelog/make-entry.sh`

- [ ] ~added corresponding documentation in the headers~
- [x] tried to abide by the [contribution guide](https://github.com/math-comp/math-comp/blob/master/CONTRIBUTING.md)
- [x] this PR contains an optimum number of meaningful commits

See [this Checklist](https://github.com/math-comp/math-comp/wiki/Checklist-for-creating-and-review-PRs) for details.

##### Automatic note to reviewers

Read [this Checklist](https://github.com/math-comp/math-comp/wiki/Checklist-for-creating-and-review-PRs).

PR #1582, opened 2026-04-21
Rename all_foo.v -> foo.v

proux01:
~Depends on https://github.com/math-comp/math-comp/pull/1581~ (merged)

##### Minimal TODO list


algebra/algebraic_hierarchy/algebraic_hierarchy.v
```diff
@@ -0,0 +1,22 @@
+From mathcomp Require Import ssreflect ssrfun.
+From mathcomp Require Export nmodule rings_modules_and_algebras divalg decfield.
```
pi8027: `nmodule.v` is not in the same directory. Do we really want to export it here? I prefer to have fewer exceptions.

On the other hand, we should move `countalg.v`, `finalg.v`, and `ring_quotient.v` to `algebraic_hierarchy/` and export it here IMO.

algebra/algebraic_hierarchy/algebraic_hierarchy.v
```diff
@@ -0,0 +1,22 @@
...
+(* Rideau, Packaging mathematical structures, TPHOLs 2009                     *)
+(******************************************************************************)
+
+Set Implicit Arguments.
+Unset Strict Implicit.
+Unset Printing Implicit Defensive.
+
+#[deprecated(since="mathcomp 2.5.0", use=GRing.monoid_morphism)]
+Definition multiplicative (R S : pzSemiRingType) (f : R -> S) : Prop :=
+  {morph f : x y / x * y}%R * (f 1 = 1)%R.
```
pi8027: Why did you move this definition from `ssralg.v`?

algebra/numeric_hierarchy/numeric_hierarchy.v
```diff
@@ -0,0 +1,38 @@
...
+
+Module Theory.
+Export Num.Theory.
+End Theory.
+
+Module Def.
+Export Num.Def.
+End Def.
+
+End Num.
```
pi8027: This is basically a copy of the `Num` module of `numfield.v`, and I guess we can remove it.

pi8027 (commented):
As I said, I'm not in favor of renaming the fingroup package, and I'd prefer to wait for a few years. Also, I'm not in favor of merging this kind of big change in a rush before a release.

In addition, I have a few comments below.

algebra/algebraic_hierarchy/algebraic_hierarchy.v
proux01: As the PR title tells, this is just a renaming of `ssralg.v`. Other changes should come in an independent PR.

algebra/algebraic_hierarchy/algebraic_hierarchy.v
proux01: Because it's used in another deprecated definition.

algebra/numeric_hierarchy/numeric_hierarchy.v
proux01: Again, not mere renaming, should go in another PR.

algebra/algebraic_hierarchy/algebraic_hierarchy.v
pi8027: Adding more exportation here later may introduce minor breakages, e.g., notation issues. IMO, renaming of the re-exportation file is the right occasion to change the set of files re-exported here. (This way, users should be able to spot and fix the breakage when switching from `ssralg` to `algebraic_hierarchy`.)

proux01: I disagree, bundling changes like that can make handling the renaming harder.

proux01:
CI green, this is ready

algebra/algebraic_hierarchy/algebraic_hierarchy.v
pi8027: Although this isn't critical, I prefer to move it to `rings_modules_and_algebras.v` then.

algebra/algebraic_hierarchy/algebraic_hierarchy.v
pi8027: My intention while working on #1504 was that the transition from `ssralg.v` to `algebraic_hierarchy.v` should not be seen as "renaming", so that we can do the following bundling changes:
- removing deprecated stuff,
- removing `nmodule.v`, and
- adding `countalg.v`, `finalg.v`, and `ring_quotient.v`.

So, IMO, `algebraic_hierarchy.v` should be presented as a "new re-exportation file". Although we deprecate `ssralg.v` in favor of `algebraic_hierarchy.v`, they are not the same thing, and this should be clarified in the deprecation message and the changelog entry.

proux01: Ok, I'm not sure I have an opinion on that and I don't have enough time to get one, so I'm removing the contentious commits from this PR and I let you do whatever you want in further PR(s).

pi8027 (approved):
A changelog entry is still needed.

pi8027:
I'm fine to do the rest of the work, but it won't happen before the release.

Issue #1487, opened 2025-10-21
Remove the `pz` prefix from (semi)ring and (semi)algebra structures


pi8027:
IIRC, this was a part of the plan, but I couldn't find any open issue for it. I suggest doing so after the release of 2.6.0 (the first release that includes #1482).

PR #1586, opened 2026-04-24
Documented but missing notations

affeldt-aist:
`SubChoice_isSubComPzSemiRing`
`SubChoice_isSubPzSemiRing`
`SubChoice_isSubPzRing`
`SubChoice_isSubComPzRing`
were documented but not implemented.

##### Minimal TODO list

### DIFF
--- a/algebra/algebraic_hierarchy/rings_modules_and_algebras.v
+++ b/algebra/algebraic_hierarchy/rings_modules_and_algebras.v
@@ -223,16 +223,20 @@ From mathcomp Require Import nmodule.
 (*                           products (1 and x * y in S for x, y in S)        *)
 (*      semiring_closed S <-> collective predicate S is closed under semiring *)
 (*                           operations (0, 1, x + y and x * y in S)          *)
-(* [SubNmodule_isSubPzSemiRing of R by <:] ==                                 *)
+(* [SubNmodule_isSubPzSemiRing of R by <:] == pzSemiRingType mixin for a      *)
+(*                           subNmodType whose base type is a pzSemiRingType  *)
+(*                           and whose predicate is a mulrClosed              *)
 (* [SubChoice_isSubPzSemiRing of R by <:] == pzSemiRingType mixin for a       *)
 (*                           subType whose base type is a pzSemiRingType and  *)
-(*                           whose predicate's is a semiringClosed            *)
+(*                           whose predicate is a semiringClosed              *)
 (*                                                                            *)
 (*  * NzSemiRing (non-commutative non-trivial semirings):                     *)
-(* [SubNmodule_isSubNzSemiRing of R by <:] ==                                 *)
+(* [SubNmodule_isSubNzSemiRing of R by <:] == nzSemiRingType mixin for a      *)
+(*                           subNmodType whose base type is a nzSemiRingType  *)
+(*                           and whose predicate is a mulrClosed              *)
 (* [SubChoice_isSubNzSemiRing of R by <:] == nzSemiRingType mixin for a       *)
 (*                           subType whose base type is a nzSemiRingType and  *)
-(*                           whose predicate's is a semiringClosed            *)
+(*                           whose predicate is a semiringClosed              *)
 (*                                                                            *)
 (*  * PzRing (non-commutative rings):                                         *)
 (*         GRing.sign R b := (-1) ^+ b in R : pzRingType, with b : bool       *)
@@ -242,42 +246,34 @@ From mathcomp Require Import nmodule.
 (*                           and opposite (-1 and x * y in S for x, y in S)   *)
 (*       subring_closed S <-> collective predicate S is closed under ring     *)
 (*                           operations (1, x - y and x * y in S)             *)
-(* [SubZmodule_isSubPzRing of R by <:] ==                                     *)
 (* [SubChoice_isSubPzRing of R by <:] == pzRingType mixin for a subType whose *)
-(*                           base                                             *)
-(*                           type is a pzRingType and whose predicate's is a  *)
-(*                           subringClosed                                    *)
+(*                           base type is a pzRingType and whose predicate is *)
+(*                           a subringClosed                                  *)
 (*                                                                            *)
 (*  * NzRing (non-commutative non-trivial rings):                             *)
-(* [SubZmodule_isSubNzRing of R by <:] ==                                     *)
 (* [SubChoice_isSubNzRing of R by <:] == nzRingType mixin for a subType whose *)
-(*                           base                                             *)
-(*                           type is a nzRingType and whose predicate's is a  *)
-(*                           subringClosed                                    *)
+(*                           base type is a nzRingType and whose predicate is *)
+(*                           a subringClosed                                  *)
 (*                                                                            *)
 (*  * ComPzSemiRing (commutative PzSemiRings):                                *)
-(* [SubNmodule_isSubComPzSemiRing of R by <:] ==                              *)
 (* [SubChoice_isSubComPzSemiRing of R by <:] == comPzSemiRingType mixin for a *)
-(*                           subType whose base type is a comPzSemiRingType   *)
-(*                          and whose predicate's is a semiringClosed         *)
+(*                          subType whose base type is a comPzSemiRingType    *)
+(*                          and whose predicate is a semiringClosed           *)
 (*                                                                            *)
 (*  * ComNzSemiRing (commutative NzSemiRings):                                *)
-(* [SubNmodule_isSubComNzSemiRing of R by <:] ==                              *)
 (* [SubChoice_isSubComNzSemiRing of R by <:] == comNzSemiRingType mixin for a *)
 (*                           subType whose base type is a comNzSemiRingType   *)
-(*                          and whose predicate's is a semiringClosed         *)
+(*                          and whose predicate is a semiringClosed           *)
 (*                                                                            *)
 (*  * ComPzRing (commutative PzRings):                                        *)
-(* [SubZmodule_isSubComPzRing of R by <:] ==                                  *)
 (* [SubChoice_isSubComPzRing of R by <:] == comPzRingType mixin for a         *)
 (*                           subType whose base type is a comPzRingType and   *)
-(*                           whose predicate's is a subringClosed             *)
+(*                           whose predicate is a subringClosed               *)
 (*                                                                            *)
 (*  * ComNzRing (commutative NzRings):                                        *)
-(* [SubZmodule_isSubComNzRing of R by <:] ==                                  *)
 (* [SubChoice_isSubComNzRing of R by <:] == comNzRingType mixin for a         *)
 (*                           subType whose base type is a comNzRingType and   *)
-(*                           whose predicate's is a subringClosed             *)
+(*                           whose predicate is a subringClosed               *)
 (*                                                                            *)
 (*  * LSemiModule (semimodule with left multiplication by external scalars).  *)
 (*                 a *: v == v scaled by a, when v is in an LSemiModule V and *)
@@ -3203,29 +3199,9 @@ HB.structure Definition SubComNzSemiRing (R : nzSemiRingType) S :=
 HB.structure Definition SubPzRing (R : pzRingType) (S : pred R) :=
   { U of SubPzSemiRing R S U & PzRing U & isSubZmodule R S U }.
 
-HB.factory Record SubZmodule_isSubPzRing (R : pzRingType) S U
-    & SubZmodule R S U := {
-  subring_closed_subproof : subring_closed S
-}.
-
-HB.builders Context R S U & SubZmodule_isSubPzRing R S U.
-HB.instance Definition _ := SubNmodule_isSubPzSemiRing.Build R S U
-  (smulr_closedM (subring_closedM subring_closed_subproof)).
-HB.end.
-
 #[short(type="subNzRingType")]
 HB.structure Definition SubNzRing (R : nzRingType) (S : pred R) :=
   { U of SubNzSemiRing R S U & NzRing U & isSubBaseAddUMagma R S U }.
-
-HB.factory Record SubZmodule_isSubNzRing (R : nzRingType) S U
-    & SubZmodule R S U := {
-  subring_closed_subproof : subring_closed S
-}.
-
-HB.builders Context R S U & SubZmodule_isSubNzRing R S U.
-HB.instance Definition _ := SubNmodule_isSubNzSemiRing.Build R S U
-  (smulr_closedM (subring_closedM subring_closed_subproof)).
-HB.end.
 
 #[short(type="subComPzRingType")]
 HB.structure Definition SubComPzRing (R : pzRingType) S :=
@@ -3416,8 +3392,8 @@ HB.factory Record SubChoice_isSubPzRing (R : pzRingType) S U
 HB.builders Context R S U & SubChoice_isSubPzRing R S U.
 HB.instance Definition _ := SubChoice_isSubZmodule.Build R S U
   (subring_closedB subring_closed_subproof).
-HB.instance Definition _ := SubZmodule_isSubPzRing.Build R S U
-  subring_closed_subproof.
+HB.instance Definition _ := SubNmodule_isSubPzSemiRing.Build R S U
+  (smulr_closedM (subring_closedM subring_closed_subproof)).
 HB.end.
 
 HB.factory Record SubChoice_isSubNzRing (R : nzRingType) S U
@@ -3576,6 +3552,11 @@ Notation "[ 'SubNmodule_isSubNzSemiRing' 'of' U 'by' <: ]" :=
   (SubNmodule_isSubNzSemiRing.Build _ _ U (@rpred1M _ _))
   (format "[ 'SubNmodule_isSubNzSemiRing'  'of'  U  'by'  <: ]")
   : form_scope.
+Notation "[ 'SubNmodule_isSubPzSemiRing' 'of' U 'by' <: ]" :=
+  (SubNmodule_isSubPzSemiRing.Build _ _ U (@rpred1M _ _))
+  (format "[ 'SubNmodule_isSubPzSemiRing'  'of'  U  'by'  <: ]")
+  : form_scope.
+
 Notation "[ 'SubChoice_isSubNzSemiRing' 'of' U 'by' <: ]" :=
   (SubChoice_isSubNzSemiRing.Build _ _ U (semiringClosedP _))
   (format "[ 'SubChoice_isSubNzSemiRing'  'of'  U  'by'  <: ]")
@@ -3588,13 +3569,25 @@ Notation "[ 'SubChoice_isSubComNzSemiRing' 'of' U 'by' <: ]" :=
   (SubChoice_isSubComNzSemiRing.Build _ _ U (semiringClosedP _))
   (format "[ 'SubChoice_isSubComNzSemiRing'  'of'  U  'by'  <: ]")
   : form_scope.
-Notation "[ 'SubZmodule_isSubNzRing' 'of' U 'by' <: ]" :=
-  (SubZmodule_isSubNzRing.Build _ _ U (subringClosedP _))
-  (format "[ 'SubZmodule_isSubNzRing'  'of'  U  'by'  <: ]")
+Notation "[ 'SubChoice_isSubComPzSemiRing' 'of' U 'by' <: ]" :=
+  (SubChoice_isSubComPzSemiRing.Build _ _ U (semiringClosedP _))
+  (format "[ 'SubChoice_isSubComPzSemiRing'  'of'  U  'by'  <: ]")
+  : form_scope.
+Notation "[ 'SubChoice_isSubPzSemiRing' 'of' U 'by' <: ]" :=
+  (SubChoice_isSubPzSemiRing.Build _ _ U (subringClosedP _))
+  (format "[ 'SubChoice_isSubPzSemiRing'  'of'  U  'by'  <: ]")
+  : form_scope.
+Notation "[ 'SubChoice_isSubPzRing' 'of' U 'by' <: ]" :=
+  (SubChoice_isSubPzRing.Build _ _ U (subringClosedP _))
+  (format "[ 'SubChoice_isSubPzRing'  'of'  U  'by'  <: ]")
   : form_scope.
 Notation "[ 'SubChoice_isSubNzRing' 'of' U 'by' <: ]" :=
   (SubChoice_isSubNzRing.Build _ _ U (subringClosedP _))
   (format "[ 'SubChoice_isSubNzRing'  'of'  U  'by'  <: ]")
+  : form_scope.
+Notation "[ 'SubChoice_isSubComPzRing' 'of' U 'by' <: ]" :=
+  (SubChoice_isSubComPzRing.Build _ _ U (subringClosedP _))
+  (format "[ 'SubChoice_isSubComPzRing'  'of'  U  'by'  <: ]")
   : form_scope.
 Notation "[ 'SubChoice_isSubComNzRing' 'of' U 'by' <: ]" :=
   (SubChoice_isSubComNzRing.Build _ _ U (subringClosedP _))
--- a/algebra/algebraic_hierarchy/rings_modules_and_algebras.v
+++ b/algebra/algebraic_hierarchy/rings_modules_and_algebras.v
@@ -3576,6 +3576,11 @@ Notation "[ 'SubNmodule_isSubNzSemiRing' 'of' U 'by' <: ]" :=
   (SubNmodule_isSubNzSemiRing.Build _ _ U (@rpred1M _ _))
   (format "[ 'SubNmodule_isSubNzSemiRing'  'of'  U  'by'  <: ]")
   : form_scope.
+Notation "[ 'SubNmodule_isSubPzSemiRing' 'of' U 'by' <: ]" :=
+  (SubNmodule_isSubPzSemiRing.Build _ _ U (@rpred1M _ _))
+  (format "[ 'SubNmodule_isSubPzSemiRing'  'of'  U  'by'  <: ]")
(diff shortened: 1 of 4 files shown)


proux01:
Apparently, the following are also documented but missing
* `SubNmodule_isSubPzSemiRing`
* `SubZmodule_isSubPzRing`
* `SubNmodule_isSubComPzSemiRing`
* `SubNmodule_isSubComNzSemiRing` (not even the factory)
* `SubZmodule_isSubComPzRing` (not even the factory)
* `SubZmodule_isSubComNzRing` (not even the factory)

Maybe we could also fix them while at it

affeldt-aist:
There is also an issue with `SubPzRing_isSubComPzRing` whose notation is not available in 2.5.0, it is available in master:
```coq
Notation "[ 'SubPzRing_isSubComPzRing' 'of' U 'by' <: ]" :=
  (SubSemiRing_isSubComSemiRing.Build _ _ U)
  (format "[ 'SubPzRing_isSubComPzRing'  'of'  U  'by'  <: ]")
  : form_scope.
```
where `GRing.SubPzRing_isSubComPzRing.Build` is not available anymore.

In other words, we can't use neither the notation nor `GRing.SubPzRing_isSubComPzRing.Build` with 2.5.0 and 2.6.0 because it is not compatible with one or the other. 

This blocks the following MathComp-Analysis PR https://github.com/math-comp/analysis/pull/1949

affeldt-aist:
> Apparently, the following are also documented but missing
> 
>     * `SubNmodule_isSubPzSemiRing`
> 
>     * `SubZmodule_isSubPzRing`
> 
>     * `SubNmodule_isSubComPzSemiRing`
> 
>     * `SubNmodule_isSubComNzSemiRing` (not even the factory)
> 
>     * `SubZmodule_isSubComPzRing` (not even the factory)
> 
>     * `SubZmodule_isSubComNzRing` (not even the factory)
> 
> 
> Maybe we could also fix them while at it

We also fixed `SubNmodule_isSubPzSemiRing` but `SubZmodule_isSubPzRing` is ok

The following are documented but nothing is defined:
- `SubNmodule_isSubComPzSemiRing`
- `SubNmodule_isSubComNzSemiRing`
- `SubZmodule_isSubComPzRing`
- `SubZmodule_isSubComNzRing`

affeldt-aist:
Can we merge the missing notations and keep an issue for the missing factory?
That splits the work in two.

proux01:
Why not, but it seems just as easy to add them now, while we are at it.

affeldt-aist:
in fact, `SubNmodule_isSubComPzSemiRing` is only cited in the doc, not
specified, so we do not see right away what it is supposed to be

on the other hand, `SubNmodule_isSubPzSemiRing` is also only cited, not specified, but it
is implemented (using `mulr_closed`)

so I took `SubNmodule_isSubPzSemiRing` as a model for the factory,
and the builders is just about calling `SubNmodule_isSubPzSemiRing`
and proving that the multiplication is commutative on the subtype

I updated the doc accordingly

does it look ok to you?

if that makes sense, I will continue with the other missing doc
and factories, i.e.:
- `SubNmodule_isSubComNzSemiRing`
- `SubZmodule_isSubComPzRing`
- `SubZmodule_isSubComNzRing`

algebra/algebraic_hierarchy/rings_modules_and_algebras.v
```diff
@@ -3161,6 +3165,28 @@ Proof. by split=> [|x y] /=; rewrite !SubK. Qed.
...
+
+HB.instance Definition _ := isMulClosed.Build R S mulr_closed_subproof.
+
+Let inU v Sv : U := Sub v Sv.
+Let mulU (u1 u2 : U) := inU (rpredM _ _ (valP u1) (valP u2)).
+
+Lemma mulrC : commutative mulU.
+Proof. by move=> x y; apply: val_inj; rewrite !SubK mulrC. Qed.
+
+HB.instance Definition _ := SemiRing_hasCommutativeMul.Build U mulrC.
```
pi8027: I think you can do this in one or two lines using `SubSemiRing_isSubComSemiRing` here.

pi8027 (commented):
Note that I plan to remove `pz` after the release (#1487). We need to review/redo the documentation again anyway.

algebra/algebraic_hierarchy/rings_modules_and_algebras.v
pi8027: (I even prefer to add this factory after removing `pz`.)

affeldt-aist:
I think that this PR is now complete: missing notations, factories, and documentation have been added @proux01

affeldt-aist:
> Note that I plan to remove `pz` after the release (#1487). We need to review/redo the documentation again anyway.

Understood.

algebra/algebraic_hierarchy/rings_modules_and_algebras.v
```diff
@@ -3217,6 +3247,17 @@ HB.end.
 HB.structure Definition SubNzRing (R : nzRingType) (S : pred R) :=
   { U of SubNzSemiRing R S U & NzRing U & isSubBaseAddUMagma R S U }.
 
+HB.factory Record SubZmodule_isSubComPzRing (R : comPzRingType) S U
```
pi8027: I guess this is redundant with something like SubNmodule_isSubComPzSemiring. (I cannot check it since I'm not with my laptop.）

algebra/algebraic_hierarchy/rings_modules_and_algebras.v
```diff
@@ -3227,6 +3268,17 @@ HB.instance Definition _ := SubNmodule_isSubNzSemiRing.Build R S U
   (smulr_closedM (subring_closedM subring_closed_subproof)).
 HB.end.
 
+HB.factory Record SubZmodule_isSubComNzRing (R : comNzRingType) S U
```
pi8027: Same here. This should be redundant with the semi one.

algebra/algebraic_hierarchy/rings_modules_and_algebras.v
affeldt-aist: I can't say for sure:
`SubNmodule_isSubComPzSemiRing` is using `mulr_closed` and
`SubZmodule_isSubComPzRing` is using `subring_closed`
but I must admit that I don't understand everything here

affeldt-aist: `subring_closed` looks wrong to me now, maybe it should have been `mulr_closed`, in which case that's indeed redundant

pi8027: I guess there is a coercion from the latter to the former.

algebra/algebraic_hierarchy/rings_modules_and_algebras.v
affeldt-aist: So redundant, so remove these two factories? You certainly know better than me.

pi8027: If you actually use this factory, you can try to see if the general one can replace the specific one or not. I cannot confirm it before Monday.

affeldt-aist: I don't have a use for them, let's enjoy the week-end!

algebra/algebraic_hierarchy/rings_modules_and_algebras.v
```diff
@@ -3217,6 +3247,17 @@ HB.end.
...
 
+HB.factory Record SubZmodule_isSubComPzRing (R : comPzRingType) S U
+    & SubZmodule R S U := {
+  subring_closed_subproof : subring_closed S
+}.
+
+HB.builders Context R S U & SubZmodule_isSubComPzRing R S U.
+HB.instance Definition _ :=
+  SubZmodule_isSubPzRing.Build R S U subring_closed_subproof.
+HB.instance Definition _ := SubSemiRing_isSubComSemiRing.Build R S U.
```
pi8027: The following works. I think this implies that `SubZmodule_isSubComPzRing` is redundant with `SubNmodule_isSubComPzSemiRing`. (Note that `ClosedExports` is exported by default.)
```suggestion
Import ClosedExports.
HB.instance Definition _ :=
  SubNmodule_isSubComPzSemiRing.Build R S U subring_closed_subproof.
```

algebra/algebraic_hierarchy/rings_modules_and_algebras.v
```diff
@@ -3227,6 +3268,17 @@ HB.instance Definition _ := SubNmodule_isSubNzSemiRing.Build R S U
...
 
+HB.factory Record SubZmodule_isSubComNzRing (R : comNzRingType) S U
+    & SubZmodule R S U := {
+  subring_closed_subproof : subring_closed S
+}.
+
+HB.builders Context R S U & SubZmodule_isSubComNzRing R S U.
+HB.instance Definition _ :=
+  SubZmodule_isSubNzRing.Build R S U subring_closed_subproof.
+HB.instance Definition _ := SubSemiRing_isSubComSemiRing.Build R S U.
```
pi8027: Same here:
```suggestion
Import ClosedExports.
HB.instance Definition _ :=
  SubNmodule_isSubComNzSemiRing.Build R S U subring_closed_subproof.
```

algebra/algebraic_hierarchy/rings_modules_and_algebras.v
```diff
@@ -3195,6 +3203,28 @@ Proof. by move=> x y; apply: val_inj; rewrite !rmorphM mulrC. Qed.
...
+HB.factory Record SubNmodule_isSubComNzSemiRing (R : comNzSemiRingType) S U
+    & SubNmodule R S U := {
+  mulr_closed_subproof : mulr_closed S
+}.
+
+HB.builders Context R S U & SubNmodule_isSubComNzSemiRing R S U.
+HB.instance Definition _ :=
+  SubNmodule_isSubNzSemiRing.Build R S U mulr_closed_subproof.
+HB.instance Definition _ := SubSemiRing_isSubComSemiRing.Build R S U.
+HB.end.
```
pi8027: Each of these two factories has only two builders, which take only 3 lines. So, an `HB.instance` declaration that uses one of these factories can always be replaced with two `HB.instance` declarations. Is it still worth adding them?

In other words, if we start adding all the possible combinations like them, it would blow up the amount of code. We probably need to find a sweet spot.

affeldt-aist: I am not sure for a general rule.
But, as far as this PR is concerned:
- we were led to these fixes because we needed a related feature (which revealed an inconsistency between 2.5.0 and 2.6.0 to be fixed by #1588)
- somebody already thought of the notations and of the factories since he/she started to right documentation for them

so maybe merging these (maybe not the redundant ones) is a safe bet.

As for the general, it sounds like a topic for the meeting.

pi8027: > somebody already thought of the notations and of the factories since he/she started to right documentation for them

I guess the thing is rather that we didn't put enough effort into revising these factories when introducing some generalisations like "pz" and "semi" structures.

pi8027: Indeed, this discrepancy between the documentation and the implementation should be fixed. But, I prefer to update the documentation based on the implementation, not the other way around.

proux01:
In any case, we should make the doc consistent with the code, which implies either:
* removing the doc (and document the removal in the changelog)
* reintroducing the code

I don't have any strong opinion on which option to choose. I agree some factories are redundant. Redundancy not always means we shouldn't have the redundant thing (e.g., `ltn_leq_trans` that is a special case of `leq_trans` but hard to dicsover). However, here HB.howto should find the factories just as well so probably no reason to keep redundancy.

pi8027:
FTR, `SubZmodule_isSubPzRing` and `SubZmodule_isSubNzRing` seem redundant with `SubNmodule_isSubPzSemiRing` and `SubNmodule_isSubNzSemiRing`, respectively.

affeldt-aist:
This PR tries to puts the implementation on par with documentation by
adding (+) and removing (-) things as follows:
```
                              | doc | implem | notation | in changelog <= 2.5.0
------------------------------|-----|--------|----------|----------------------
uncontroversial               |
------------------------------|-----|--------|----------|----------------------
SubNmodule_isSubPzSemiRing    |     |        | +        | yes
SubChoice_isSubPzRing         |     |        | +        | yes
SubChoice_isSubComPzRing      |     |        | +        | yes
SubChoice_isSubPzSemiRing     |     |        | +        | yes
SubChoice_isSubComPzSemiRing  |     |        | +        | yes
SubZmodule_isSubComPzRing     | -   |        |          | no
SubZmodule_isSubComNzRing     | -   |        |          | no
SubZmodule_isSubPzRing        |     |        | +        | no
------------------------------|-----|--------|----------|-----------------------
remove                        |
------------------------------|-----|--------|----------|-----------------------
SubNmodule_isSubNzSemiRing    | +   |        |          | yes
SubNmodule_isSubComPzSemiRing | +   | +      |          | no
SubNmodule_isSubComNzSemiRing | +   | +      |          | no
```
I think that the upper part is uncontroversial because the things are in the changelog deserve notations.
I am less sure about the debatable lower part (it was only partially documented...).

As for the factories that @pi8027 identified as redundant, they are already removed.

pi8027:
After taking a close look at the code, I changed my mind a little bit. `SubNmodule_isSubNzSemiRing` is used by `SubNmodule_isSubComNzSemiRing`. It's probably better to keep them for now so that we can merge this PR quickly.

On the other hand, `SubZmodule_isSubPzRing` (and `SubZmodule_isSubNzRing`) are duplicates of the semi ones. I can prepare a patch for removing them.

algebra/algebraic_hierarchy/rings_modules_and_algebras.v
```diff
@@ -223,16 +223,20 @@ From mathcomp Require Import nmodule.
 (*                           products (1 and x * y in S for x, y in S)        *)
 (*      semiring_closed S <-> collective predicate S is closed under semiring *)
 (*                           operations (0, 1, x + y and x * y in S)          *)
-(* [SubNmodule_isSubPzSemiRing of R by <:] ==                                 *)
+(* [SubNmodule_isSubPzSemiRing of R by <:] == pzSemiRingType mixin for a      *)
+(*                           subType whose base type is a pzSemiRingType and  *)
```
pi8027: To be precise, I believe that `subType` here should be `subNmodType`.

algebra/algebraic_hierarchy/rings_modules_and_algebras.v
```diff
@@ -223,16 +223,20 @@ From mathcomp Require Import nmodule.
...
+(*                           whose predicate is a mulr_closed                 *)
```
pi8027: And `mulr_closed` here should probably be `mulrClosed`. The former is a predicate (over predicates), and the latter is a structure.

algebra/algebraic_hierarchy/rings_modules_and_algebras.v
```diff
@@ -244,40 +248,40 @@ From mathcomp Require Import nmodule.
...
-(*                           type is a nzRingType and whose predicate's is a  *)
-(*                           subringClosed                                    *)
+(*                           base type is a nzRingType and whose predicate is *)
+(*                           a subringClosed                                  *)
 (*                                                                            *)
 (*  * ComPzSemiRing (commutative PzSemiRings):                                *)
-(* [SubNmodule_isSubComPzSemiRing of R by <:] ==                              *)
+(* [SubNmodule_isSubComPzSemiRing of R by <:] == comPzSemiRingType mixin for  *)
+(*                          a subType whose base type is a comPzSemiRingType  *)
+(*                          and whose predicate is a ring_closed              *)
```
pi8027: Similarly, `subType` should be `subNmodType`, and `ring_closed` should be `subringClosed`.

pi8027 (commented):
I spotted a few documentation issues. It looks pretty hard to make them very accurate and keep them up to date. (Also, it's not worth the effort IMO.)

pi8027:
Perhaps it's reasonable to say something like "the following factories are provided for building a `(sub)PzSemiRingType` instance through the subtype mechanism" and list only the notations, but remove their exact meaning (the things that come after `==`) to avoid these documentation issues.

affeldt-aist:
> Perhaps it's reasonable to say something like "the following factories are provided for building a `(sub)PzSemiRingType` instance through the subtype mechanism" and list only the notations, but remove their exact meaning (the things that come after `==`) to avoid these documentation issues.

Now that it has been written and checked, I would keep it because, thanks to them, an ignorant user like me does not need to reconstruct the naming scheme in one's head, which I find pretty hard.

affeldt-aist:
I think that I have carefully applied all comments, redundant things spotted by Sakaguchi-san have been removed, and the documentation is on par with the code.

algebra/algebraic_hierarchy/ssralg.v
```diff
@@ -363,8 +363,19 @@ End SubNmodule_isSubSemiRing.
...
+(* NB: this factory was in rings_modules_and_algebras.v but was deemed redundant with SubNmodule_isSubNzSemiRing, yet still used to implemented the deprecation for SubZmodule_isSubRing.Build, it could be removed together with the latter *)
+HB.factory Record SubZmodule_isSubNzRing (R : nzRingType) S U
+    & SubZmodule R S U := {
+  subring_closed_subproof : subring_closed S
+}.
+
+HB.builders Context R S U & SubZmodule_isSubNzRing R S U.
+HB.instance Definition _ := SubNmodule_isSubNzSemiRing.Build R S U
+  (smulr_closedM (subring_closedM subring_closed_subproof)).
+HB.end.
```
pi8027: I think we can properly deprecate this factory as follows (not tested):
```suggestion
Module SubZmodule_isSubNzRing.
#[deprecated(since="mathcomp 2.6.0", use=SubNmodule_isSubNzSemiRing.Build)]
Notation Build R S U := (SubNmodule_isSubNzSemiRing.Build R S U) (only parsing).
End SubZmodule_isSubNzRing.
```

algebra/algebraic_hierarchy/ssralg.v
```diff
@@ -363,8 +363,19 @@ End SubNmodule_isSubSemiRing.
...
+
 Module SubZmodule_isSubRing.
-#[deprecated(since="mathcomp 2.4.0", use=SubZmodule_isSubNzRing.Build)]
+#[deprecated(since="mathcomp 2.4.0", use=SubNmodule_isSubNzSemiRing.Build)]
 Notation Build R S U := (SubZmodule_isSubNzRing.Build R S U) (only parsing).
```
pi8027: And I think this works (not tested):
```suggestion
Notation Build R S U := (SubNmodule_isSubNzSemiRing.Build R S U) (only parsing).
```

algebra/algebraic_hierarchy/rings_modules_and_algebras.v
```diff
@@ -3588,14 +3599,26 @@ Notation "[ 'SubChoice_isSubComNzSemiRing' 'of' U 'by' <: ]" :=
   (SubChoice_isSubComNzSemiRing.Build _ _ U (semiringClosedP _))
   (format "[ 'SubChoice_isSubComNzSemiRing'  'of'  U  'by'  <: ]")
   : form_scope.
-Notation "[ 'SubZmodule_isSubNzRing' 'of' U 'by' <: ]" :=
-  (SubZmodule_isSubNzRing.Build _ _ U (subringClosedP _))
-  (format "[ 'SubZmodule_isSubNzRing'  'of'  U  'by'  <: ]")
```
pi8027, line 3593: We may want to deprecate this notation (and move to `ssralg.v`) too.

algebra/algebraic_hierarchy/rings_modules_and_algebras.v
pi8027: @affeldt-aist If we don't have a strong reason to add them, I would rather remove them (from documentation too), open an issue, and revisit it after removing the `pz` prefix. What do you think?

algebra/algebraic_hierarchy/ssralg.v
```diff
@@ -363,9 +363,19 @@ End SubNmodule_isSubSemiRing.
 Notation SubNmodule_isSubSemiRing R S U :=
   (SubNmodule_isSubNzSemiRing R S U) (only parsing).
 
+Module SubZmodule_isSubNzRing.
+#[deprecated(since="mathcomp 2.6.0", use=SubNmodule_isSubNzSemiRing.Build)]
+Notation Build R S U := (SubNmodule_isSubNzSemiRing.Build R S U) (only parsing).
+End SubZmodule_isSubNzRing.
+
+Notation "[ 'SubZmodule_isSubNzRing' 'of' U 'by' <: ]" :=
```
pi8027:
```suggestion
#[deprecated(since="mathcomp 2.6.0",
             note="Use [ SubNmodule_isSubNzSemiRing of U by <: ] instead.")]
Notation "[ 'SubZmodule_isSubNzRing' 'of' U 'by' <: ]" :=
```
We also need to move this to the `SubExports` module below; otherwise, it won't be exported correctly.

affeldt-aist: I'm starting to understand why your computer display is set vertically...

algebra/algebraic_hierarchy/ssralg.v
```diff
@@ -363,9 +363,21 @@ End SubNmodule_isSubSemiRing.
...
+#[deprecated(since="mathcomp 2.6.0", use=SubNmodule_isSubNzSemiRing.Build)]
+Notation Build R S U := (SubNmodule_isSubNzSemiRing.Build R S U) (only parsing).
+End SubZmodule_isSubNzRing.
+
+#[deprecated(since="mathcomp 2.6.0",
+             note="Use [ SubNmodule_isSubNzSemiRing of U by <: ] instead.")]
+Notation "[ 'SubZmodule_isSubNzRing' 'of' U 'by' <: ]" :=
+  (SubNmodule_isSubNzSemiRing.Build _ _ U (subringClosedP _))
+  (format "[ 'SubZmodule_isSubNzRing'  'of'  U  'by'  <: ]")
+  : form_scope.
```
pi8027: - The deprecation of the `SubZmodule_isSubNzRing` factory is only halfway done. (Sorry, my bad.)
- Since the notation for `SubZmodule_isSubNzRing` has been added to `SubExports`, we can remove it here.

So:
```suggestion
End SubZmodule_isSubNzRing.

#[deprecated(since="mathcomp 2.6.0", use=SubNmodule_isSubNzSemiRing)]
Notation SubZmodule_isSubNzRing R S U :=
  (SubNmodule_isSubNzSemiRing R S U) (only parsing).
```

affeldt-aist:
I trust you, I am completely lost now.

pi8027:
I'm a little bit lost too, but not completely. Do you prefer me to push my suggestions (as separate commits) directly to this branch?

affeldt-aist:
For now, yes, because I need to go to bed and I am out of office tomorrow :-(

pi8027:
So, I made two changes:
- The first commit (26868519afb6508c1834fd2cf6392857bbb46db8) fixes the removal and deprecation of `SubZmodule_isSubPzRing` and `SubZmodule_isSubNzRing`.
- The second commit (306f9eff45245fdd4b0a7d404726bb6cf3594d02) removes `SubNmodule_isSubComPzSemiRing` and `SubNmodule_isSubComNzSemiRing` (the factories you added). If there is a good reason to add them (which I still don't see), I'm fine with it. See https://github.com/math-comp/math-comp/pull/1586#discussion_r3168143518.

affeldt-aist:
I am fine with the removal of the factories. I only added them to match the documentation, we do not really add a use for them.

doc/changelog/01-added/1586-notations.md
```diff
@@ -0,0 +1,9 @@
+- in `rings_modules_and_algebras.v`
+  + Notations for
+    `SubNmodule_isSubPzSemiRing`
+    `SubZmodule_isSubPzRing`
```
pi8027: I should have removed this line (`SubZmodule_isSubPzRing`) too.

pi8027 (approved):
I think we are good to go modulo the issue below and CI.
@affeldt-aist Can you rebase and squash the commits to refresh CI?

pi8027:
Since I authored a part of the PR, it would be better if someone could double-check.

proux01 (approved):
LGTM, FTR here is the table I built while reviewing
```
factory                       | ref    | doc | factory | notation | changelog
------------------------------+--------+-----+---------+----------+----------
SubChoice_isSubPzSemiRing     | master | 1   | 1       | 0        | 1
                              | #1586  | 1   | 1       | 1        | +notation
SubChoice_isSubComPzSemiRing  | master | 1   | 1       | 0        | 1
                              | #1586  | 1   | 1       | 1        | +notation
SubChoice_isSubPzRing         | master | 1   | 1       | 0        | 1
                              | #1586  | 1   | 1       | 1        | +notation
SubChoice_isSubComPzRing      | master | 1   | 1       | 0        | 1
                              | #1586  | 1   | 1       | 1        | +notation
SubNmodule_isSubPzSemiRing    | master | 1   | 1       | 0        | 1
                              | #1586  | 1   | 1       | 1        | +notation
SubNmodule_isSubComPzSemiRing | master | 1   | 0       | 0        | 0
                              | #1586  | 0   | 0       | 0        |
SubNmodule_isSubComNzSemiRing | master | 1   | 0       | 0        | 0
                              | #1586  | 0   | 0       | 0        |
SubZmodule_isSubPzRing        | master | 1   | 1       | 0        | 1
                              | #1586  | 0   | 0       | 0        | -factory
SubZmodule_isSubNzRing        | master | 1   | 1       | 1        | 1
                              | #1586  | 0   | 0       | 0        | -factory
SubZmodule_isSubComNzRing     | master | 1   | 0       | 0        | 0
                              | #1586  | 0   | 0       | 0        |
```

PR #1606, opened 2026-06-03
Revert " Tensor formalization"

CohenCyril:
Reverts math-comp/math-comp#1535

Dear @hoheinzollern, @proux01 explicitly stated that he did not review the `tensor.v` file.
Please allow a bit more time to do it and reopen your PR after this revert.

PR #1610, opened 2026-06-04
fix: remove tensor_scope, move tensor to ring_scope, drop incompatible covariant notation

hoheinzollern:

This PR fixes the issues remained unaddressed from #1535 and discussed in #1606 with @pi8027:
- removes `tensor_scope` in favor of `ring_scope` and drops all confilcting notation in the library
- the changelog is updated in place of the old entry instead of creating a new entry
- future PR #1613 will reintroduce the notation

##### Minimal TODO list

### DIFF
--- a/algebra/tensor.v
+++ b/algebra/tensor.v
@@ -29,37 +29,22 @@ From mathcomp Require Import interval interval_inference numdomain.
 (*                          'oT[R]_([tuple of d1%:posnat :: .. :: dl%:posnat])*)
 (*           'sT[R], 'sT == scalar tensors (no contravariant nor covariant    *)
 (*                          dimensions), i.e. 'T[R]_([tuple], [tuple])        *)
-(*                  t^^i == the tensor obtained by fixing the first           *)
+(*            nindex t i == the tensor obtained by fixing the first           *)
 (*                          contravariant dimension of t to i                 *)
-(*                  t`_i == the tensor obtained by fixing the first           *)
-(*                          covariant dimension of t to i                     *)
-(* \tensor^^(i < n) Expr(i) ==                                                *)
-(*                         the 'T_([tuple of n :: _], _) tensor t such that   *)
-(*                         t^^i = Expr(i) with i : 'I_n, the < n bound can    *)
-(*                         usually be omitted                                 *)
-(* \tensor`_(j < n) Expr(j) ==                                                *)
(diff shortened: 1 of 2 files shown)


algebra/tensor.v
```diff
@@ -848,7 +843,7 @@ Qed.
 End TensorProduct.
 
 Notation "*t%R" := mults : ring_scope.
```
pi8027:
```suggestion
Notation "*t%R" := mults : function_scope.
```

algebra/tensor.v
```diff
@@ -169,8 +166,6 @@ HB.instance Definition _ := [isNew for tensor_val].
 
 End TensorDef.
 
-Bind Scope tensor_scope with tensor.
-
```
pi8027, line 173:
```suggestion
Bind Scope ring_scope with tensor.

```

algebra/tensor.v
pi8027: It looks like `ring_scope` is already open. So, I suggest removing this line:
```suggestion
```
And idem for the same lines below.

pi8027 (approved):
I confirm that this PR does what I suggested. On the other hand, I have a few minor suggestions below (note tested), and I have no opinion on what the right notation for `oindex` is.

hoheinzollern:
Still fails downstream in due to the use of `\_` in `analysis`. @affeldt-aist could you suggest a solution that is compatible with the use in `analysis` for the subscript notation?

hoheinzollern:
After today's meeting, we decided to drop the covariant notation as it's currently not used, and solve the issue in a separate PR that will land in a future release. @CohenCyril @pi8027

hoheinzollern:
real-closed has a clash on the `*t` notation, I will drop that too since we are not using it.
(ideally, we should fix the scope for the `*t` notation on `mxtens.v`)

hoheinzollern:
Finally, abel has a conflict with the `^^` notation. I'm reverting that too, so later we can discuss how to introduce proper notation for tensors. It might require patching downstream, or a totally different approach.

algebra/tensor.v
pi8027: I think these notations should be declared in `type_scope`.

hoheinzollern: Done. I'll wait for the CI to complete, then I'll squash to a single commit.

hoheinzollern:
Should be ready for merging now.

PR #1621, opened 2026-06-29
Compile without warning on Rocq 9.2

proux01:


CohenCyril (approved):
Note we need to open issues for each warning to fix it without silencing it as soon as we drop the appropriate versions of Rocq.

proux01:
Ideally yes, in practice I find it more convenient to do a grep and find back the comment in `_CoqProject` whenever we drop a Rocq version.

PR #1622, opened 2026-06-29
Drop support for Rocq 9.0

proux01:
~Depends on https://github.com/math-comp/math-comp/pull/1621~ (merged)

#### Overlay (to be merged before the current PR)

* https://github.com/math-comp/analysis/pull/2085
* https://github.com/affeldt-aist/infotheo/pull/230


proux01:
Apparently this doesn't simplify much, but this means we should now consistently have the new behavior of the matching of rewrite, which is probably good to take.

proux01:
CI "green":
* failure of gitlab CI is unrelated (looks like the dev image is not up to date for some reason)
* 9.0 failures are expected pull_request_target (since we are removing 9.0 support)
* ssprove is known to be broken
* rocqnavi/mathcomp-doc is known to be broken

So this can now be merged (modulo a quick rview) since we agreed on the principle of dropping 9.0 support during our last meeting.

pi8027 (approved):

Issue #1624, opened 2026-07-13
TODOs `archimedean.v`


pi8027:
- Add a new lemma `floor_default: x \isn't real_num -> floor x = 0` https://github.com/math-comp/math-comp/pull/1510#discussion_r3021540989
- Change the statement of `truncEfloor` to ``if 0 <= x then `|floor x|%N else 0%N`` https://github.com/math-comp/math-comp/pull/1510#discussion_r3347694493

PR #1625, opened 2026-07-14
algebra(archimedean): add to_real_ceil_floor lemma

SNAPKITTYWEST:
Closes part of #1624.

Implements the TODO stub at line 607:

```coq
(* TODO: *)
(* Lemma to_real_ceil_floor x : ceil x = floor x + (to_real x \isn't a int_num). *)
```

The proof follows directly from `real_ceil_floor` with `num_real`. The lemma name uses `num_to_real` (the current accessor name) rather than the deprecated `to_real`.


SNAPKITTYWEST:
Hi! I noticed this PR was closed without feedback. I'd appreciate understanding the reason so I can improve future contributions.

If the implementation doesn't match the project's preferred style or if the TODO is being addressed another way, I'd be grateful for a brief explanation. Thanks for your time.

PR #1635, opened 2026-07-29
Handle new Rocq 9.3 warnings

proux01:
And test warning-free compilation on Rocq 9.3+rc1

This should only silence warnings, not introduce any semantic change.

Close #1599 (Edit by Kazuhiko)


proux01:
CI "green" (the experimental-reals and infotheo failures are already there on master), this can be merged.

field/finfield.v
```diff
@@ -392,7 +392,7 @@ Implicit Types (a b : F) (x y : L) (K E : {subfield L}).
 
 Let galL K : galois K {:L}.
 Proof.
-without loss {K} ->: K / K = 1%AS.
+without loss ->: K / K = 1%AS.
```
pi8027, line 395: This looks like the same change as #1599 and thus seems related to rocq-prover/rocq#21987. Is it really a good idea to remove `{K}` here? If we write `-> {K}` instead of `{K} ->`, does it also address the issue? (Indeed, `clear` cannot be performed before `rewrite` here. So `-> {K}` makes more sense to me, although I don't know what's happening internally.)

pi8027 (commented):
This PR needs a rebase. I have another comment below.

field/finfield.v
proux01: `without loss -> {K}` is not valid syntax (`Error: No binder or s-item allowed here: {K}` on Rocq 9.1) whereas `without loss {K} -> : K / K = 1%AS.` and `without loss -> : K / K = 1%AS.` produce the exact same result.

proux01:
@pi8027 rebased, CI "green" again (well, as green as it can currently be)

field/finfield.v
pi8027: Since `{K}` is there, I guess the intention behind the proof script was to clear `K` (in the second subgoal), and I'm not in favor of working around the issue by removing `{K}`. But, I confirm that the script has not been working "as intended" at least since Coq 8.11. So, it's probably not worth investigating it further.

@gares Do you have any idea?

pi8027: I understand almost nothing here, but I have the impression that `wlog` tactic first attempts `clear` (in `tacigens`) and then performs `rewrite` (in `tacipat pats`), and `clear` should probably be delayed to recover the behavior we want: https://github.com/rocq-prover/rocq/blob/258e98651f44f6d3f227e70f2bcb81f8d4642ba2/plugins/ssr/ssrfwd.ml#L384

proux01: I think we really shouldn't delay this PR for that since there is no semantic change anyway. This can be handled in a further PR. We can add a comment to remember the `{K}` was there but doing nothing for long if you want.

pi8027: I will open an issue (rather than adding comments) after merging this PR.

pi8027 (approved):

field/finfield.v
SkySkimmer: where's the issue?

PR #1640, opened 2026-08-25
Make individual package compilations less verbose

proux01:
These compilation flags are in _CoqProject but were forgotten in individual packages Makefiles. This only affects verbosity of the build when building on a per package basis.

In the long run, we should probably find a way to avoid all that flag duplication business.


proux01:
CI "green", this can be merged

pi8027 (approved):

PR #1560, opened 2026-03-18
Move `orderedzmod.v` from `algebra/numeric_hierarchy/` to `order/`

pi8027:
##### Motivation for this change

This PR does three things:
- adding exportation modules (`Algebra.Theory` and `Algebra.ClosedExports`) to `nmodule.v`,
- eliminating the dependency from `orderedzmod.v` to `rings_modules_and_algebras.v` (which was a part of #1548), and
- moving `orderedzmod.v` from `algebra/numeric_hierarchy/` to `order/`.

I'm open to splitting the PR into smaller pieces if the reviewers prefer doing so.

##### Dependencies

- #1551
- #1580

##### Minimal TODO list

- [x] added changelog entries with `doc/changelog/make-entry.sh`

- [x] added corresponding documentation in the headers
- [x] tried to abide by the [contribution guide](https://github.com/math-comp/math-comp/blob/master/CONTRIBUTING.md)
- [x] this PR contains an optimum number of meaningful commits

See [this Checklist](https://github.com/math-comp/math-comp/wiki/Checklist-for-creating-and-review-PRs) for details.

##### Automatic note to reviewers

Read [this Checklist](https://github.com/math-comp/math-comp/wiki/Checklist-for-creating-and-review-PRs).

Issue #1557, opened 2026-03-13
naming in `algebra`


affeldt-aist:
In algebra, there is now a subdirectory with a very short name: `alg`. I think that a longer name would be more informative and help navigating the files. Similarly, renamings such as `zmod_morphism` for `additive` are not very inviting; at the very last, shouldn't we try to have both?

affeldt-aist:
Looking at PR #1547, maybe the issue extends to file names.
It is clear that `all_alg.v` and `all_num.v` contain all that `alg` and `num` contain,
but for the end user it means that scripts will start with (potentially long) commands like `Require Import all_alg all_num all_xyz.`.
Wouldn't it be nicer if the `Require Import`s were more "readable"? Like in `Require Import Algebra Numeric XYZ.`

mkerjean:
My 2 cents on the second topic, which I mentioned to Reynald and may have led to this issue. As a new user to the library several years ago, it was incredibly frustrating to look for lemmas with `(_+_)` and to realise after several hours that I should have used `additive _`. I'm afraid that replacing it with `Zmorphism`, or the lowest morphism structure in the hierarchy that satisfies the equation, will make it even harder to find and to understand when it arrives as the result of a `Search`. Said otherwise, I understand the need to properly name morphisms according to the HB structure, but it would still be nice to be able to look for lemmas using the usual mathematical terminology.

pi8027:
> I dunno. Yet we are suffering here from a missing feature in Rocq which would be akin to python's `__init__.py`. Maybe it is worth a RFC? WDYT?

@CohenCyril IMO, this might be worth an RFC, but we shouldn't wait for such a feature to introduce `all_*.v`.

It would be nice to reach a consensus on the directory names soonish, since I won't introduce `all_*.v` before that.

CohenCyril:
@mkerjean the subdirectory `alg` is not strictly necessary, but it helps us remember that this used to be the contents of `ssralg` dispatched in several files, which you can still bulk import through `ssralg.v`: this is probably poorly documented as of now and a directory `README.md` could help.
Still I am not opposed to a full reorganization of the files and directories (as long as we can preserve backward compat across 2 releases). There are just too many files to keep a flat structure anymore, so subdirectories, even with another structuration are welcome I think.
What do you have in mind?

proux01:
Let's discuss it during next meeting https://github.com/math-comp/math-comp/wiki/TopicsNextMeeting

pi8027:
[...]

PR #1642, opened 2026-08-26
Add `algebraic_hierarchy.v` and `numeric_hierarchy.v`, and deprecate `ssralg.v` and `ssrnum.v`

### LINKED ISSUE #1505: Uniformize the name of the re-exportation files

pi8027:

This is the remaining work from #1582.

Closes #1505

Closes (Topic 1 of) #1557

The difference between `ssralg.v` and `algebraic_hierarchy.v`:
- [added] the latter reexports `countalg.v`, `finalg.v`, and `ring_quotient.v` (because this PR moves them to `algebraic_hierarchy/`)
- [removed] the latter does not reexport `nmodule.v` (because it belongs to the boot package)
- [removed] the latter does not contain the deprecated definitions from `ssralg.v`

The difference between `ssrnum.v` and `numeric_hierarchy.v`:
- [removed] the latter does not reexport `orderedzmod.v` (because #1560 moves it to the order package)
- [removed] the latter does not contain the deprecated definition `Num.ExtraDef.sqrtr`

##### Dependency

- #1560

##### Overlay

- https://github.com/math-comp/Abel/pull/108

##### Minimal TODO list


algebra/algebraic_hierarchy/algebraic_hierarchy.v
```diff
@@ -0,0 +1,18 @@
...
+From mathcomp Require Export divalg.
+From mathcomp Require Export decfield.
+From mathcomp Require Export countalg.
+From mathcomp Require Export finalg.
+
+(******************************************************************************)
+(*                            Ring-like structures                            *)
+(*                                                                            *)
+(* This file re-exports the contents of rings_modules_and_algebras.v,         *)
+(* divalg.v, decfield.v, countalg.v, finalg.v, and ring_quotient.v:           *)
```
proux01, line 12: according to the above:
```suggestion
(* divalg.v, decfield.v, countalg.v and finalg.v:           *)
```
but maybe it would be easier to say "the content of the algebraic_hierarchy directory". I observed that these header comments tend to become easily inaccurate with time.

algebra/algebraic_hierarchy/ssralg.v
```diff
@@ -1,22 +1,14 @@
 (* (c) Copyright 2006-2016 Microsoft Corporation and Inria.                  *)
 (* Distributed under the terms of CeCILL-B.                                  *)
+
+Attributes deprecated(since="mathcomp 2.7.0",
+  note="Use algebraic_hierarchy.v instead").
```
proux01, line 5: To mitigate the issue that this warning will show before all other deprecations:
```suggestion
Attributes deprecated(since="mathcomp 2.7.0",
  note="Use algebraic_hierarchy.v instead. We recommend handling all other deprecations from ssralg first.").
```

proux01:
>     * [added] the latter reexports `countalg.v`, `finalg.v`, and `ring_quotient.v` (because this PR moves them to `algebraic_hierarchy/`)

Do we really want to do that? These are maybe not of such general use but adding them could make the `Require` command significantly slower (haven't benchmarked though).
 
>     * [removed] the latter does not contain the deprecated definitions from `ssralg.v`

Still not a big fan of this mixing between file renaming and deprecations removal (it just makes the renaming harder for users). However, see my comments for mitigation recommendations.

algebra/algebraic_hierarchy/algebraic_hierarchy.v
pi8027: IMO, the idea is rather to use the header doc of `algebra_hierarchy.v` as the general description of the files in the directory (like `order.v`).

pi8027:
> > [added] the latter reexports `countalg.v`, `finalg.v`, and `ring_quotient.v` (because this PR moves them to `algebraic_hierarchy/`)
> 
> Do we really want to do that? These are maybe not of such general use but adding them could make the `Require` command significantly slower (haven't benchmarked though).

I suggest discussing this point in a meeting. Perhaps the general recommendation for users should be to import only what they need (most of the time only `rings_modules_and_algebras` and `divalg` I guess).

pi8027:
By the way, I don't think it's very useful to review this PR in detail before merging #1560.

algebra/algebraic_hierarchy/algebraic_hierarchy.v
proux01: (In addition, maybe we could have one line README.md in each directory saying something like "look at [order.v]")

proux01:
I was just reacting to

> pi8027 requested a review from proux01 20 hours ago

above (but maybe these are automatic, I never really mastered them).

algebra/ring_tactic.v
```diff
@@ -5,9 +5,9 @@ From elpi Require Import derive.std param2.
 From micromega_plugin Require Import NatDef.
 From Corelib Require Import IntDef.
 From micromega_plugin Require Import formula witness checker eval.
-From mathcomp Require Import ssreflect ssrfun ssrbool eqtype ssrnat seq.
+From mathcomp Require Import ssreflect ssrfun ssrbool eqtype ssrnat seq nmodule.
 From mathcomp Require Import preorder.
-From mathcomp Require Export ssralg.
+From mathcomp Require Import rings_modules_and_algebras divalg.
```
pi8027, line 10: @proux01 Do you know why `ssralg` was exported here and also in `field_tactic.v`? (It looks like we need to turn them into `Import` to fix some breakages in CI.)

pi8027: Or we need to export `decfield` here, even though it is not needed here. Otherwise `Import GRing.Theory` doesn't import the right `GRing.Theory` module.

pi8027: Ok, now I see that not exporting `ssralg` here breaks `hmap_ops.v` in the `graph-theory` library because it imports `ring_tactic` but not `rings_modules_and_algebras`, and it cannot find the semiring instance on `nat`. I think the right fix is to explicitly import `rings_modules_and_algebras` there.

field/falgebra.v
```diff
@@ -1064,11 +1066,6 @@ Proof. by apply/ahomP_tmp; case: f. Qed.
 HB.instance Definition _ (f : ahom aT rT) :=
   GRing.isMonoidMorphism.Build aT rT f (ahom_is_monoid_morphism f).
 
-#[warning="-deprecated-reference-since-mathcomp-2.5.0",
-  deprecated(since="mathcomp 2.5.0", use=ahom_is_monoid_morphism)]
-Definition ahom_is_multiplicative (f : ahom aT rT) : multiplicative f :=
-  (fun p => (p.2, p.1)) (ahom_is_monoid_morphism f).
```
pi8027, line 1070: I thought we could safely remove this since it is a deprecated internal lemma, but it is actually used in `mathcomp-abel`. I will move the definition of `multiplicative` from `ssralg` to `rings_modules_and_algebras`, and revert this change.

algebra/ring_tactic.v
proux01: I agree, graph-theory should be fixed (overlay to merge before the current PR).

field/falgebra.v
proux01: Couldn't we rather fix mathcomp-abel?

pi8027: math-comp/Abel#108 should address the issue on Abel's side.

algebra/algebraic_hierarchy/algebraic_hierarchy.v
pi8027: And indeed, I forgot to export `ring_quotient` here. Fixed.

field/falgebra.v
pi8027: `mathcomp-abel` compiles in CI now. No need to fix the issue on the MC side.

algebra/algebraic_hierarchy/ssralg.v
pi8027: Done.

algebra/ring_tactic.v
pi8027: I decided to revert the change and suppress the warning since I'm not going to write the overlay.

pi8027:
CI green. This PR is ready.

